PR #5, opened 2017-04-25
Add mapping for Container entities

zakiva:
Adding a static mapping between k8s entities to miq entities.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
@@ -1,0 +1,25 @@
+module ManageIQ::Providers::Kubernetes::ContainerManager::EntitiesMapping
+  # Update this mapping with any new containers entity added to manageiq
+  MAPPING = {
+    'node'                    => 'ContainerNode',
+    'namespace'               => 'ContainerProject',
+    'resource_quota'          => 'ContainerQuota',
+    'limit_range'             => 'ContainerLimit',
+    'replication_controller'  => 'ContainerReplicator',
+    'persistent_volume_claim' => 'PersistentVolumeClaim',
+    'persistent_volume'       => 'PersistentVolume',
+    'pod'                     => 'ContainerGroup',
+    'service'                 => 'ContainerService',
+    'component_status'        => 'ContainerComponentStatus',
+    'project'                 => 'ContainerProject',
+    'route'                   => 'ContainerRoute',
+    'build_config'            => 'ContainerBuild',
+    'build'                   => 'ContainerBuildPod',
+    'template'                => 'ContainerTemplate',
+    'image'                   => 'ContainerImage'
+  }.freeze
+
+  def miq_entity(entity)
+    MAPPING[entity]
+  end
+end
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -3,6 +3,7 @@ require 'shellwords'
 module ManageIQ::Providers::Kubernetes
   class ContainerManager::RefreshParser
     include Vmdb::Logging
+    include ContainerManager::EntitiesMapping
     def self.ems_inv_to_hashes(inventory, options = Config::Options.new)
       new.ems_inv_to_hashes(inventory, options)
     end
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
```diff
@@ -0,0 +1,24 @@
...
+    MAPPING = {
+      'node'                    => 'ContainerNode',
+      'namespace'               => 'ContainerProject',
+      'resource_quota'          => 'ContainerQuota',
+      'limit_range'             => 'ContainerLimit',
+      'replication_controller'  => 'ContainerReplicator',
+      'persistent_volume_claim' => 'PersistentVolumeClaim',
+      'persistent_volume'       => 'PersistentVolume',
+      'pod'                     => 'ContainerGroup',
+      'endpoint'                => 'ContainerEndpoint', # I see them in the refresh but do we save them?
```
moolitayer: Looks like we don't

moolitayer:
At the bottom of app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb we have container_image and container_image_registries. I think you can replace those two. (they might use singular as data index while all the others use plural, so you would have to change that)

app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
cben: We parse into `:container_endpoints`, extracting subsets -> addresses with targetRef.kid == Pod.
Then `parse_service` looks at those to compute the service's container_groups.
AFAICT we don't store anything for the endpoint itself, in any case there is no container_endpoints table nor model.

moolitayer: ahh thanks, didn't read my grep result well

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -31,85 +31,93 @@ def ems_inv_to_hashes(inventory, _options = Config::Options.new)
...
-        @data_index.store_path(:container_services, :by_namespace_and_name, se[:namespace], se[:name], se)
+      services = map_entity("service")
+      process_collection(inventory["service"], services) { |s| parse_service(s) }
+      @data[services].each do |se|
+        @data_index.store_path(services, :by_namespace_and_name, se[:namespace], se[:name], se)
       end
     end
 
     def get_replication_controllers(inventory)
+      replication_controllers = map_entity("replication_controller")
```
cben: I must say I'm not sure it's a win.
My first thought also was "ooh, you have this table, where could we use it — I know, in the refresh!".
But the old code was explicit, it was trivial to line up the parsing code producing say :replication_controllers with saving code using them; now it will require looking at a third file...

What's worse, it's now hard to line up `store_path` with corresponding `fetch_path`s in this file.  What do you think of changing them to `fetch_path(map_entity(...), ...)`?

However, there are also benefits.  (I'm not really opposed).  
Ideas (your call if you adopt them):
- call the var `key = map_entity(...)` everywhere?
- have process_collection() return `@data[foos]` so you can chain `.each` on it directly?

simon3z:
@zakiva can you address the comments / fix travis? Thanks.

simon3z:
@zakiva we'll probably have to discuss this, ping me when available.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -145,6 +153,10 @@ def map_labels(model_name, labels)
       ContainerLabelTagMapping.map_labels(@label_tag_mapping, model_name, labels)
     end
 
+    def table_for(entity)
+      miq_entity(entity).pluralize.underscore.to_sym
```
simon3z: @zakiva @cben I think there is a more rails-native way to do this.

simon3z: @zakiva this is how `activerecord` is generating the name:

https://github.com/rails/rails/blob/97bd56e674603529d96c6f6e85d695dd24208afb/activerecord/lib/active_record/model_schema.rb#L481-L484

```ruby
        def undecorated_table_name(class_name = base_class.name)
          table_name = class_name.to_s.demodulize.underscore
          pluralize_table_names ? table_name.pluralize : table_name
        end
```

Too bad we can't reuse it (it's private) but at least we can get good insights (e.g. `demodulize`). Although remember that for us it's not a table name.

simon3z:
@cben @moolitayer can you review/approve this?

moolitayer:
I think we can waver the code climate "Similar code found" issue.
That is an issue not introduced here. This PR can actually help to solve this in the future.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -73,43 +78,46 @@ def get_endpoints(inventory)
...
     def get_limit_ranges(inventory)
-      process_collection(inventory["limit_range"], :container_limits) { |n| parse_range(n) }
+      process_collection(inventory["limit_range"], path_for_entity("limit_range")) { |n| parse_range(n) }
     end
 
     def get_component_statuses(inventory)
+      key = path_for_entity("component_status")
       process_collection(inventory["component_status"],
-                         :container_component_statuses) do |cs|
+                         key) do |cs|
```
cben: I'd join this with prev line.  also 5 lines below.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -145,6 +153,10 @@ def map_labels(model_name, labels)
       ContainerLabelTagMapping.map_labels(@label_tag_mapping, model_name, labels)
     end
 
+    def path_for_entity(entity)
+      miq_entity(entity).tableize.to_sym
+    end
```
cben: nit: can you place this towards the end, after all calls to it?  this file is unfortunately not well sorted.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
zakiva: @cben done, thanks

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
zakiva: done

simon3z:
@cben can you reivew/approve?

PR #6, opened 2017-04-26
add authentication to image inspector requests [old]

ilackarms:
passes shared secret to image inspector using pod env

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
@@ -14,6 +14,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   IMAGE_INSPECTOR_SA = 'inspector-admin'
   INSPECTOR_ADMIN_SECRET_PATH = '/var/run/secrets/kubernetes.io/inspector-admin-secret-'
   PROXY_ENV_VARIABLES = %w(no_proxy http_proxy https_proxy)
+  INSPECTOR_AUTH_TOKEN = "INSPECTOR_AUTH_TOKEN".freeze
 
   def load_transitions
     self.state ||= 'initializing'
@@ -53,7 +54,8 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
       :image_full_name => image.full_name,
       :pod_name        => "manageiq-img-scan-#{guid[0..4]}",
       :pod_port        => INSPECTOR_PORT,
-      :pod_namespace   => namespace
+      :pod_namespace   => namespace,
+      :auth_token      => SecureRandom.hex
     ))
 
     _log.info("Getting inspector-admin secret for pod [#{pod_full_name}]")
@@ -136,7 +138,8 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
       :pod_namespace => options[:pod_namespace],
       :pod_name      => options[:pod_name],
       :pod_port      => options[:pod_port],
-      :guest_os      => IMAGES_GUEST_OS
+      :guest_os      => IMAGES_GUEST_OS,
+      :auth_token    => auth_token
     }
 
     verify_error = verify_scanned_image_id(image_inspector_client.fetch_metadata)
@@ -287,7 +290,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
         :verify_ssl => ext_management_system.verify_ssl_mode,
         :cert_store => ext_management_system.ssl_cert_store
       },
-      :auth_options   => kubeclient.auth_options,
+      :auth_options   => kubeclient.auth_options.merge(:auth_token => auth_token),
       :http_proxy_uri => kubeclient.http_proxy_uri
     )
   end
@@ -351,6 +354,8 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   end
 
   def pod_definition(inspector_admin_secret_name)
+    pod_env = inspector_proxy_env_variables << { :name  => INSPECTOR_AUTH_TOKEN,
+                                                 :value => auth_token }
     pod_def = {
       :apiVersion => "v1",
       :kind       => "Pod",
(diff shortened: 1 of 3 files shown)


ilackarms:
Token authentication is being added to all HTTP requests on the SSA image inspector pod in https://github.com/openshift/image-inspector/pull/38

The intent of this PR is to secure the image inspector pod, which currently performs no authentication to requests that may potentially come from other pods within the cluster.

These changes update ManageIQ:
- generate a shared secret token when creating the Image Inspector Pod
- authenticate all requests to Image Inspector Pod with header `X-Auth-Token: $TOKEN` 
- pass `:auth_token` parameter to `image_inspector_client` which is also responsible for making some HTTP requests to image inspector (see PR https://github.com/moolitayer/image-inspector-client/pull/17)

ilackarms:
@enoodle @simon3zitayer  please re-review this when you have a chance. note that merging moolitayer/image-inspector-client#17 and openshift/image-inspector#38 are necessary for this PR to merge cleanly

ilackarms:
when Image Inspector 2.2 is released https://github.com/openshift/image-inspector/milestone/1 this change should be merged

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -325,6 +328,7 @@ def pod_health_poll(client, health_url, http_opts)
     Net::HTTP.start(health_url.host, health_url.port, http_opts) do |http|
       request = Net::HTTP::Get.new(health_url.path)
       client.headers.each { |k, v| request[k.to_s] = v }
+      request["X-Auth-Token"] = auth_token
```
moolitayer: Hi @ilackarms,
How come the line above (330) isn't enough?

moolitayer: I suspect we don't use this method anymore since https://github.com/ManageIQ/manageiq-providers-kubernetes/commit/83def745cf74eb8e541624409fa8e6f7c0f9f6f1. @enoodle WDYT?

enoodle: This is true, we use the ReadinessProbe from Openshift. We can add this header to the readiness probe that is defined in the pod definition.
We should delete this function as it is not being used.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -396,7 +402,7 @@ def pod_definition(inspector_admin_secret_name)
                 :name      => "docker-socket"
               }
             ],
-            :env             => inspector_proxy_env_variables,
+            :env             => pod_env,
             :readinessProbe  => {
               "initialDelaySeconds" => 15,
               "periodSeconds"       => 5,
```
enoodle: @ilackarms inside the "httpGet" (can't comment there)
```
        "httpHeaders" => {
            "name" => "X-Auth-Token",
            "value" => auth_token
         },
```

from here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
ilackarms: i needed to rebase; this PR is pretty old. We needed to make sure to authenticate health checks, but since they don't happen here anymore, it's removed.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
ilackarms: thanks!

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -137,7 +139,8 @@ def analyze
       :pod_namespace => options[:pod_namespace],
       :pod_name      => options[:pod_name],
       :pod_port      => options[:pod_port],
-      :guest_os      => IMAGES_GUEST_OS
+      :guest_os      => IMAGES_GUEST_OS,
+      :auth_token    => options[:auth_token]
```
moolitayer: `auth_token` ?

moolitayer: :auth_token    => auth_token

ilackarms: whoops, good catch

ilackarms:
@moolitayer any idea why travis suddenly started failing? it doesn't look related to my changes

moolitayer:
> @moolitayer any idea why travis suddenly started failing? it doesn't look related to my changes

This seems related to graph refresh and is also failing on my local master.
@cben what do we have to do to fix Travis?

cben:
Looking into graph refresh failure...

cben:
Dunno, but hopefully will go away (once bin/setup is fixed and we can run Travis).

Works for me on master, and on checkout of this PR.
`./app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb:106`, nor anything else from this repo, doesn't and didn't pass `use_ar_object`.

The core manageiq source Travis is quoting is indeed before https://github.com/ManageIQ/manageiq/pull/15439 that added `use_ar_option:` to that constructor (that commit looks old but was merged tonight), but then (and even now), I can't find any code *passing* `use_ar_object`.

(I looked at Mooli's local failure, it's slightly different, was caused by old spec/manageiq).

ilackarms:
@moolitayer @cben any thoughts on these failures? https://travis-ci.org/ManageIQ/manageiq-providers-kubernetes/jobs/264266393

cben:
Weird.  Looks as if it was tested with #69 which sets `quota_desired_int` without the schema change adding this column https://github.com/ManageIQ/manageiq-schema/pull/44
Except neither of those is merged, and your commit is not based on #69 so what is `quota_desired_int` doing in this Travis run?!  @zakiva any idea?
Anyway I'd try rebasing this PR...

ilackarms:
@cben i just rebased, same issue

zakiva:
> @zakiva any idea?

don't know... @ilackarms maybe try to open a new PR

ilackarms:
closing this and retrying with a new PR

PR #7, opened 2017-04-27
Avoid metrics collection workers unless endpoint

moolitayer:
Starting with k8s since container_manager_mixin is shared with openshift.
We will then need to do one of:
- copy emses_in_zone to openshift's metrics_collector_worker.rb
OR
- create a common component for both metrics collectors

supports_metrics? aims to be replaced by a support feature.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1437138#c0

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_collector_worker.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_collector_worker.rb
@@ -11,5 +11,14 @@ module ManageIQ::Providers
     def self.ems_class
       ManageIQ::Providers::Kubernetes::ContainerManager
     end
+
+    # Override PerEmsTypeWorkerMixin.emses_in_zone to limit metrics collection
+    def self.emses_in_zone
+      super.select do |ems|
+        ems.supports_metrics?.tap do |supported|
+          _log.info("Skipping [#{ems.name}] since it has no metrics endpoint") unless supported
+        end
+      end
+    end
   end
 end
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -14,6 +14,10 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     end
   end
 
+  def supports_metrics?
+    connection_configurations.hawkular != nil
+  end
+
   module ClassMethods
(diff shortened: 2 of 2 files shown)


moolitayer:
cc @durandom @yaacov @cben @simon3z

app/models/manageiq/providers/kubernetes/container_manager/metrics_collector_worker.rb
```diff
@@ -11,5 +11,14 @@ def friendly_name
     def self.ems_class
       ManageIQ::Providers::Kubernetes::ContainerManager
     end
+
+    # Override PerEmsTypeWorkerMixin.emses_in_zone to limit metrics collection
+    def self.emses_in_zone
+      super.select do |ems|
+        (ems.supporty_metrics?).tap do |hawkular|
```
durandom: * I would not call the arg `hawkular`, because  `supports_metrics?` just returns a boolean. How about `available` or `supported`
*  the `()`are not needed
* Typo: supporty vs supports

durandom:
I'm ok with adding the method before adding a feature. @moolitayer just be sure to push adding the feature once you have a full featured and working implementation

cben:
Nice.
@yaacov is right.  Backward compatibility is hard, depends on when provider was created (or last Edited):

- ≤capablanca was creating single-endpoint providers.  their metrics did work on ≥darga if they had it on master port 5000 (or by changing `metrics_port` setting).  though on capablanca itself I don't see this code (no hawkular_client_mixin.rb, no `metrics_port`).
  Are you OK with such providers no longer getting metrics (requiring manual Edit to fix)?

- ≥darga always created hawkular endpoints, sometimes with empty hostname.
  I suppose such providers will just continue to try (and fail) getting metrics, we can later do a migration deleting empty-hostname endpoints...

simon3z:
@moolitayer @cben any label? (bug? enhancement? backport? etc.)

@moolitayer can you approve this?

moolitayer:
@simon3z should reach 5.8.1
@miq-bot add_label enhancement, fine/yes

simaishi:
@moolitayer @simon3z https://bugzilla.redhat.com/show_bug.cgi?id=1437138 is now for 'master' only. In the BZ, it says it's safe to backport this alone, but not sure what issue this will fix. Should this still be backported?

moolitayer:
I was just on my way here... Better not to backport

@miq-bot add_label fine/no

PR #10, opened 2017-05-01
kubernetes_connect: add timeout settings

cben:
Ability to constrol kubeclient timeouts from settings.yml

openshift half: ~~https://github.com/ManageIQ/manageiq-providers-openshift/pull/8~~ unnecessary on master given https://github.com/ManageIQ/manageiq-providers-openshift/pull/7, but will add it in backports.

Tested:
- [x] [Exercised](https://github.com/abonas/kubeclient/pull/246#issuecomment-298221765) in rails console and refresh, events, SSA in manageiq, confirmed low enough values cause timeouts.
- [x] core manageiq tests passed locally with this, but I don't think they cover anything relevant.

https://bugzilla.redhat.com/show_bug.cgi?id=1440950
@miq-bot add-label euwe/yes, fine/yes (actually we plan a 5.7.1 hotfix but I suppose we also want normal backports, so customer uprading will not lose hotfix functionality?)

@moolitayer @agrare Please review.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -36,6 +36,8 @@
         :network:
           :critical:
             - POD_HOSTPORTCONFLICT
+    :open_timeout: 60.seconds
+    :read_timeout: 60.seconds
 :http_proxy:
   :kubernetes:
     :host:
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -31,7 +31,11 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
         options[:version] || kubernetes_version,
         :ssl_options    => Kubeclient::Client::DEFAULT_SSL_OPTIONS.merge(options[:ssl_options] || {}),
         :auth_options   => kubernetes_auth_options(options),
-        :http_proxy_uri => VMDB::Util.http_proxy_uri
+        :http_proxy_uri => VMDB::Util.http_proxy_uri,
+        :timeouts       => {
+          :open => Settings.ems.ems_kubernetes.open_timeout.to_f_with_method,
+          :read => Settings.ems.ems_kubernetes.read_timeout.to_f_with_method
+        }
       )
     end
 
--- a/spec/models/manageiq/providers/kubernetes/container_manager_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager_spec.rb
@@ -112,7 +112,8 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager do
       require 'kubeclient'
       expect(Kubeclient::Client).to receive(:new).with(
         instance_of(URI::HTTPS), 'v1',
-        hash_including(:http_proxy_uri => VMDB::Util.http_proxy_uri)
+        hash_including(:http_proxy_uri => VMDB::Util.http_proxy_uri,
+                       :timeouts       => match(:open => be > 0, :read => be > 0))
       )
       described_class.raw_connect(hostname, port, options)
     end


cben:
https://bugzilla.redhat.com/show_bug.cgi?id=1440950
@miq-bot add-label euwe/yes, fine/yes

moolitayer:
@cben will our gem dependency in euwe bring in the new gem automatically or will we need to upgrade it?

cben:
we have strict `=2.3.0` on euwe, fine and master, will need a gems-pending PR and backport it.

agrare:
@cben any luck getting a new gem released?

simon3z:
> @cben any luck getting a new gem released?

@cben @agrare gem is released here:  https://rubygems.org/gems/kubeclient/versions/2.4.0

config/settings.yml
```diff
@@ -3,6 +3,8 @@
   :ems_kubernetes:
     :event_handling:
       :event_groups:
+    :open_timeout: 60.seconds
+    :read_timeout: 60.seconds
```
simon3z: @cben @agrare wasn't the default 2 minutes? Are we shortening this timeout by default?

cben: The default read_timeout in ruby / restclient was always 60.  (open_timeout was infinite in ruby 2.2 but afaict euwe appliance was already ruby 2.3)

I was never able to account for the "2 minutes" number.
- "60 + 60 = 120" is not a plausible explanation, doesn't take a minute to establish tcp + tls.
- Curl showed server takes >2min but that doesn't mean anything, client may timeout after 1min.
- IIRC the source for us believing we timeout at 2 minutes is log lines ~2min apart.  But we don't have per-request log lines, we have something like "start of refresh – error = 2min"...  there are many requests before images.

cben: Darn.  Customer log strongly suggests timeout was 2min.
It contains >70 timeouts, and times are very stable: 9~12sec from first connect (/api) to second connect (/oapi) then 126~128sec to timeout.
This agrees perfrect with per-request timing in VCR: all /api requests total 11sec, /oapi requests without images total 7sec.

Gonna simulate/reproduce a slow server and measure actual timeout before & after patch...

simon3z:
@cben let me know when this is ready.

config/settings.yml
cben: => It takes 2 minutes on euwe with old kubeclient too.  Nothing changed here.
The reason turned out to be that ruby's Net::HTTP unconditionally retries requests that are supposed to be idempotent (e.g. GET, DELETE but not POST) [https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts/issues/8, https://bugs.ruby-lang.org/issues/10674]

@simon3z ready for merge.

config/settings.yml
```diff
@@ -36,11 +36,11 @@
...
 :http_proxy:
   :kubernetes:
     :host:
     :password:
     :port:
     :user:
-:container_scanning:
-  :scanning_job_timeout: 20.minutes
```
simon3z: @cben really? :open_mouth: I guess you needed a more careful rebase :blush:

cben: whoa thanks fixed.
need to revise my mergetool config, meld without `--auto-merge` is error prone...

simaishi:
@cben Marking as `euwe/conflict` for now. Please remove the conflict label when you have Euwe PR. Thanks!

cben:
@miq-bot remove-label euwe/conflict
https://github.com/ManageIQ/manageiq/pull/15188

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/15188

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/15090

PR #11, opened 2017-05-03
Option needed for new ems_refresh.openshift.store_unused_images setting

cben:
#kubernetes part of https://github.com/ManageIQ/manageiq/pull/14662 after repo split.
This will have to be merged before openshift part https://github.com/ManageIQ/manageiq-providers-openshift/pull/9.

Strawman alternative/baseline for https://github.com/ManageIQ/manageiq/pull/14628.
Will allow to still get_container_images=true, but instead of saving metadata (labels etc) on all images openshift gave us, save it only for images that have been mentioned by pods.

Here called `store_new_images` because this function doesn't have the context to understand which images are "used".
For `store_unused_images`=false setting, it'll be called twice:
- from parse_pods with true, adding all images
- from parse_openshift_image with false, only enriching previous images

cc @agrare @enoodle @zeari

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -1100,7 +1100,8 @@ module ManageIQ::Providers::Kubernetes
       res
     end
 
-    def parse_container_image(image, imageID)
+    # may return nil if store_new_images = false
+    def parse_container_image(image, imageID, store_new_images: true)
       container_image, container_image_registry = parse_image_name(image, imageID)
       host_port = nil
 
@@ -1124,6 +1125,7 @@ module ManageIQ::Providers::Kubernetes
         :container_image, :by_digest, container_image_identity)
 
       if stored_container_image.nil?
+        return nil unless store_new_images
         @data_index.store_path(
           :container_image, :by_digest,
           container_image_identity, container_image
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
@@ -658,6 +658,15 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser do
         expect(first_obj).to be(second_obj)
       end
     end
+
+    it "returns existing image or nil with store_new_images=false" do
+      obj1 = parser.parse_container_image(shared_image_without_host, shared_ref)
+      obj2 = parser.parse_container_image(shared_image_without_host, shared_ref, :store_new_images => false)
+      obj3 = parser.parse_container_image(shared_image_without_host, unique_ref, :store_new_images => false)
+      expect(obj1).not_to be nil
+      expect(obj2).to be obj2
+      expect(obj3).to be nil
+    end
   end
 
   describe "cross_link_node" do


cben:
@simon3z @blomquisg @agrare please review/merge...

cben:
The relevant BZ is probably https://bugzilla.redhat.com/show_bug.cgi?id=1436176 "OPenShift Refresh duration exceeds default two hour timeout and grows > 8GB never fully completing"

cben:
I believe this is ready
@miq-bot assign simon3z

cben:
@miq-bot add-label enhancement

cben:
ping @simon3z.
rebased both again.  this should be merged first, then will rerun https://github.com/ManageIQ/manageiq-providers-openshift/pull/9 tests.

simon3z:
@cben let's focus on the graph refresh for now.

cben:
Sure, but is anything stopping this from being merged?

simon3z:
> Sure, but is anything stopping this from being merged?

@cben yes capacity in supporting different flows of collection for images (that both needs to be tested).
On a related note, isn't https://github.com/ManageIQ/manageiq/pull/14808 related to this? In order to avoid losing information about openshift images that were scanned, etc.

cben:
@simon3z @enoodle I'm closing these for now, if you'll want to revive them, it should be a simple rebase.

Good point that this would increase exposure to ManageIQ/manageiq#14808.

cben:
Reviving, @enoodle @agrare PTAL.  (openshift PR WIP, will reopen soon)

@simon3z: yes it will increase image disconnection (#103) but so will `get_container_images: false` which is already merged.  This is more expensive but strictly safer.
We have a user env where this is neccessary (they need image labels in reports but don't need images that are not running).

agrare:
@cben :+1: from me

simon3z:
@cben ok on the idea but I wish we had a better code solution (I wish we had two different methods to call, one to just enrich and one to add new found images, of course they could reuse some common code).

I don't want to block this if it's urgent but I really hope we can improve it.

cben:
Ack, I'd love to refactor the image code. 
`parse_container_image` is only called twice — from pods and from openshift images.
I can extract the bulk of it into 2 helpers `get_or_store_container_image_registry`,
`get_or_store_container_image`; then I'd call `parse_image_name` and those 2 directly, with a variation when store_unused_images=false.
But that's starting to resemble `InventoryCollection#find_or_build_by` primitive a lot, and it's funny to reinvent that wheel in data_index then possibly convert to `InventoryCollection`.
I also have an upcoming *new* use for data_index for tag mapping, and want to think if there is a cleaner way.

=> For now this is small and backportable, I'd like to merge it, and do a followup after I think more...

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -1100,7 +1100,8 @@ def parse_container_state(state_hash)
       res
     end
 
-    def parse_container_image(image, imageID)
+    # may return nil if store_new_images = false
+    def parse_container_image(image, imageID, store_new_images: true)
```
Ladas: store_new_images or store_unused_images ?

simon3z: @cben @Ladas also, it's so confusing with the other variables named `stored_container_image_*`

cben: This function doesn't have the context to understand what's used.
It gets called twice, first adding images from pods, second maybe not adding images from openshift registry.

Everybody asked this.  OK, I see I better rewrite this.

cben:
reusing BZ https://bugzilla.redhat.com/show_bug.cgi?id=1436176, same as get_container_images and other band-aids?

agrare:
@cben that BZ is already in post so I think we should open a new BZ

agrare:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1486483

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/15917

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/15907

PR #12, opened 2017-05-03
bump ruby to 2.3.3 and add 2.4.1

durandom:
cc @jrafanie 
same as https://github.com/ManageIQ/manageiq-providers-lenovo/pull/36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 rvm:
-- '2.3.1'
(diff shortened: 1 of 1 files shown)


simon3z:
@moolitayer can you review/verify this? (Probably in conjunction with https://github.com/ManageIQ/manageiq-providers-openshift/pull/10)

simon3z:
cc @cben

Issue #2, opened 2017-04-21
move kubernetes specific events and settings


durandom:
see the [amazon settings](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/config/settings.yml) for what can be moved

similar to moving the settings we can also move event names (see [amazon](https://github.com/ManageIQ/manageiq-providers-amazon/pull/163/files) for an example)

simon3z:
@nimrodshn can you look into this? Thanks.

simon3z:
@nimrodshn @cben this can be closed right? Can you reference the PR here? Thanks.

cben:
Previous PRs:
- https://github.com/ManageIQ/manageiq/pull/15034
- https://github.com/ManageIQ/manageiq-providers-openshift/pull/11 later undone by https://github.com/ManageIQ/manageiq-providers-openshift/pull/12 and redone to kubernetes in #15.

That was just the events config, there are other settings to move.
https://github.com/ManageIQ/manageiq/blob/master/config/settings.yml contains containers/kube/openshift stuff under `:ems_refresh:`, `:container_scanning:`, `:container_entities:` and more.  (see amazon example above)

simon3z:
> That was just the events config, there are other settings to move.
> https://github.com/ManageIQ/manageiq/blob/master/config/settings.yml contains containers/kube/openshift stuff under :ems_refresh:, :container_scanning:, :container_entities: and more. (see amazon example above)

cc @nimrodshn

nimrodshn:
@simon3z on in :+1:

cben:
I believe everything has been moved, can be closed.

PR #14, opened 2017-05-09
configurable timeout for SSA

enoodle:
This adds an option to configure timeout for container ssa jobs:

- With the global settings file, under "container_scanning" in the field "scanning_job_timeout"

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1447672

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
@@ -264,6 +264,10 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
     end
   end
 
+  def self.current_job_timeout(timeout_adjustment = 1)
+    ::Settings.container_scanning.scanning_job_timeout.to_f_with_method * timeout_adjustment
+  end
+
   private
 
   def ext_management_system
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -9,3 +9,5 @@
     :password:
     :port:
     :user:
+:container_scanning:
+  :scanning_job_timeout: 20.minutes
--- a/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
@@ -160,6 +160,15 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
       end
     end
 
+    context "#current_job_timeout" do
+      it "checks for timeout in Settings" do
+        stub_settings_merge(:container_scanning => {:scanning_job_timeout => '15.minutes'})
+        expect(@job.send(:current_job_timeout)).to eq(900)
+        stub_settings_merge(:container_scanning => {:scanning_job_timeout => 600})
+        expect(@job.send(:current_job_timeout)).to eq(600)
+      end
+    end
+
     it 'should add correct environment variables' do
       att_name = 'http_proxy'
       my_value = "MY_TEST_VALUE"


enoodle:
cc @moolitayer @simon3z @roliveri 

This is a replacement for https://github.com/ManageIQ/manageiq/pull/15032 which seems too messy to me now as we can manage this in this repo.

enoodle:
@simon3z I can remove the `custom_attribute` part but then it won't be possible to set the timeout per provider, are we ok with this?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -442,4 +443,16 @@ def inspector_proxy_env_variables
               :value => att.value} unless att.value.blank?
     end
   end
+
+  def current_job_timeout(timeout_adjustment = 1)
+    timeout_adjustment = 1 if defined?(RSpec)
```
cben: Please find other way to change behavior in tests (e.g. stub settings, pass `timeout_adjustment` arg, maybe use timecop...)

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -442,4 +443,16 @@ def inspector_proxy_env_variables
...
+    target = target_entity
+    ems_settings = target.try(:ext_management_system).try(:custom_attributes)
+    if ems_settings && ems_settings.find_by(:name => "scanning_job_timeout", :section => "metadata")
+      return ems_settings.find_by(:name => "scanning_job_timeout").value.to_i
+    end
+    settings_timeout = ::Settings.try(:container_scanning).try(:scanning_job_timeout)
```
cben: I think you can add this in settings.yml and then assume it always exists, so drop DEFAULT_TIMEOUT.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: :+1: I just moved it, I didn't really need it and got it from "vm_scan.rb"

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: Ok, I will have to do it in a separate PR, So I prefer to leave this here until it is merged.

cben: Because of repos?  You can add to this repo's config/settings.yml (they're all deep-merged together).

Soon all container-related settings will be in the container repos (#2, ManageIQ/manageiq-providers-openshift#3), but new settings can already go here.

enoodle: :+1: This is nice. I moved the settings there.

simon3z:
@miq-bot assign roliveri 

@roliveri can you review this?

simon3z:
@enoodle let me know when this is finalized (timeout tested with Einat).

config/settings.yml
```diff
@@ -9,3 +9,5 @@
     :password:
     :port:
     :user:
+:container_scanning:
+  :scanning_job_timeout: 1200
```
moolitayer: :+1: for 20 min. 
This is a compromise between giving the job enough time without blocking the cluster from scans too long if something is wrong(only three concurrent scans)

moolitayer: Also scanning that takes > 20 min is probably useless for testing as well.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -442,4 +442,13 @@ def inspector_proxy_env_variables
               :value => att.value} unless att.value.blank?
     end
   end
+
+  def current_job_timeout(timeout_adjustment = 1)
+    target = target_entity
+    ems_settings = target.try(:ext_management_system).try(:custom_attributes)
```
moolitayer: this can simplify the if statement
```
ems_settings = target.try(:ext_management_system).try(:custom_attributes) || ExtManagementSystem.none
```

enoodle: I think that we should assume that we have an ems here as we already assume that in other places in this class.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -442,4 +442,13 @@ def inspector_proxy_env_variables
...
+    if ems_settings && ems_settings.find_by(:name => "scanning_job_timeout", :section => "metadata")
+      return ems_settings.find_by(:name => "scanning_job_timeout").value.to_i
+    end
+    ::Settings.container_scanning.scanning_job_timeout * timeout_adjustment
```
cben: If you use `::Settings.container_scanning.scanning_job_timeout.to_i_with_method`
you'll be able to write more readable `:scanning_job_timeout: 20.minutes` in the yml.
See #10 for example.
Could do it with a custom attribute too.

and maybe consider `to_f` / `to_f_with_method`, really short fractional timeouts can be handy in tests.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
moolitayer: That's fine, I'm suggesting a cosmetic change to save a conditional. Maybe:
```
ems_settings = target.try(:ext_management_system).try(:custom_attributes) || CustomAttributes.none
timeout_attribute = ems_settings.find_by(:name => "scanning_job_timeout", :section => "metadata")
return timeout_attribute.value.to_i if timeout_attribute
::Settings.container_scanning.scanning_job_timeout * timeout_adjustment
```

enoodle: :+1:

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -264,6 +264,14 @@ def queue_callback(state, msg, _)
     end
   end
 
+  def current_job_timeout(timeout_adjustment = 1)
+    target = target_entity
+    ems_settings = target.ext_management_system.custom_attributes
+    timeout_attribute = ems_settings.find_by(:name => "scanning_job_timeout", :section => "metadata")
```
simon3z: @enoodle I already previously mentioned that we should stop using `custom_attributes` in this fashion until a proper solution is identified (even more in the `metadata` section, which is definitely not designed for this).

enoodle: > @simon3z I can remove the custom_attribute part but then it won't be possible to set the timeout per provider, are we ok with this?

simon3z: @enoodle yes

simon3z:
@moolitayer @cben can you review/approve?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -264,6 +264,10 @@ def queue_callback(state, msg, _)
     end
   end
 
+  def current_job_timeout(timeout_adjustment = 1)
```
moolitayer: can you add 
```delegate :current_job_timeout, :to => :class```
Like models/job.rb has and have this as a `self` method

I'm not even sure if the delegate here is needed as well.

this will work with self right? (i)

Comment (enoodle) [inline app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb]: :+1: I couldn't make it a class method before but now we can. I added these.

Comment (enoodle) [inline app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb]: the `delegate :current_job_timeout, :to => :class` line is from the base class. I am also not sure if it is needed twice.

File: spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb [hunk 8]
```diff
@@ -160,6 +160,15 @@ def fetch_oscap_arf
       end
     end
 
+    context "#current_job_timeout" do
+      it "checks for timeout in Settings" do
+        stub_settings_merge(:container_scanning => {:scanning_job_timeout => 15.minutes})
```
cben: `15.minutes` evaluates to the number 900 (technically an `ActiveSupport::Duration` but pretty much a number).
I'd stub to string `:scanning_job_timeout => '15.minutes'` like what happens in settings.yml,
to actually test the ability to parse these strings.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: I checked and tested - its not needed. this line at the base class tells it to search the function in the objects' class.

spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
enoodle: :+1:

simon3z:
@enoodle did you test this with Einat?
@enoodle any label for this PR? Bug? Backports?

enoodle:
@miq-bot  add_label bug fine/yes

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -264,6 +264,10 @@ def queue_callback(state, msg, _)
     end
   end
 
+  def self.current_job_timeout(timeout_adjustment = 1)
+    ::Settings.container_scanning.scanning_job_timeout.to_f_with_method * timeout_adjustment
```
simon3z: @cben @enoodle any chance `scanning_job_timeout` could be undefined for old version (backports)? If so I assume the value would be 0, resulting in 0 as timeout for the job? Should we add a default in code (if undefined) as well?

enoodle: We can to this. Also, When we backport this we can also backport the line we add in `config/settings.yml`.
What do you prefer?

enoodle:
@simon3z I have verified this with Einat successfully

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
simon3z: @enoodle of course we want to backport the `config/settings.yml` as well but I am asking if they'll be effective even on existing installations. cc @cben

cben:
hmm.  I believe these separate files didn't exist before repo split.
=> will require a manual backport PR putting them in manageiq's
config/settings.yml.  that works far back (darga)...

can add test using real unstubbed settings, assert it's a positive number,
that way we'll get failure if we backport without the setting.

enoodle:
@simon3z I added a backport to fine PR that covers the settings and tests.

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/15074

PR #15, opened 2017-05-10
move kubernetes related events

nimrodshn:
Moving kubernetes related events from provider-openshift to correct repo.
cc: @cben @simon3z


simon3z:
@cben can you review/approve?

PR #16, opened 2017-05-16
Make recursive openstruct an explicit dependency

jrafanie:
Previously, we got this dependency implicitly through manageiq,
kubeclient, etc.  Since we use this heavily in tests here, we should
state our reliance on this gem.

Needed for: https://github.com/ManageIQ/manageiq/pull/15096


jrafanie:
@simon3z please review

PR #18, opened 2017-05-18
Provide a reverse lookup for provider entities

kevensen:
There are instances where are "reverse" lookup of entity mappings is desirable.  This is needed for pull requests:

https://github.com/ManageIQ/manageiq/pull/14798
https://github.com/ManageIQ/manageiq-providers-openshift/pull/2
https://github.com/ManageIQ/manageiq-automation_engine/pull/7

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
@@ -22,4 +22,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EntitiesMapping
   def miq_entity(entity)
     MAPPING[entity]
   end
+
+  # NOTE: Use of this method may result in unexpected behavior.  If more than
+  # one ManageIQ class maps to an entity, this method will only return the first
+  # instance.  For example, ContainerProject maps to 'namespace' and 'project'.
+  # This method will only return 'namespace'
+  def entity_by_resource(entity)
+    MAPPING.key(entity)
+  end
 end


cben:
LGTM :+1:
Naming is hard, but I'm not sure I have good suggestions.  @zakiva?

There is discussion about dependencies direction on https://github.com/ManageIQ/manageiq/pull/13619/files#diff-8d2ff9c844b00e0d9e40e56562e23eb7R3 but this is tiny, shouldn't be blocked, and even if we'll end up moving the mapping both directions will stay together.

zakiva:
> LGTM 
> Naming is hard, but I'm not sure I have good suggestions. @zakiva?

I'm OK with `provider_name`, maybe `provider_entity`/`provider_entity_name` or `provider_resource`/`provider_resource_name` also can work. 

Another thing, the values in the mapping are almost unique,  but we have one exception (so far): Both `namespace` and `project` are mapped to `ContainerProject`. Using `:key` on the mapping will return the first match which is not necessarily the name you are looking for. @kevensen I don't know if it's matter for your use case, but as we still have to be more carful here and consider the provider type (k8s/openshift).

kevensen:
@zakiva Fortunately it doesn't matter for my use case, right now.  I get your point and I'll have to think about a good solution.  Are you ok merging this for now?

zakiva:
> @zakiva Fortunately it doesn't matter for my use case, right now. I get your point and I'll have to think about a good solution. Are you ok merging this for now?

@kevensen We'll need at least to add a note to mention this unexpected result. 
cc @simon3z

kevensen:
@zakiva agreed

simon3z:
@zakiva @cben what about:

```ruby
def resource_by_entity(entity)
...

def entity_by_resource(kubernetes_kind)
...
```

or

```ruby
def resource_kind_by_entity(entity)
...

def entity_by_resource_kind(kubernetes_kind)
...
```

zakiva:
@simon3z I'm in favor of the first option: `resource_by_entity`, `entity_by_resource`.
@kevensen you can just change the name of the method you are adding, I will send a refactor for the other method in another PR.

kevensen:
@simon3z @zakiva Works for me!

simon3z:
@zakiva @kevensen @cben who is taking the other renaming? `miq_entity` => `resource_by_entity` ?

zakiva:
> @zakiva @kevensen @cben who is taking the other renaming? miq_entity => resource_by_entity ?

@simon3z I will :)

app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
```diff
@@ -22,4 +22,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EntitiesMapping
   def miq_entity(entity)
     MAPPING[entity]
   end
+
+  # NOTE: Use of this method may result in unexpected behavior.  If more than
+  # one ManageIQ class maps to an entity, this method will only return the first
+  # instance.  For example, ContainerProject maps to 'namespace' and 'project'.
+  # This method will only return 'namespace'
```
moolitayer: Is the behavior described here desirable? would you not prefer to return a list?
Are you expecting to always use 'namespace' and never 'project'.

Also don't you have a need to know if `resource` is from openshift or kubernetes?

kevensen: I'd prefer this not be the behavior but I didn't want to alter the hash for fear of breaking something else.  A more robust solution is preferable.  Do you have a recommendation?

moolitayer: I wasn't thinking of changing the hash, I was thinking on having the reverse behavior return a list - that seems more extensible.

I also expect that we will need to know if resource X is from kubernetes or openshift
(I think currently in some template creation code we try both clients and swallow failures)

I tried to figure out the usage of this by searching in the depending PRs, but could not find any.
Can you provide a link OR a code snippet for your usage of this?
Thanks!

kevensen: I think I understand what you are suggesting.  The following is a modified snippet from a [pull request](https://github.com/kevensen/manageiq/blob/324aa290bb1cf72dc875de095394152e80a832d0/app/models/mixins/container_resource_mixin.rb) that depends on this one.  This method is found in 'app/models/mixins/container_resource_mixin.rb'.
```ruby
  def update_in_provider(resource)
    method = "update_#{entity_by_resource(self.class.name)}"
    api_version = container_project.ext_management_system.api_version
    client = container_project.ext_management_system.connect_client(api_version, method)
    client.send(method, resource)
  end
```

moolitayer: So you don't need to know openshift/kubernetes since connect_client checks which one you need.
For this specific case (ContainerProject) a namespace is created when a project is created and vise versa so I don't think it matters.

kevensen:
All, thanks for your help.

simon3z:
> > @zakiva @kevensen @cben who is taking the other renaming? miq_entity => resource_by_entity ?
> @simon3z I will :)

@zakiva let me know the PR. Thanks.

PR #20, opened 2017-05-21
Add support for reading prometheus endpoint

yaacov:
**Description**
Add support for reading Prometheus endpoint

If prometheus endpoint exist, read metrics from prometeus


yaacov:
@miq-bot add_label compute/containers, wip

cc// @simon3z @cben @moolitayer @zeari

simon3z:
@yaacov let's split this in 3 parts (2 of which we can actually merge):

1. evaluate if we can replace the complex computation of `cpu_usage_rate_average`(based on the cpu time and the number of cores) with the `machine/x/node_utilization`
2. extract the framework about collection that you did here (to generalize hawkular/prometheus)
3. prometheus implementation (as WIP) leveraging the refactor/framework of point 2

yaacov:
> @yaacov let's split this in 3 parts (2 of which we can actually merge):
>
>    - evaluate if we can replace the complex computation of cpu_usage_rate_average(based on the cpu time and the number of cores) with the machine/x/node_utilization
>    - extract the framework about collection that you did here (to generalize hawkular/prometheus)
>    - prometheus implementation (as WIP) leveraging the refactor/framework of point 2

@simon3z @moolitayer 
> extract the framework about collection that you did here (to generalize hawkular/prometheus)

Moved to https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/25

> evaluate if we can replace the complex computation of cpu_usage_rate_average(based on the cpu time and the number of cores) with the machine/x/node_utilization

Implementation depend on #25.

simon3z:
> > evaluate if we can replace the complex computation of cpu_usage_rate_average(based on the cpu time and the number of cores) with the machine/x/node_utilization
> 
> Implementation depend on #25.

@yaacov I'd rather not have the new metrics depending on a refactor if possible. I want to backport them in the cleanest way possible.

yaacov:
@simon3z @cben @moolitayer @zeari please review

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,85 @@
...
+        prometheus_client.get(
+          "query_range",
+          :query => resource,
+          :start => start_sec.to_i,
+          :end   => end_sec.to_i,
+          :step  => "#{@interval}s"
+        )
+      )
+    rescue SystemCallError, SocketError, OpenSSL::SSL::SSLError => e
+      raise CollectionFailure, e.message
```
cben: It's not always clear from exception message which exception it was.  Consider `#{e.class.name}: #{e.message}`.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,85 @@
...
+      sort_and_normalize(
+        prometheus_client.get(
+          "query_range",
+          :query => resource,
+          :start => start_sec.to_i,
+          :end   => end_sec.to_i,
+          :step  => "#{@interval}s"
+        )
+      )
+    rescue SystemCallError, SocketError, OpenSSL::SSL::SSLError => e
```
cben: When would you see `SystemCallError`?  Doesn't Faraday translate these (and SocketError) to something nicer?
What about HTTP errors?
Ideally this knowledge would be wrapped inside client; can wait until we extract a real client gem.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,85 @@
...
+      raise CollectionFailure, e.message
+    end
+
+    def sort_and_normalize(response)
+      JSON.parse(response.body)["data"]["result"][0]["values"].map do |x|
```
cben: can these [] explode?  specifically, can prometheus return errors as json, that wouldn't contain these keys?

I'm fine with approach "if something unexpected happens we'll get exception", but the exception would be hard to diagnose, could be nice to log [truncated] response.body.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,85 @@
...
+        # prometheus gives the time of last reading:
+        # devide and multiply to convert time to start of interval window
+        start_sec = (x[0] / @interval).to_i * @interval.to_i
```
cben: last `.to_i` looks suspicious.  You might be dividing then multiplying by different numbers?  Did you mean
```
        start_sec = ((x[0] / @interval).to_i * @interval).to_i
```

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -0,0 +1,62 @@
...
+        :open_timeout => 2, # opening a connection
+        :timeout      => 5  # waiting for response
+      },
+      :headers => {
+        :Authorization => "Bearer " + credentials[:token]
+      }
+    )
+  end
+
+  # may be nil
```
cben: This was once true for hawkular (and I think Mooli fixed it?).
Let's not repeat it, can we have this code require `connection_configurations.prometheus`?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    URI::HTTPS.build(
+      :host => prometheus_endpoint_empty ? prometheus_default_hostname : prometheus_endpoint.hostname,
+      :port => prometheus_endpoint_empty ? prometheus_default_port : prometheus_endpoint.port,
+      :path => "/api/v1/"
+    )
+  end
+
+  def prometheus_credentials
+    {:token => @ext_management_system.try(:authentication_token)}
```
cben: it's normally the same token, but I think correct place to get it is the "prometheus" Authentication.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  # may be nil
+  def prometheus_endpoint
+    @ext_management_system.connection_configurations.prometheus.try(:endpoint)
+  end
+
+  def prometheus_uri
+    prometheus_default_port = 80
+    prometheus_default_hostname = @ext_management_system.hostname
+    prometheus_endpoint_empty = prometheus_endpoint.try(:hostname).blank?
```
cben: can we drop this empty fallback?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -57,7 +58,11 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
                 "[#{start_time}] [#{end_time}]")
 
       begin
-        context = CaptureContext.new(target, start_time, end_time, INTERVAL)
+        context = if ems && ems.connection_configurations.prometheus.try(:endpoint)
+                    PrometheusCaptureContext.new(target, start_time, end_time, INTERVAL)
+                  else
```
yaacov: > could we have neither?

yes, the hawkular client will fail correctly in this case.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,85 @@
...
+    def fetch_counters_data(resource)
+      start_sec = ((@starts - @interval.to_i) / 1_000)
+      end_sec = (@ends / 1_000)
+
+      sort_and_normalize(
+        prometheus_client.get(
+          "query_range",
+          :query => resource,
+          :start => start_sec.to_i,
+          :end   => end_sec.to_i,
```
yaacov: > should end be rounded-up?

yes, but one second mistake is irrelevant, in this context

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,85 @@
...
+          :step  => "#{@interval}s"
```
yaacov: 👍 YES, i will fix it above in the error (@starts - @interval) thing

simon3z:
@yaacov should we add the prometheus spec tests as well?

yaacov:
> should we add the prometheus spec tests as well?

@simon3z 👍 I will

yaacov:
> @yaacov should we add the prometheus spec tests as well?

@simon3z 👍 specs added

a. specs found error in pod collection, fixed [ labels changed current prometheus deployment ]
b. update error handling to work with both pre 4.0.0 and post 4.0.0. versions of hawkular client [[56]](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/56)

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -57,7 +58,11 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
                 "[#{start_time}] [#{end_time}]")
 
       begin
-        context = CaptureContext.new(target, start_time, end_time, INTERVAL)
+        context = if ems && ems.connection_configurations.prometheus.try(:endpoint)
```
moolitayer: you defined a method called prometheus_endpoint. Can you use that?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
moolitayer: I agree this feels wrong (seems to assume prometheus or hawkular). 
Can you please check if this code is even called when we have no metric endpoint?
if that is the case (should be the case) you can raise an error unless prometheus OR hawkular

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
yaacov: no we can not use it here :-(
it's a chicken and egg problem, to use it we need a context, but if we do not have a context we can't use it :-(

moolitayer: So move it to the manager and the capture context

yaacov:
> Please split it to ease reviewers work - one for refactor related stuff (moving stuff around, renaming etc) and one for new functionality

We already tried that [1] I do not think we should do that again.

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/25

yaacov:
> metrics collection is supposed to be disabled unless we have an endpoint named hawkular

than we should fix the `ems.supports_metrics?` function, we can do that in a new PR.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
yaacov: moving functionality around just for the sake of moving it is pointless, is their a reason for moving it ?

moolitayer: you clearly need this function in a wider scope then the capture context (for example to determine which context to create) then you won't have to copy it in different places.

moolitayer:
> We already tried that [1] I do not think we should do that again.
> 
> [1] #25

I'm not asking that to create extra work for you.
Reviewing many changes takes a lot of time (and I have my own assignments on top of that).
If the changes are clearly separated so that I don't have to figure out what is being copied and what is new functionality that is a lot easier. I promise you that with smaller changes we will get better code quality.

yaacov:
> I'm not asking that to create extra work for you.

@simon3z @moolitayer 

I did not think so :-) it's a matter of time usage and priorities - 
1. we have two other PR's that wait for this one to be merged [1] [2] they are also high priority.
2. this PR depend on Prometheus version and the way Promethean is deployed, for example the pod labelling changed in the last month ( because of new version or relabeling [3] ) . We will have a **lot** of bugs on this new functionalty. we can not wait for the prometheus deployment to be finalized.

New functionality, especially of this size, will come with bugs, and we will fix them when we find them -
1. first will be fixing the [4] `supports_metrics?` bug :-)

[1] https://github.com/ManageIQ/manageiq-ui-classic/pull/1677
[2] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/47
[3] https://github.com/zgalor/scripts/blob/master/openshift/prometheus_template.yaml#L238
[4] https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb#L18

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
yaacov: > Can you please check if this code is even called when we have no metric endpoint?

see discussion below about `supports_metrics?`

> you can raise an error unless prometheus OR hawkular

Why ? the hawkular client will fail correctly in this case.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
yaacov: > (for example to determine which context to create)

you give this case as an example for wider scope ?

yaacov:
> I promise you that with smaller changes we will get better code quality.

Yes, but:

xkcd

yaacov:
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/59 fixes the `supports_metrics?` bug

moolitayer:
@yaacov I've asked to move a few files to a different PR so that review is more effective - I do it all the time. I will not approve this pull request as it is.

yaacov:
> I will not approve this pull request as it is.

Ok :-)

yaacov:
@simon3z as you requested in todays meeting, this PR is closed in favour of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/62, since the Prometheus part depend on #62 being merged, I will post it once #62 is merged.

Issue #19, opened 2017-05-19
Remove kubernetes code from manageiq-gems-pending


chessbyte:
Please extract https://github.com/ManageIQ/manageiq-gems-pending/tree/master/lib/gems/pending/kubernetes to this repo.

simon3z:
@nimrodshn can you look into this? Thanks!

chessbyte:
https://github.com/ManageIQ/manageiq-gems-pending/tree/master/lib/gems/pending has 5 dirs left.  :-)
@simon3z @nimrodshn The sooner we can complete this, the sooner we get to gems-pending being empty!

nimrodshn:
@chessbyte on it :+1:

simon3z:
@nimrodshn @moolitayer is this solved? Can we close?

moolitayer:
Fixed in:
- https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/26
- https://github.com/ManageIQ/manageiq-gems-pending/pull/185/files

```
# checkout master
$ ls lib/gems/pending/
appliance_console  appliance_console.rb  postgres_ha_admin  util
```
Closing

PR #26, opened 2017-05-29
Move kubernetes code from gems-pending

nimrodshn:
Move kubernetes code from gems pending to relevant repo.
Issue #19 
Miq gems PR:https://github.com/ManageIQ/manageiq-gems-pending/pull/185
cc: @cben @simon3z @chessbyte

### DIFF
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,2 +1,3 @@
-# GlobalVars:
-#   AllowedVariables:
+GlobalVars:
+  AllowedVariables:
+  - $kube_log
--- a/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
@@ -14,8 +14,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EventCatcherMixin
   }
 
   def event_monitor_handle
-    require 'kubernetes/events/kubernetes_event_monitor'
-    @event_monitor_handle ||= KubernetesEventMonitor.new(@ems)
+    @event_monitor_handle ||= ManageIQ::Providers::Kubernetes::ContainerManager::KubernetesEventMonitor.new(@ems)
   end
 
   def reset_event_monitor_handle
--- a/app/models/manageiq/providers/kubernetes/container_manager/kubernetes_event_monitor.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/kubernetes_event_monitor.rb
@@ -1,0 +1,39 @@
+class ManageIQ::Providers::Kubernetes::ContainerManager::KubernetesEventMonitor
+  def initialize(ems)
+    @ems = ems
+  end
+
+  def inventory
+    # :service is required to handle also the case where @ems is Openshift
+    @inventory ||= @ems.connect(:service => ManageIQ::Providers::Kubernetes::ContainerManager.ems_type)
+  end
+
+  def watcher(version = nil)
+    @watcher ||= inventory.watch_events(version)
+  end
+
+  def start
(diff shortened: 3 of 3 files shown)


nimrodshn:
@simon3z me and @cben werent sure about where to put this file inside `manageiq-providers-kubernetes`. Thoughts?

moolitayer:
Are there tests that need to move too @nimrodshn

nimrodshn:
@moolitayer nope.

moolitayer:
I don't think belongs under lib in this repository.
I think this belongs in our usual namespace and should sit with other event collection classes

nimrodshn:
@moolitayer ofcourse.
@cben and I thought we should first move it under `/lib` - corresponding to the original path, and then move it to `app/models` etc in a different PR thus making it easier to follow the history of the repo split.
If you think it's better to do it at once I'll do it. @cben Thoughts?

moolitayer:
If we are moving this into this repo I prefer to move it into our namespace and in the same PR switch the collection code to look at that namespace. Lets start by investigating how other providers handled this issue

nimrodshn:
@moolitayer here's what they did in manageiq-providers-ovirt: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/9

moolitayer:
ok thanks @nimrodshn I've looked at https://github.com/ManageIQ/manageiq-providers-ovirt/pull/9/files#r111120109 I'm still in favor of moving into moving to our namespace.

Deferring namespace question to @simon3z

Of curse we will need to carefully test this anyway (Run all the tests in manageiq/manageiq on top of this PR and ManageIQ/manageiq-gems-pending#185 to make sure we don't break manageiq master like these changes often do)

simon3z:
> ok thanks @nimrodshn I've looked at https://github.com/ManageIQ/manageiq-providers-ovirt/pull/9/files#r111120109 I'm still in favor of moving into moving to our namespace.
> 
> Deferring namespace question to @simon3z

@moolitayer @nimrodshn in oVirt it seems that they moved itto their namespace, isn't it? (`OvirtEventMonitor` => `ManageIQ::Providers::Ovirt::Legacy::EventMonitor`).
I am in favor of that.

moolitayer:
> @moolitayer @nimrodshn in oVirt it seems that they moved itto their namespace, isn't it? (OvirtEventMonitor => ManageIQ::Providers::Ovirt::Legacy::EventMonitor).
> I am in favor of that.

:+1: for the class namespace we need to do adapt.

 I was looking at the the fact that they put the class under `lib/` directory. According to the discussion [there](https://github.com/ManageIQ/manageiq-providers-ovirt/pull/9/files#r111119033) we should not have the same problem and should be able to move the class under our regular directory structure as well.

@nimrodshn Let me know if there is any problem with:
- Moving the class to our namespace
- Moving the file under a directory to fit that namespace

Thanks

nimrodshn:
@moolitayer PTAL .

moolitayer:
@nimrodshn Looks good, lets see if you can move the global variable ignores from the main repo as discussed

nimrodshn:
@moolitayer PTAL again :)

moolitayer:
@nimrodshn did you get a chance to test events collection on top of this and the ManageIQ pr?

nimrodshn:
@moolitayer nope :crying_cat_face: 
Show me how to do this over PM?

moolitayer:
> @moolitayer nope 
> Show me how to do this over PM, plz.

Sure come find me when you have some free time.

chessbyte:
@nimrodshn @moolitayer bump

moolitayer:
Waiting for @nimrodshn to test this

nimrodshn:
Tested this along side the manageiq-gems-pending PR and works fine ! :+1: 
@moolitayer  @chessbyte

nimrodshn:
Hooray! :tada:

PR #29, opened 2017-05-31
Add Dependency on kubeclient gem

chessbyte:
Moving from https://github.com/ManageIQ/manageiq-gems-pending/blob/master/manageiq-gems-pending.gemspec


simon3z:
@chessbyte sorry, but I guess that in any merge order these two PR would have conflicted. Needs rebase :worried:

chessbyte:
@simon3z Rebased and re-pushed

PR #30, opened 2017-06-01
Configurable graph refresh

agrare:
Builds on (rips-off? :smile:) @cben's work in https://github.com/ManageIQ/manageiq/pull/14337
Adds a configuration setting to turn on/off graph refresh and don't require building the hashes the old way before building the inventory collections.


app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -0,0 +1,226 @@
...
+        :model_class => ComputerSystem,
+        :parent => ems,
+        :association => :container_node_computer_systems,
+        :manager_ref => [:managed_entity],
+      )
+    @inv_collections[:container_node_computer_system_hardwares] =
+      ::ManagerRefresh::InventoryCollection.new(
+        :model_class => Hardware,
+        :parent => ems,
+        :association => :container_node_computer_system_hardwares,
```
cben: given `association`, can drop `arel`, right?
(my claim here "can't nest has_many through" is wrong, I later discovered the through can point to another has_many that's itself through...)

cben:
BTW, we'll benefit from extra tests https://github.com/ManageIQ/manageiq-providers-openshift/pull/18 (when they work, currently red).

agrare:
> Where are the added model relationships? Is there another PR you're testing against?

I was using your model changes from https://github.com/ManageIQ/manageiq/pull/14337, basically took everything you had as is but dropped any changes in `app/models/ems_refresh/save_inventory_container.rb`

> What's your plan re kubernetes / openshift split? Finish & merge kubernetes (behind an option that won't work with openshift), then do same in openshift repo? Or work on parallel PRs?

Yeah I think that's a good plan, let's get your model changes in first, then get kubernetes working, then openshift.

agrare:
> BTW, we'll benefit from extra tests ManageIQ/manageiq-providers-openshift#18 (when they work, currently red).

:heart:

agrare:
@cben it looks like the tests are failing because the `capture_context_spec` is going through the openshift refresher https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_spec.rb#L14 and those args haven't been updated yet.

Should I change that spec to be use `:ems_kubernetes` instead of `:ems_openshift`?

agrare:
There are some collections that are not implemented yet, including:

- ContainerImages
- ContainerImageRegistries
- CustomAttributes
- ContainerConditions

I marked the test for graph refresh as pending so we can work through these in follow-up PRs

cben:
A spec in kubernetes repo ought not to use `:ems_openshift` factory, but I don't think we want to test only `:ems_kubernetes`.  Ideally should test both, in real life metrics are only collected on openshift but we like pretending that we do support them on kubernetes (if you Hawkular for some reason)...
Anyway, just using `:ems_kubernetes` doesn't work, fails with tons of unused HTTP interactions.

This test does pass on master with `:ems_openshift`.  I'll try to understand why it now fails.

agrare:
> This test does pass on master with :ems_openshift. I'll try to understand why it now fails.

It is because I added `ems` to the args for `ems_inv_to_hashes` here https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/30/files#diff-0324981fdb3019ce6d98f9c86d97f2bbR9

cben:
Oh, I see.  This would require a matching signature change in openshift repo, and running openshift code from this repo makes it impossible to pass Travis for either PR :-(  On the upside, it gave us early warning.

I'm opening an issue to disentangle this test; 
meanwhile what do you think of bdc0310e9e59c5c328e6531fa7ab805f69f7fe17 ?
I'm not sure if it's elegant.  `ems_inv_to_hashes` really should produce ems-independent hashes, so not taking `ems` makes some sense.  Setting ems-dependent ivars in instance method `parser.ems_inv_to_inv_collections()` is fishy — but this parser object is "one-shot", not designed for repeated parsing anyway.

agrare:
> meanwhile what do you think of bdc0310 ?

Looks great, mind if I cherry-pick it here?

agrare:
> I'm not sure if it's elegant. ems_inv_to_hashes really should produce ems-independent hashes, so not taking ems makes some sense. Setting ems-dependent ivars in instance method parser.ems_inv_to_inv_collections() is fishy — but this parser object is "one-shot", not designed for repeated parsing anyway.

Yeah maybe if we create a subclass for the graph refresh parser we can set the inventory collections in that initializer.  We could re-define the e.g.: `get_nodes` methods but keep using the existing e.g.:  `parse_node`.  Might do a follow-up PR with that to see if it is cleaner.

agrare:
@miq-bot remove-label wip

cben:
@enoodle @zeari This is still incomplete (see https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/30#issuecomment-306212702), but can you take a look?

zeari:
is the parent always `ems`? i might not understand this very well, but shouldnt the parent of `container_service_port_configs` be `container_service`?

cben:
Yeah it's confusing.  It's not what we think of semantically as parent.
`parent` is just the object from which to take `association` to define the set which this inv. collection is refreshing.
Things in this set which we don't supply get deleted, things we supply not it the set get created, in both get updated.

Ladas:
This looks great, since this is hidden behind a configuration now, I would just merge and continue adding new PRs, so it's easier to review.

cben:
I got a couple commits on `cben/graph_refresh` branch that fix images & registries:
https://github.com/agrare/manageiq-providers-kubernetes/compare/graph_refresh...cben:graph_refresh
but I hit a wall with relations e.g. `container.container_node` is nil.
This is because this PR skips the most `@data_index.store_path`, but linking relies on being able to `@data_index.fetch_path`.  I agree we don't want to keep relying on order between parse functions, I'm working on a bigger refactoring to allow graph refresh to plug `lazy_find`s instead.

Meanwhile I second Ladislav's suggestion to merge this and do followup PRs.
@enoodle @moolitayer @simon3z, we don't normally merge non-working code, but this is inactive behind a setting, what do you say?

moolitayer:
@agrare do you plan to address any of the code climate issues?
Is this PR dependent on any other current work or not?

@cben @Ladas can you give this a final review?

Handing over to @simon3z as I'll be on PTO for the rest of the week

agrare:
@moolitayer all of the code climate issues are similar code for the inventory collection definitions.  These will be moved out into an inventory persister class soon which will reduce the duplication so I don't plan on addressing these here.

agrare:
ping @simon3z

agrare:
> This is because this PR skips the most @data_index.store_path, but linking relies on being able to @data_index.fetch_path

I think we want to be using `lazy_find` wherever we can instead of store + fetch on big hash, @Ladas ?

Ladas:
@agrare indeed, ```store + fetch on big hash``` is not good when we head to stream/batches being sent. @cben is there anything I can do to help? :-)

cben:
Code-wise, I'm now trying to lift `@data_index` reads as much as possible out of `parse_foo` functions to `get_foos` methods.  Then `get_foos_graph` can do the linking differently.

I've hit a worse problem: much of the references in k8s api responses are by (momentarily unique) "namespace" and "name" fields, not eternally unique uuid.
Can I `lazy_find` by fields other than the collection's `manager_ref`?

Using names as collection `manager_ref` is I think out of the question, would risk IMO scarier records<->objects confusions.  We have to accept some risk of *relations* pointing to wrong object, but those should get fixed in following refreshes.

Ladas:
@cben we can lazy_find only on manager_ref now, but it should not be hard to add additional indexes

I am thinking interface would be something like
```ruby
InventoryCollection.new(:secondary_indexes => {:my_index => [:name, :label]})

# then
objects.lazy_find("#{name}_#{label}", :index => :my_index )
# or
objects.lazy_find([name, label], :index => :my_index )
```

by default it would return 1st match, we could have multiples

simon3z:
@cben can you review/approve if you're OK with merging this now (and refine with other PRs going forward).

agrare:
> Please update :association or :arel for InventoryCollection:<ContainerProject> to return a DISTINCT result

Hm that's a new one I wasn't seeing that before, yeah agreed good for a follow-up PR

Ladas:
@agrare @cben ```Please update :association or :arel for InventoryCollection: to return a DISTINCT result``` is based on the data you have. But yeah, the relation itself needs to be fixed similar to https://github.com/Ladas/manageiq/blob/3a37928b285917325d8b08bc4db3e14470e19687/app/models/container_group.rb#L17-L17

PR #31, opened 2017-06-01
Rename miq_entity method

zakiva:
As discussed in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/18

@miq-bot add_label refactoring

@simon3z @moolitayer Please review

cc @kevensen

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb
@@ -19,7 +19,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EntitiesMapping
     'image'                   => 'ContainerImage'
   }.freeze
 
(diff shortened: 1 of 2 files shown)


moolitayer:
Looks good. @zakiva I've noticed we have also https://github.com/manageiq/manageiq/blob/e7554695bb3030e997cae4afe208df31ee96db5d/app/models/container_template.rb#L16-L40 Do we need that mapping as well ?

zakiva:
> Looks good. @zakiva I've noticed we have also https://github.com/manageiq/manageiq/blob/e7554695bb3030e997cae4afe208df31ee96db5d/app/models/container_template.rb#L16-L40 Do we need that mapping as well ?

@moolitayer No, we don't need it anymore, I'm removing it in https://github.com/ManageIQ/manageiq/pull/13619 (where there is still a question regarding the dependency direction)

PR #32, opened 2017-06-05
[POC] Prometheus event collection

moolitayer:
- Prometheus recently removed resolved events from alert-manager api[1]. They also mention that the alerts api is not final and might change. In order to understand resolved alerts we would have to deduce the resolved ones from the active ones and update them during collection.
- Another possible direction to look at is pushing the alerts into ManageIQ using alert managers web hook plugin, but there is currently no endpoint in ManageIQ (?) and as mentioned by @simon3z means we have to keep ManageIQ auth info in alert manager.
- There is currently no support for alerts in prometheus ruby client 
- Alerts are currently exposed in alert manager and not Prometheus, but there is a PR[TBD] to expos it 
- There does not seem to be a mechanism to filter seen alerts but since we are getting only active alerts that is expected to be a small set.

[1] https://github.com/prometheus/alertmanager/pull/820

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
@@ -1,21 +1,48 @@
 module ManageIQ::Providers::Kubernetes::ContainerManager::EventCatcherMixin
   extend ActiveSupport::Concern
 
-  # https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/container/event.go
-  # 'Created', 'Failed', 'Started', 'Killing', 'Stopped' and 'Unhealthy' are in fact container related events,
-  # returned as part of a pod event.
-  ENABLED_EVENTS = {
-    'Node'                  => %w(NodeReady NodeNotReady Rebooted NodeSchedulable NodeNotSchedulable InvalidDiskCapacity
-                                  FailedMount),
-    'Pod'                   => %w(Scheduled FailedScheduling FailedValidation HostPortConflict DeadlineExceeded
-                                  OutOfDisk NodeSelectorMismatching InsufficientFreeCPU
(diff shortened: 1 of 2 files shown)


simon3z:
@moolitayer please move this out of the Kubernetes event catcher. Maybe @agrare has a suggestion on where to move this (provided that you explain the background). It should belong to a manager probably (or another catcher).

agrare:
@moolitayer I'm not that familiar with Prometheus so maybe you can help explain how it is going to work?
I've heard Prometheus talked about as a metrics backend but this looks like an event endpoint.  Are you thinking that Prometheus will replace the existing event system, augment it?  How does this fit with the metrics work that is going on?

moolitayer:
@agrare you can take a look [here](http://metrics-vm-48-45.eng.lab.tlv.redhat.com/alerts) for prometheus alert ui and [here](https://github.com/moolitayer/scripts-public/blob/master/ansible/prometheus.yaml#L176-L182) for an alert definitions (based on metrics that prometheus knows to collect with its k8s collector)

This should not come instead of our kubernetes event collection. It's only here as POC since we are discussing making our container provider have multiple managers and that will probably take some time.
I think (?) that is what @simon3z mentioned we need help with. We probably need a provider class and have some relationship between container manager and a monitoring manager. I think we might end up having something similar to [foreman](https://github.com/ManageIQ/manageiq-providers-foreman/tree/1a9257535c6b60183a07141dc08996411f263e7a/app/models/manageiq/providers/foreman)

This also relates to our work on metrics; If we are moving away garbing short term metrics into Postgres we would have to off load alerts based on metrics to prometheus / hawkular

agrare:
@moolitayer Ah okay so it sounds like you want a `MetricsManager` and that would have its own `EventCatcher`
You'll have to create `app/models/managieq/providers/metrics_manager` which inherits from `ManageIQ::Providers::BaseManager`

moolitayer:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/40

PR #36, opened 2017-06-07
[WIP] Add prometheus verification

nimrodshn:
Add Prometheus verification when adding a new provider with a Prometheus Endpoint.

cc: @simon3z  @moolitayer @yaacov 
related to: https://github.com/ManageIQ/manageiq-ui-classic/pull/1501

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -61,6 +61,10 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     client.hawkular_try_connect
   end
 
+  def verify_prometheus_credentials
+    true
+  end
+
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -128,6 +132,8 @@ def verify_credentials(auth_type = nil, options = {})
     options = options.merge(:auth_type => auth_type)
     if options[:auth_type].to_s == "hawkular"
       verify_hawkular_credentials
+    if options[:auth_type].to_s == "prometheus"
```
yaacov: Should be elsif

Note to future:
The if `:auth_type == 'hawkular'` was a hack, we should fix the 'hawkular' case, not add new ones :-(

nimrodshn:
@yaacov fixed. thnx mate! :cat2:

PR #38, opened 2017-06-12
Add kubernetes monitoring manager

moolitayer:
This is the second item of ManageIQ/manageiq-providers-kubernetes#39

Depends on: 
- [merged] https://github.com/ManageIQ/manageiq/pull/15354
- [merged] https://github.com/ManageIQ/manageiq/pull/15604
- [merged] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/68 (fix tests)
- [merged] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/72 (fix tests)

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -15,6 +15,13 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
 
   include ManageIQ::Providers::Kubernetes::ContainerManagerMixin
 
+  # See HasMonitoringManagerMixin
+  has_one :monitoring_manager,
+          :foreign_key => :parent_ems_id,
+          :class_name  => "ManageIQ::Providers::Kubernetes::MonitoringManager",
+          :autosave    => true,
+          :dependent   => :destroy
+
   # This is the API version that we use and support throughout the entire code
   # (parsers, events, etc.). It should be explicitly selected here and not
   # decided by the user nor out of control in the defaults of kubeclient gem
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager.rb
@@ -1,0 +1,18 @@
+module ManageIQ::Providers
+  class Kubernetes::MonitoringManager < ManageIQ::Providers::MonitoringManager
+    include ManageIQ::Providers::Kubernetes::MonitoringManagerMixin
+
+    belongs_to :parent_manager,
+               :foreign_key => :parent_ems_id,
+               :class_name  => "ManageIQ::Providers::Kubernetes::ContainerManager",
+               :inverse_of  => :monitoring_manager
+
+    def self.ems_type
+      @ems_type ||= "kubernetes_monitor".freeze
+    end
+
+    def self.description
+      @description ||= "Kubernetes Monitor".freeze
+    end
+  end
+end
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
@@ -1,0 +1,69 @@
+module ManageIQ::Providers::Kubernetes::MonitoringManagerMixin
+  extend ActiveSupport::Concern
+  ENDPOINT_ROLE = :prometheus_alerts
+  DEFAULT_PORT = 9093
+  included do
+    delegate :authentications,
+             :endpoints,
+             :to => :parent_manager
+
+    default_value_for :port do |manager|
+      manager.port || DEFAULT_PORT
+    end
+  end
+
+  module ClassMethods
+    def raw_connect(hostname, port, options)
+      ManageIQ::Providers::Kubernetes::Prometheus::MessageBufferClient.new(hostname, port).connect(options)
+    end
+  end
+
+  def default_endpoint
+    endpoints.detect { |x| x.role == ENDPOINT_ROLE.to_s }
+  end
+
+  def supports_port?
+    true
+  end
+
+  # Authentication related methods, see AuthenticationMixin
+  def authentications_to_validate
+    [ENDPOINT_ROLE]
+  end
+
+  def required_credential_fields(_type)
+    [:auth_key]
+  end
+
+  def default_authentication_type
+    ENDPOINT_ROLE
+  end
+
+  def verify_credentials(auth_type = nil, options = {})
+    with_provider_connection(options.merge(:auth_type => auth_type)) do |conn|
+      # TODO: move to a client method, once we have one
+      conn.get.body.key?('generationID')
+    end
+  rescue Faraday::ClientError => err
+    raise MiqException::MiqUnreachableError, err.message, err.backtrace
+  end
+
+  def connect(options = {})
+    self.class.raw_connect(
+      options[:hostname] || hostname,
+      options[:port] || port,
+      :bearer     => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :verify_ssl => options[:verify_ssl] || verify_ssl,
+      :cert_store => options[:cert_store] || ssl_cert_store
+    )
+  end
+
+  def ssl_cert_store
+    # nil === use system CA bundle
+    default_endpoint.try(:ssl_cert_store)
+  end
+
+  def verify_ssl
+    default_endpoint.verify_ssl?
+  end
+end
--- a/app/models/manageiq/providers/kubernetes/prometheus/message_buffer_client.rb
+++ b/app/models/manageiq/providers/kubernetes/prometheus/message_buffer_client.rb
@@ -1,0 +1,33 @@
+class ManageIQ::Providers::Kubernetes::Prometheus::MessageBufferClient
+  require 'faraday'
+
+  def initialize(host, port)
+    @host = host
+    @port = port
+  end
+
+  def alert_url
+    "https://#{@host}:#{@port}/topics/alerts"
+  end
+
+  def connect(client_options)
+    settings = ::Settings.ems.ems_kubernetes.ems_monitoring.alerts_collection
+    Faraday.new(
+      :url     => alert_url,
+      :headers => {'Authorization' => "Bearer #{client_options[:bearer]}"},
+      :request => {
+        # opening a connection
+        :open_timeout => settings.open_timeout.to_f_with_method,
+        # waiting for response
+        :timeout      => settings.timeout.to_f_with_method,
+      },
+      :ssl     => {
+        :verify     => client_options[:verify_ssl],
+        :cert_store => client_options[:cert_store],
+      },
+    ) do |conn|
+      conn.response(:json)
+      conn.adapter(Faraday.default_adapter)
+    end
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,6 +1,10 @@
 ---
 :ems:
   :ems_kubernetes:
+    :ems_monitoring:
+      :alerts_collection:
+        :open_timeout: 10.seconds
+        :timeout: 20.seconds
     :miq_namespace: management-infra
     :image_inspector_registry: docker.io
     :image_inspector_repository: openshift/image-inspector
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,3 +1,4 @@
 GlobalVars:
   AllowedVariables:
   - $kube_log
+  - $cn_monitoring_log
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -14,6 +14,15 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
       # checking `provider.port` first prevents this from overriding it.
       provider.port || provider.class::DEFAULT_PORT
     end
+
+    before_create :ensure_monitoring_manager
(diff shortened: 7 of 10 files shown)


moolitayer:
@simon3z @blomquisg @Ladas please take a look

Ladas:
we should fix the rubocop issues, then this looks good 👍

Ladas:
looks great, just fix the rubocop issues (took me a while to figure out that the k8s mixin is included here :-))

moolitayer:
Thanks @Ladas I'll probably have a few more changes to this PR once I finish the event collection that uses the new managers.

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -15,6 +15,13 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
 
   include ManageIQ::Providers::Kubernetes::ContainerManagerMixin
 
+  # Override HasMonitoringManagerMixin
```
cben: no longer an override, as you dropped this from base class?

app/models/manageiq/providers/kubernetes/monitoring_manager.rb
```diff
@@ -0,0 +1,19 @@
+module ManageIQ::Providers
+  class Kubernetes::MonitoringManager < ManageIQ::Providers::MonitoringManager
+    include ManageIQ::Providers::Kubernetes::MonitoringManagerMixin
+
+    belongs_to :parent_manager,
+               :foreign_key => :parent_ems_id,
+               :class_name  => "ManageIQ::Providers::Kubernetes::ContainerManager",
+               :autosave    => true,
```
cben: Are there / do we have conventions about autosave?
Intuitively, I'd expect parent to autosave the child but not vice-versa.
I have no idea what I'm talking about :-)

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,52 @@
+require ''
```
cben: require what?

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,52 @@
...
+  def connect(options)
+    raise ArgumentError, "Unsupported service [#{service}]" unless options[:service] == :alerts
+    ManageIQ::Providers::Kubernetes::Prometheus::Client.new(
+        alerts_url,
+        {
+          # :http_proxy_uri => VMDB::Util.http_proxy_uri.to_s,
+          # :verify_ssl     => @ext_management_system.verify_ssl_mode(prometheus_endpoint),
+          # :ssl_cert_store => @ext_management_system.ssl_cert_store(prometheus_endpoint),
+        }
+    ).connect(options[:service])
```
cben: Service is keyword are here, positional in client's `connect`.  Can it consistently be one style in both?

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,52 @@
...
+             :authentication_status_ok?,
+             :authentications,
+             :authentication_for_summary,
+             :zone,
+             :verify_credentials,
+             :address,
+             :ip_address,
+             :hostname,
+             :default_endpoint,
+             :endpoints,
```
cben: please sort (unless you have logic behind the order)

app/models/manageiq/providers/kubernetes/prometheus/client.rb
```diff
@@ -0,0 +1,34 @@
...
+    #    :open_timeout => 2, # opening a connection
+    #    :timeout      => 5  # waiting for response
+    #  }
+    # )
+    raise ArgumentError, "Unsupported service [#{service}]" unless service == :alerts
+    Faraday.new(
+        :url     => @alerts_uri.to_s,
+        :request => {
+            :open_timeout => 2, # opening a connection
+            :timeout      => 5  # waiting for response
```
cben: 2 then 5 seconds will not be enough in some deployments.
Experience suggests these should be settings.
[can be done later]

app/models/manageiq/providers/kubernetes/monitoring_manager.rb
moolitayer:
```
:autosave
If true, always save the associated object or destroy it if marked for destruction, when saving the parent object. If false, never save or destroy the associated object. By default, only save the associated object if it's a new record.
```

`:dependent => :destroy` on steroids.

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
moolitayer: forgot to push

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
moolitayer: connect should be a hash in connect. All other parameters to the client are passed in the constructor.
The idea is to change this to a proper client when we have one.

app/models/manageiq/providers/kubernetes/prometheus/client.rb
moolitayer: sure, I'll move to settings (do you have a suggestion for better defaults?)

moolitayer:
@Ladas @cben PR changed as a result of adding authentication. Please take a look

config/settings.yml
```diff
@@ -41,6 +41,10 @@
             - POD_HOSTPORTCONFLICT
     :open_timeout: 60.seconds
     :read_timeout: 60.seconds
+  :ems_monitoring:
```
durandom: do you expect these to be global or only relevant to kubernetes?
from where they are used, they seem to be even prometheus specific.
I suggest nesting them under ems_kubernetes

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,72 @@
...
+  rescue Faraday::ClientError => err
+    raise MiqException::MiqUnreachableError, err.message, err.backtrace
+  end
+
+  def connect(options)
+    ManageIQ::Providers::Kubernetes::MonitoringManagerMixin.raw_connect(
+      options[:hostname] || hostname,
+      options[:port] || port,
+      :bearer     => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :ssl_verify => options[:ssl_verify] || ssl_verify,
```
cben: consider s/ssl_verify/verify_ssl/, as many other places use the wording "verify_ssl".
though I can see how putting ssl first for consistency can make sense.

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,72 @@
...
+  def connection_configurations
+    parent_manager.connection_configurations[ENDPOINT_ROLE] || []
+  end
+
+  def endpoints
+    parent_manager.endpoints.where(:role => ENDPOINT_ROLE)
+  end
+
+  def default_endpoint
+    endpoints.first
```
cben: Above methods redefine a bunch of EMS functionality to do something different.
E.g. there is an EMS method `default_endpoint` that looks for an endpoint with role `default`.

I can see how you want to start with something working, without embarking on migrations etc. but I think this should not the end goal.

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def supports_port?
+    true
+  end
+
+  # Authentication related methods, see AuthenticationMixin
```
cben: AuthenticationMixin is pretty complicated, combined with all these dont-mean-what-I-think-they-mean methods, scares me :-|

Is there a way to have AuthenticationMixin work on the parent?
If yes that might be simpler, though I'm afraid it's not possible as parent won't understand our auth_type etc.

@blomquisg While what we're trying here is an abuse, one way or the other, I think this is a symptom of AuthenticationMixin–EMS relationship being unnecessary coupling, it taking the auth_type params everywhere and rummaging through the EMS's associations... 
AuthenticationMixin methods IMO should have directly taken Authentication (and/or Endpoint) params, and possibly not been a mixin to EMS at all.

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -0,0 +1,72 @@
...
+      options[:port] || port,
+      :bearer     => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :ssl_verify => options[:ssl_verify] || ssl_verify,
+      :cert_store => options[:cert_store] || ssl_cert_store
+    )
+  end
+
+  def ssl_cert_store
+    # nil === use system CA bundle
+    default_endpoint.try(:ssl_cert_store)
```
cben: Concrete suggestion: since you know "default_endpoint" doesn't return "default" role, minimize its use.
I'd write `endpoints.first` or something even closer to the metal here.

app/models/manageiq/providers/kubernetes/monitoring_manager.rb
moolitayer: oh I get it now, I didn't realize I was saving from both directions
Thanks @cben

moolitayer:
@enoodle  @cben @ilackarms please review

I'm adding some tests

moolitayer:
@simon3z please review

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
moolitayer: As discussed this is intentional. The new manager should check authentication and only if the authentication is valid start it's workers.

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
moolitayer: default_endpoint is intentionally used as the default endpoint for this manager.

moolitayer:
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/68 Should fix the failing tests

app/models/manageiq/providers/kubernetes/prometheus/message_buffer_client.rb
```diff
@@ -0,0 +1,33 @@
...
+        # opening a connection
+        :open_timeout => settings.open_timeout.to_f_with_method,
+        # waiting for response
+        :timeout      => settings.timeout.to_f_with_method,
+      },
+      :ssl     => {
+        :verify     => client_options[:verify_ssl],
+        :cert_store => client_options[:cert_store],
+      },
+    ) do |conn|
```
ilackarms: what happens to this connection when it's created?

i see it's consumed in `MonitoringManagerMixin#verify_credentials`. Where else is this consumed?

Messages can be consumed in two ways from message buffer, either:
1. `HTTP GET /api/v1/topics/{topic}?generationID=<genID returned from your first request>&fromIndex=<results array start from this message index>` returns an array of messages, starting from `fromIndex`. Every time you get messages this way, bump the internal counter for `fromIndex` so you don't get repeats of the same message
2. Use a websocket client to dial `/api/v1/topics/{topic}/watch?generationID=<genID>&fromIndex=<index>`. The live connection will watch for new messages (returned individually, not as part of an array). `fromIndex` works the same way here (the first received message will have index == `fromIndex`).

moolitayer: Right, in this PR it's used only for validation and in the next PR for collection[1]. I didn't know the websocket option is already implemented, that's great. I will convert my code to use that.

Let's first get this one in and then move to the collection part.

ilackarms: gotcha!

if you want to do a watch, you'll need a websocket client (although I'm not sure we even need websockets; streaming HTTP might be enough). however feel free to ignore this comment :P

moolitayer: Yeah I'm looking into that. For the validation a synchronous get is preferred.

FYI For some reason not all the messages end up in the buffer with my current deployment - I haven't investigated why yet (could be due to many factors along the way).

moolitayer:
@durandom @Ladas @cben @simon3z review please...
I'd like to get this out of the way to continue to the more interesting stuff (alerts)

Travis will keep  failing until ManageIQ/manageiq#15604 is merged

cben:
LGTM :+1:

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
durandom: > Is there a way to have AuthenticationMixin work on the parent?

you mean on the parent_manager? Looking at the providers that implement the `Provider` -> `Manager` relation, they handle all auth stuff via the provider.

> AuthenticationMixin methods IMO should have directly taken Authentication (and/or Endpoint) params, and possibly not been a mixin to EMS at all.
 
totally agree. This is another good example of the over use (abuse) of mixins in our app.
In case you have time, how about doing a more concise way to implement authentication/enpoints in the kubernetes provider, show that it works and guide the way for others?

simon3z:
@moolitayer can you check why travis is still failing? https://github.com/ManageIQ/manageiq/pull/15604 is now merged.

moolitayer:
Tests didn't actually run

PR #40, opened 2017-06-14
Prometheus alerts

moolitayer:
Issue: https://github.com/ManageIQ/manageiq/issues/14238
See issue for setup

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -77,6 +77,11 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     client.prometheus_try_connect
   end
 
+  def verify_prometheus_alerts_credentials
+    ensure_monitoring_manager
+    monitoring_manager.verify_credentials
+  end
+
   # UI methods for determining availability of fields
   def supports_port?
     true
@@ -131,6 +136,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     at = [:bearer]
     at << :hawkular if has_authentication_type?(:hawkular)
     at << :prometheus if has_authentication_type?(:prometheus)
+    at << :prometheus_alerts if has_authentication_type?(:prometheus_alerts)
     at
   end
 
@@ -148,6 +154,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
       verify_hawkular_credentials
     elsif options[:auth_type].to_s == "prometheus"
       verify_prometheus_credentials
+    elsif options[:auth_type].to_s == "prometheus_alerts"
+      verify_prometheus_alerts_credentials
     else
       with_provider_connection(options, &:api_valid?)
     end
@@ -166,7 +174,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
   end
 
   def supported_auth_types
-    %w(default password bearer hawkular prometheus)
+    %w(default password bearer hawkular prometheus prometheus_alerts)
   end
 
   def supports_authentication?(authtype)
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
@@ -1,0 +1,118 @@
+module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerMixin
+  extend ActiveSupport::Concern
+
+  # This module is shared between:
+  # - Kubernetes::MonitoringManager::EventCatcher
+  # - Openshift::MonitoringManager::EventCatcher
+
+  def event_monitor_handle
+    @event_monitor_handle ||= ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::Stream.new(@ems)
+  end
+
+  def reset_event_monitor_handle
+    @event_monitor_handle = nil
+  end
+
+  def stop_event_monitor
+    @event_monitor_handle.stop unless @event_monitor_handle.nil?
+  rescue => err
+    $cn_monitoring_log.error("Event Monitor error [#{err.message}]")
+    $cn_monitoring_log.error("Error details: [#{err.details}]")
+    $cn_monitoring_log.log_backtrace(err)
+  ensure
+    reset_event_monitor_handle
+  end
+
+  def monitor_events
+    $cn_monitoring_log.info("[#{self.class.name}] Event Monitor started")
+    @target_ems_id = @ems.parent_manager.id
+    event_monitor_handle.start
+    event_monitor_running
+    event_monitor_handle.each_batch do |events|
+      @queue.enq(events) unless events.blank?
+      sleep_poll_normal
+    end
+  ensure
+    reset_event_monitor_handle
+  end
+
+  def queue_event(event)
+    event_hash = extract_event_data(event)
+    if event_hash
+      $cn_monitoring_log.info("Queuing event [#{event_hash}]")
+      EmsEvent.add_queue("add", @target_ems_id, event_hash)
+    end
+  end
+
+  def extract_event_data(event)
+    # EXAMPLE:
+    #
+    # {
+    #     "annotations": {
+    #         "message": "Node ocp-compute01.10.35.48.236.nip.io is down",
+    #         "severity": "HIGH",
+    #         "source": "ManageIQ",
+    #         "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
+    #     },
+    #     "endsAt": "0001-01-01T00:00:00Z",
+    #     "generatorURL": "http://prometheus-4018548653-w3str:9090/graph?g0.expr=container_fs_usage_bytes%7Bcontainer_name%3D%22%22%2Cdevice%3D%22%2Fdev%2Fmapper%2Fvg0-lv_root%22%7D+%3E+4e%2B07&g0.tab=0",
+    #     "labels": {
+    #         "alertname": "Node down",
+    #         "beta_kubernetes_io_arch": "amd64",
+    #         "beta_kubernetes_io_os": "linux",
+    #         "device": "/dev/mapper/vg0-lv_root",
+    #         "id": "/",
+    #         "instance": "ocp-compute01.10.35.48.236.nip.io",
+    #         "job": "kubernetes-nodes",
+    #         "kubernetes_io_hostname": "ocp-compute01.10.35.48.236.nip.io",
+    #         "region": "primary",
+    #         "zone": "default"
+    #     },
+    #     "startsAt": "2017-07-17T12:18:00.457154718Z",
+    #     "status": "firing",
+    #     "generationID" : "323e0863-f501-4896-b7dc-353cf863597d", # Added in stream
+    #     "index": 1, # Added in stream
+    # },
+    event = event.dup
+
+    annotations = event["annotations"]
+    event[:url] = annotations["url"]
+    event[:severity] = parse_severity(annotations["severity"])
+    labels = event["labels"]
+    event[:ems_ref] = incident_identifier(event, labels, annotations)
+    event[:resolved] = event["status"] == "resolved"
+    timestamp = event["timestamp"]
+
+    target = find_target(labels)
+    {
+      :ems_id              => @cfg[:ems_id],
+      :source              => "DATAWAREHOUSE",
+      :timestamp           => timestamp,
+      :event_type          => "datawarehouse_alert",
+      :target_type         => target.class.name,
+      :target_id           => target.id,
+      :container_node_id   => target.id,
+      :container_node_name => target.name,
+      :message             => annotations["message"],
+      :full_data           => event.to_h
+    }
+  end
+
+  def find_target(labels)
+    instance = ContainerNode.find_by(:name => labels["instance"], :ems_id => @target_ems_id)
+    $cn_monitoring_log.error("Could not find alert target from labels: [#{labels}]") unless instance
+    instance
+  end
+
+  def parse_severity(severity)
+    MiqAlertStatus::SEVERITY_LEVELS.find { |x| x == severity.to_s.downcase } || "error"
+  end
+
+  def incident_identifier(event, labels, annotations)
+    # When event b resolves event a, they both have the same startAt.
+    # Labels are added to avoid having two incidents starting at the same time.
+    Digest::SHA256.hexdigest(
+      [event["startsAt"], annotations["url"], labels["instance"], labels["alertname"]].join('|')
+    )
+  end
+end
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
@@ -1,0 +1,79 @@
+class ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::Stream
+  def initialize(ems)
+    @ems = ems
+  end
+
+  def start
+    @collecting_events = true
+  end
+
+  def stop
+    @collecting_events = false
+  end
+
+  def each_batch
+    while @collecting_events
+      yield(fetch)
+    end
+  rescue EOFError => err
+    $cn_monitoring_log.info("Monitoring connection closed #{err}")
+  end
+
+  def fetch
+    unless @current_generation
+      @current_generation, @current_index = last_position
+    end
+    $cn_monitoring_log.info("Fetching alerts. Generation: [#{@current_generation}/#{@current_index}]")
+
+    response = @ems.connect.get do |req|
+      req.params['generationID'] = @current_generation
+      req.params['fromIndex'] = @current_index
+    end
+    # {
+    #   "generationID":"323e0863-f501-4896-b7dc-353cf863597d",
+    #   "messages":[
+    #   ...
+    #   ]
+    # }
+    alert_list = response.body
+    alerts = []
+    @current_generation = alert_list["generationID"]
+    return alerts if alert_list['messages'].blank?
+    alert_list["messages"].each do |message|
+      @current_index = message['index']
+      unless message.fetch_path("data", "commonAnnotations", "miqTarget") == 'ContainerNode'
+        $cn_monitoring_log.info("Skipping alert due to missing annotation")
+        next
+      end
+      message["data"]["alerts"].each_with_index do |alert, i|
+        alert['generationID'] = @current_generation
+        alert['index'] = @current_index
+        alert['timestamp'] = timestamp_indent(alert, i)
+        alerts << alert
+      end
+      @current_index += 1
+    end
+    $cn_monitoring_log.info("[#{alerts.size}] new alerts. New generation: [#{@current_generation}/#{@current_index}]")
+    $cn_monitoring_log.debug(alerts)
+    alerts
+  end
+
+  def timestamp_indent(alert, indent)
+    # This is currently needed due to a uniqueness constraint on ems events
+    # see https://github.com/ManageIQ/manageiq/pull/15719
+    # Prometheus alert timestamp equals the evaluation cycle start timestamp
+    # We are adding an artificial indent of the lest significant bit since several alerts
+    # for different entities or from different alert definitions are likely to have the same timestamp
+    timestamp = alert["status"] == 'resolved' ? alert["endsAt"] : alert["startsAt"]
+    Time.zone.at((Time.parse(timestamp).to_f + (0.000001 * indent)))
+  end
+
+  def last_position
+    last_event = @ems.parent_manager.ems_events.last || OpenStruct.new(:full_data => {})
+    last_index = last_event.full_data['index']
+    [
+      last_event.full_data['generationID'].to_s,
+      last_index ? last_index + 1 : 0
+    ]
+  end
+end
--- a/spec/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin_spec.rb
@@ -1,0 +1,141 @@
+describe ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerMixin do
+  subject do
+    test_class.new(monitoring_manager)
+  end
+
+  let(:test_class) do
+    Class.new do
+      def initialize(ems)
+        @ems = ems
+      end
+    end.include(described_class)
+  end
+
+  let(:default_endpoint) { FactoryGirl.create(:endpoint, :role => "default", :hostname => "host") }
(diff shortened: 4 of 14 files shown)


moolitayer:
@cben @zgalor please review

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    event[:severity] = parse_severity(severity)
+    event[:resolved] = event["status"] == 'resolved'
+    event[:ems_ref] = incident_identifier(event["startsAt"], labels)
+    timestamp = event["timestamp"]
+
+    target = find_target(labels)
+    {
+      :ems_id              => @cfg[:ems_id],
+      :source              => 'DATAWAREHOUSE',
```
zgalor: Should this be `MONITORING` ?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,115 @@
...
+      :timestamp           => timestamp,
+      :event_type          => 'datawarehouse_alert',
```
zgalor: see comment from 2 lines above

spec/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  let(:default_endpoint) { FactoryGirl.create(:endpoint, :role => 'default', :hostname => 'host') }
+  let(:default_authentication) { FactoryGirl.create(:authentication, :authtype => 'bearer') }
+  let(:prometheus_alerts_endpoint) do
+    EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+    FactoryGirl.create(
+      :endpoint,
+      :role       => 'prometheus_alerts',
+      :hostname   => 'alerts-prometheus.10.35.48.141.nip.io',
```
zgalor: maybe replace (here and in recordings) to fake hostname and token

spec/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+      :port       => 443,
+      :verify_ssl => false
+    )
+  end
+  let(:prometheus_authentication) do
+    FactoryGirl.create(
+      :authentication,
+      :authtype => 'prometheus_alerts',
+      :auth_key => 'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJtYW5hZ2VtZW50LWluZnJhIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZWNyZXQubmFtZSI6Im1hbmFnZW1lbnQtYWRtaW4tdG9rZW4tZzRscDciLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC5uYW1lIjoibWFuYWdlbWVudC1hZG1pbiIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VydmljZS1hY2NvdW50LnVpZCI6ImQxOWUzMDk4LTZmOGUtMTFlNy04YTQ1LTAwMWE0YTE2MjY4YyIsInN1YiI6InN5c3RlbTpzZXJ2aWNlYWNjb3VudDptYW5hZ2VtZW50LWluZnJhOm1hbmFnZW1lbnQtYWRtaW4ifQ.m4ckxIoTj__rCFA25V8lurlapNgL6doCpP51Z0JA5VgiU9mRxdk8KGyCe6w_mhrh5xPO9UuiK_pYIIMPQ-lAijRaQrKITZ-smRMEcsqKU0uP10UliJIj7ICGxn7c2zDYQuxMOQwI4ELWPv4xBHVXkF0hBKhVCPWRmdnIwADIfQi4RqoN_1gQf8S2O8C_eDS-A1f5SUf1Bvhexs997NWFBVyKEUQKxdqAlHf59V3Zt2VaE2HUme_ppYvp5L9QSa-CEZU_XxmUwNeCzcgbzts8OMujuHB_8lwKcLy9H0k3_oh-EcOn4CvzVz6pTfh8HPDoFhqpnBHfLB4pnsRkoHbDrA',
```
zgalor: see previous comment

spec/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  let(:default_endpoint) { FactoryGirl.create(:endpoint, :role => 'default', :hostname => 'host') }
+  let(:default_authentication) { FactoryGirl.create(:authentication, :authtype => 'bearer') }
+  let(:prometheus_alerts_endpoint) do
+    EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+    FactoryGirl.create(
+      :endpoint,
+      :role       => 'prometheus_alerts',
+      :hostname   => 'alerts-prometheus.10.35.48.141.nip.io',
```
zgalor: maybe replace (here and in recordings) to fake hostname and token

moolitayer:
@enoodle please review

spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb
```diff
@@ -14,7 +14,7 @@
   let(:prometheus_authentication) do
     FactoryGirl.create(
       :authentication,
-      :authtype => :prometheus_alerts,
+      :authtype => 'prometheus_alerts',
       :auth_key => 'eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJtYW5hZ2VtZW50LWluZnJhIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZWNyZXQubmFtZSI6Im1hbmFnZW1lbnQtYWRtaW4tdG9rZW4tMnBzZjMiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC5uYW1lIjoibWFuYWdlbWVudC1hZG1pbiIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VydmljZS1hY2NvdW50LnVpZCI6IjJiMjg1MmIyLTZjNzAtMTFlNy1hYTlmLTAwMWE0YTE2MjYxOSIsInN1YiI6InN5c3RlbTpzZXJ2aWNlYWNjb3VudDptYW5hZ2VtZW50LWluZnJhOm1hbmFnZW1lbnQtYWRtaW4ifQ.OSh7pgdRXAIUh8hPipfj_me-T3cwI_DsqQUSV1yYo1qvGEd1Aa-oVaBAeKsbwjCDhcNw2nNTWkdeYoy4-MyoTjleZbvdsbtSgs84LCPdLfaVd7NXBiXtNj6o4a2tbcE_GUTBiOEyZfzFh5pwM1n9BbU6qwAvf5B-uiAuGI1VZXyiUWtlspNAEPOy3awFNt9vausNahfMRox9MLB62BYzj2NV36inpNY_UWMNV0X0Q1VnZWO6-v28JAkWhuqaRHgSOUKV1FKJDY6R4rCGxt5BnVS6_81au80vouZmv0oR6kvDPWZo6IVPG8JCIpxK0liJW65pxKIBf7oOkEzhi3wy9w',
```
zgalor: maybe replace (here and in recordings) to fake token

moolitayer:
@ilackarms please review
(BTW since we are blocked on watch this is still using pulling)

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
moolitayer: Unfortunately that would not work as `DATAWAREHOUSE` is our current hook into the event system. I have a [PR](https://github.com/ManageIQ/manageiq/pull/15295/files#diff-edcc6e6ea6f974c08f1323b292309dc6R179) renaming these two fields to `MONITORING/monitoring_alert`. I update here one that has been resolved

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
moolitayer: see previous response

moolitayer:
code climate asks to use safe navigation operator (&.) available since ruby 2.3.
According to travis.yml here and in the main repo, we do not support 2.2 (which is in security maintenance phase)
so that should be fine. I do not see any usages of safe navigation although I would like to start using it. 
cc @jrafanie @cben

moolitayer:
@cben @zgalor @enoodle @simon3z needs your review please

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,115 @@
...
+    #         "region": "primary",
+    #         "zone": "default"
+    #     },
+    #     "startsAt": "2017-07-17T12:18:00.457154718Z",
+    #     "status": "firing"
+    # },
+    event = event.dup
+
+    annotations, labels = event["annotations"], event["labels"]
+    severity, event[:url] = annotations["severity"], annotations["url"]
```
cben: IMHO these are paired randomly and could read better as separate assignments.
- `event[:url]` write is hiding here in middle of line.
- `labels` and its uses far below could be closer together.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    event[:severity] = parse_severity(severity)
+    event[:resolved] = event["status"] == 'resolved'
+    event[:ems_ref] = incident_identifier(event["startsAt"], labels)
```
cben: event will contain a mixture of string and symbol keys.  intentional?
all the symbol keys are added only so that they appear in `full_data`, right?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def parse_severity(severity)
+    MiqAlertStatus::SEVERITY_LEVELS.find { |x| x == severity.to_s.downcase } || 'error'
+  end
+
+  def incident_identifier(incident_start, labels)
+    # When event b resolves event a, they both have the same startAt.
+    # Labels are added to avoid having two incidents starting at the same time.
+    "#{incident_start}_#{labels.hash}"
```
cben: `.hash` is the ruby method returning an integer, right?
It's not stable between ruby runs (randomized seed against DOS)!  And obviously not between ruby versions.
You're persisting it, which sounds bad idea.

Can you use some canonic json / yaml serialization (or its SHA if length is a concern)?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
moolitayer: > all the symbol keys are added only so that they appear in full_data, right?

Right

> event will contain a mixture of string and symbol keys. intentional?

to be taken into account these should be symbols[1], if we care about all the keys having the same type we can call deep_symbolize_keys (or have the client return symbol keyed objects). WDYT @cben?

[1] https://github.com/moolitayer/manageiq/blob/408cc00831ae9fc669a79f6be002f8e6effc264f/app/models/ems_event.rb#L177-L183

jrafanie:
> code climate asks to use safe navigation operator (&.) available since ruby 2.3.
> According to travis.yml here and in the main repo, we do not support 2.2 (which is in security maintenance phase) so that should be fine. I do not see any current usages of safe navigation although I would like to start using it.
> cc @jrafanie @cben

@moolitayer we can't yet because we haven't officially dropped 2.2.

I haven't gotten getting this [ruby 2.4 issue done to completion](https://github.com/ManageIQ/manageiq/issues/14446).  Once that is complete, we'd drop support for 2.2 and only support 2.3 and 2.4.  If you have feedback on which gems are complete from that list, or ones that are missing, I could use help making sure our gem ecosystem works on 2.4.

moolitayer:
@cben @enoodle @zgalor please take another look
@durandom could you please take a look?

ilackarms:
@moolitayer I don't see anything here specific to prometheus-alert-buffer. do you want to work on implementing https://github.com/openshift/prometheus-alert-buffer/issues/1? if not i can submit it as  a pr to unblock this for you

moolitayer:
> @moolitayer we can't yet because we haven't officially dropped 2.2.

ahh I see, thanks.

> I haven't gotten getting this ruby 2.4 issue done to completion. Once that is complete, we'd drop support for 2.2 and only support 2.3 and 2.4. If you have feedback on which gems are complete from that list, or ones that are missing, I could use help making sure our gem ecosystem works on 2.4.

Sure. All the ones we maintain seem fine now. I will make sure we test for 2.4 in all our repos. Let me know if there is anything else we can do to help

moolitayer:
> @moolitayer I don't see anything here specific to prometheus-alert-buffer. do you want to work on implementing openshift/prometheus-alert-buffer#1?

Yeah, sure @ilackarms thanks

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def fetch
+    unless @current_generation
+      @current_generation, @current_index = last_position
+    end
+
+    response = @ems.connect.get do |req|
+      req.params['generationID'] = @current_generation
+      req.params['fromIndex'] = @current_index
```
moolitayer: Every time prometheus-message-buffer is started a generation starts. While it is up the generation will not change and the index will increment for new messages. If you request for a previous generation fromIndex is ignored and you get all messages. If you request for the current generation you will get all messages with index > fromIndex. If there are no such messages you get this empty response:
```
{"generationID":"95b45589-ebee-49ca-a1bc-978c401dfde0","messages":[]}
```

moolitayer: Oh also, currently we are using polling but once we implement https://github.com/openshift/prometheus-alert-buffer/issues/1 we will move to an asynchronous get message when available

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,69 @@
...
+    alert_list = response.body
+    return if alert_list['messages'].blank?
+    @current_generation = alert_list["generationID"]
+    @current_index = alert_list['messages'].last['index'] + 1
+    alerts = []
+    alert_list["messages"].each do |message|
+      next unless message.fetch_path("data", "commonAnnotations", "miqTarget") == 'ContainerNode'
+      message["data"]["alerts"].each_with_index do |alert, i|
+        alert['generationID'] = @current_generation
+        alert['index'] = @current_index
```
moolitayer: right I'll change to the current index for each alert. I guess currently queue == db, but it does not have to be that way forever. Your second example we would queue the event twice and only the first one would be added, since we have a uniqueness check - do we prefer to add it twice and not take a change in not adding it once (like the case you described in `1.`)

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+    end
+    $cn_monitoring_log.info("[#{alerts.size}] new alerts. Generation: [#{@current_generation}/#{@current_index}]")
+    $cn_monitoring_log.debug(alerts)
+    alerts
+  end
+
+  def timestamp_indent(alert, indent)
+    timestamp = alert["status"] == 'resolved' ? alert["endsAt"] : alert["startsAt"]
+    Time.zone.at((Time.parse(timestamp).to_f + (0.000001 * indent)))
```
moolitayer: The resolution is micro seconds. The list is nested
- alert type (e.g high node mem usage)
  - label combination (e.g high node mem usage in node host1.example.com)
  - label combination (e.g high node mem usage in node host2.example.com)
- alert type
...

for alerts items to be on the same level they need to be not only originating from the same alert definition, but they have to come from the same evaluation cycle. In that case they all have the same start time.

There is currently no way in ManageIQ to have two alerts (fro the same provider) with the same timestamp and that is why I did this ugly hack. We got a suggestion of a way to change that so this ugly code should go away now, but I still need to think about it some more..

enoodle:
:+1:  LGTM as much as I can understand. Nice amount of tests :cake:

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
moolitayer: So we have https://github.com/ManageIQ/manageiq/pull/15719
Unfortunately I could not yet remove this, since there is another problem not addressed there.
I will add a comment explaining the issues here.

moolitayer:
@agrare could you please review? this is replacing https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/32

moolitayer:
@zgalor there is currently an active change request from you. Please remove that if your concerns have been addressed.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,118 @@
...
+    event[:severity] = parse_severity(annotations["severity"])
+    labels = event["labels"]
+    event[:ems_ref] = incident_identifier(event, labels, annotations)
+    event[:resolved] = event["status"] == "resolved"
+    timestamp = event["timestamp"]
+
+    target = find_target(labels)
+    {
+      :ems_id              => @cfg[:ems_id],
+      :source              => "DATAWAREHOUSE",
```
agrare: Is `DATAWAREHOUSE` the right source for this?  I thought this was a `MonitoringManager` not a `DatawarehouseManager`?

moolitayer: @zgalor had the same concern. See https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/40#discussion_r130555667

agrare: `:source => "PROMETHEUS"` might be more obvious

agrare: Ah I missed that, would it make sense to fix that first?

moolitayer: Unfortunately that change will take weeks to merge and I don't want it blocking this PR.
I'll update the code here once possible.

moolitayer:
@simon3z can you handle this one? i'd like to avoid merging my own PRs

simon3z:
> @simon3z can you handle this one? i'd like to avoid merging my own PRs

@moolitayer I'll wait to see this in action as well. Ping me for the merge after the demo.

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -72,6 +72,12 @@ def verify_hawkular_credentials
     client.hawkular_try_connect
   end
 
+  def verify_prometheus_credentials
+    ensure_monitoring_manager
+    monitoring_manager.verify_credentials
```
moolitayer: cc @nimrodshn this should fix validation for prometheus_alerts

nimrodshn: @moolitayer nice!

ilackarms:
tested working based on 4.6 deployment efforts

simon3z:
> tested working based on 4.6 deployment efforts

@ilackarms IIUC this still requires some manual activation?

ilackarms:
@simon3z yes, need to manually create alert policy in Control dashboard

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    annotations = event["annotations"]
+    event[:url] = annotations["url"]
+    event[:severity] = parse_severity(annotations["severity"])
+    labels = event["labels"]
+    event[:ems_ref] = incident_identifier(event, labels, annotations)
+    event[:resolved] = event["status"] == "resolved"
+    timestamp = event["timestamp"]
+
+    target = find_target(labels)
```
simon3z: @moolitayer all the `target` references below (e.g. `target.id`, etc.) are crashing when target is `nil`.
cc @ilackarms @joelddiaz

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -0,0 +1,118 @@
...
+  def event_monitor_handle
+    @event_monitor_handle ||= ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::Stream.new(@ems)
+  end
+
+  def reset_event_monitor_handle
+    @event_monitor_handle = nil
+  end
+
+  def stop_event_monitor
+    @event_monitor_handle.stop unless @event_monitor_handle.nil?
```
simon3z: @moolitayer there's a codeclimate comment to remember here for future fixes/refactorings:
```
Use safe navigation (&.) instead of checking if an object exists before calling the method. 
```

moolitayer: See this comment: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/40#issuecomment-319721554

simon3z: @moolitayer why can't you use `try` meanwhile?

PR #41, opened 2017-06-15
Add openshift image resource

enoodle:
as a supplement to https://github.com/ManageIQ/manageiq/pull/15386 a new resource type for openshift container images is needed.


enoodle:
Sorry, at least at this stage I understand that this is not needed.

PR #45, opened 2017-06-20
Using the new options for image-scanning options

enoodle:
Based on https://github.com/ManageIQ/manageiq/pull/15398

This will use the options field for per-provider configuration in image scanning.
Also, This will try to use the http_proxy set in the options to connect to the provider instead of the one defined in the settings file. Originally it was planned to have `https_proxy` and `no_proxy` options as well for the connection to the provider but as they are not used currently I left them out.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -14,6 +14,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
   require_nested :Scanning
 
   include ManageIQ::Providers::Kubernetes::ContainerManagerMixin
+  include ManageIQ::Providers::Kubernetes::ContainerManager::Options
 
   # See HasMonitoringManagerMixin
   has_one :monitoring_manager,
--- a/app/models/manageiq/providers/kubernetes/container_manager/options.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/options.rb
@@ -1,0 +1,75 @@
+module ManageIQ::Providers::Kubernetes::ContainerManager::Options
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def proxy_settings
+      {
+        :http_proxy => {
+          :label          => N_('HTTP Proxy'),
+          :help_text      => N_('HTTP Proxy to connect ManageIQ to the provider. example: http://user:password@my_https_proxy'),
+          :global_default => VMDB::Util.http_proxy_uri,
+        },
+      }
+    end
+
+    def advanced_settings
+      {
+        :image_inspector_options => {
+          :label     => N_('Image Inspector Options'),
+          :help_text => N_('Settings for Image Inspector tool'),
+          :settings  => {
+            :http_proxy  => {
+              :label     => N_('HTTP Proxy'),
+              :help_text => N_('HTTP Proxy to connect image inspector pods to the internet. example: http://user:password@my_https_proxy'),
+            },
+            :https_proxy => {
+              :label     => N_('HTTPS Proxy'),
+              :help_text => N_('HTTPS Proxy to connect image inspector pods to the internet. example: https://user:password@my_https_proxy'),
+            },
+            :no_proxy    => {
+              :label     => N_('NO Proxy'),
+              :help_text => N_('NO Proxy lists urls that should\'nt be sent to any proxy. example: my_file_server.org'),
+            },
+            :repository  => {
+              :label          => N_('Image-Inspector Repository'),
+              :help_text      => N_('Image-Inspector Repository. example: openshift/image-inspector'),
+              :global_default => Settings.ems.ems_kubernetes.image_inspector_repository,
+            },
+            :registry    => {
+              :label          => N_('Image-Inspector Registry'),
+              :help_text      => N_('Registry to provide the image inspector repository. example: docker.io'),
+              :global_default => Settings.ems.ems_kubernetes.image_inspector_registry,
+            },
+            :image_tag   => {
+              :label          => N_('Image-Inspector Tag'),
+              :help_text      => N_('Image-Inspector image tag. example: 2.1'),
+              :global_default => ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job::INSPECTOR_IMAGE_TAG,
+            },
+            :cve_url     => {
+              :label     => N_('CVE location'),
+              :help_text => N_('Enables defining a URL path prefix for XCCDF file instead of accessing the default location.
+  example: http://my_file_server.org:3333/xccdf_files/
+  Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
+              # Future versions of image inspector will extend this.
+            },
+          }
+        }
+      }
+    end
+
+    def provider_settings
+      {
+        :proxy_settings    => {
+          :label     => N_('Proxy Settings'),
+          :help_text => N_('Proxy Settings for connection to the provider'),
+          :settings  => proxy_settings,
+        },
+        :advanced_settings => {
+          :label     => N_('Advanced Settings'),
+          :help_text => N_('Advanced Settings for provider configuration'),
+          :settings  => advanced_settings,
+        }
+      }
+    end
+  end
+end
--- a/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
@@ -13,7 +13,6 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   ERRCODE_NOTFOUND = 404
   IMAGE_INSPECTOR_SA = 'inspector-admin'
   INSPECTOR_ADMIN_SECRET_PATH = '/var/run/secrets/kubernetes.io/inspector-admin-secret-'
-  ATTRIBUTE_SECTION = 'cluster_settings'
   PROXY_ENV_VARIABLES = %w(no_proxy http_proxy https_proxy)
 
   def load_transitions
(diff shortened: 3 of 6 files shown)


simon3z:
@enoodle nice! I think this would do it.

app/models/manageiq/providers/kubernetes/container_manager/options.rb
```diff
@@ -0,0 +1,70 @@
...
+            :global_default => Settings.ems.ems_kubernetes.image_inspector_registry,
+          },
+          :image_tag   => {
+            :label          => N_('Image-Inspector Tag'),
+            :help_text      => N_('Image-Inspector image tag'),
+            :global_default => ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job::INSPECTOR_IMAGE_TAG,
+          },
+          :cve_url     => {
+            :label     => N_('CVE location'),
+            :help_text => N_('Enables defining a URL for XCCDF file instead of accessing the Internet'),
```
moolitayer: I remember something about this being a prefix  and not url? 

would it be possible to add an example to each value, even only a comment?

I'm thinking this would great if I need to use one of these I could look here and see what I need.

enoodle:
@moolitayer I can also have the example in the `help_text` that I plan to present in the UI (when you hover over the field)

moolitayer:
> @moolitayer I can also have the example in the help_text that I plan to present in the UI (when you hover over the field)

:+1: That would be much better :hot_pepper:

enoodle:
@moolitayer done.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -348,6 +347,7 @@ def inspector_admin_secret
   end
 
   def pod_definition(inspector_admin_secret_name)
+    @ems_options = ext_management_system.options[:image_inspector_options]
```
moolitayer: Can we please have ems_options as a method and not an instance variable?
I might be reading this wrong but the usage of `@ems_options[:registry]` in line 429 looks like a side effect of calling this method first. (you can memoize it if you want to save lookups)

enoodle: @moolitayer I wanted to prevent accessing the DB multiple times to fetch this value. I think maybe to move it to the `initializing` function. what do you think?

app/models/manageiq/providers/kubernetes/container_manager/options.rb
```diff
@@ -0,0 +1,73 @@
...
+            :global_default => Settings.ems.ems_kubernetes.image_inspector_registry,
+          },
+          :image_tag   => {
+            :label          => N_('Image-Inspector Tag'),
+            :help_text      => N_('Image-Inspector image tag. example: 2.1'),
+            :global_default => ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job::INSPECTOR_IMAGE_TAG,
+          },
+          :cve_url     => {
+            :label     => N_('CVE location'),
+            :help_text => N_('Enables defining a URL path prefix for XCCDF file instead of accessing the default location.
```
moolitayer: :+1:

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -347,6 +346,10 @@ def inspector_admin_secret
     return nil
   end
 
+  def ems_options
+    @provider_options ||= ext_management_system.options[:image_inspector_options]
```
moolitayer: :+1:

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -425,17 +429,25 @@ def add_secret_to_pod_def(pod_def, inspector_admin_secret_name)
...
   end
 
   def inspector_proxy_env_variables
-    settings = ext_management_system.custom_attributes
-    settings.where(:section => ATTRIBUTE_SECTION,
-                   :name    => PROXY_ENV_VARIABLES).each_with_object([]) do |att, env|
-      env << {:name  => att.name.upcase,
-              :value => att.value} unless att.value.blank?
+    PROXY_ENV_VARIABLES.each_with_object([]) do |var_name, env|
+      if ems_options.keys.include?(var_name.to_sym)
```
moolitayer: `ems_options.key?(var_name.to_sym)`

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -108,11 +108,12 @@ def ssl_cert_store(endpoint = default_endpoint)
...
-      :port     => options[:port] || port,
-      :user     => options[:user] || authentication_userid(options[:auth_type]),
-      :pass     => options[:pass] || authentication_password(options[:auth_type]),
-      :bearer   => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :hostname    => options[:hostname] || address,
+      :port        => options[:port] || port,
+      :user        => options[:user] || authentication_userid(options[:auth_type]),
+      :pass        => options[:pass] || authentication_password(options[:auth_type]),
+      :bearer      => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :http_proxy  => self.options.fetch_path(:proxy_settings, :http_proxy),
```
moolitayer: ems_options.fetch_path maybe? `self.options` is nil according to Travis

moolitayer:
cc @cben @simon3z

simon3z:
@enoodle can you fix travis?

enoodle:
@moolitayer I added tests and now travis is green. PTAL

app/models/manageiq/providers/kubernetes/container_manager/options.rb
```diff
@@ -0,0 +1,73 @@
+module ManageIQ::Providers::Kubernetes::ContainerManager::Options
+  extend ActiveSupport::Concern
```
cben: do you need ActiveSupport::Concern?  what does it do here?  (`extend` into an `extend`ed module is too much for my brain ;-)

I think you don't need it, this simply has methods that become class methods due to `extend ...ContainerManager::Options`.
or if you want to use ActiveSupport::Concern idiomatically, switch to `include`ing this module and put the methods under `ClassMethods`.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -347,6 +346,10 @@ def inspector_admin_secret
     return nil
   end
 
+  def ems_options
```
cben: consider renaming to `ems_image_inspector_options` (or something intermediate).
because `ems_options[:registry]` sounds like these are top-level options on the ems.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -425,17 +429,24 @@ def add_secret_to_pod_def(pod_def, inspector_admin_secret_name)
...
   end
 
   def inspector_proxy_env_variables
-    settings = ext_management_system.custom_attributes
-    settings.where(:section => ATTRIBUTE_SECTION,
-                   :name    => PROXY_ENV_VARIABLES).each_with_object([]) do |att, env|
-      env << {:name  => att.name.upcase,
-              :value => att.value} unless att.value.blank?
+    PROXY_ENV_VARIABLES.each_with_object([]) do |var_name, env|
+      next unless ems_options.key?(var_name.to_sym)
```
cben: not sure it matters, but previous code checking `.blank?` allowed "unsetting" by setting to empty string, this requires actually unsetting the option — is that even possible with the new per-EMS options?

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -108,11 +108,12 @@ def ssl_cert_store(endpoint = default_endpoint)
...
-      :port     => options[:port] || port,
-      :user     => options[:user] || authentication_userid(options[:auth_type]),
-      :pass     => options[:pass] || authentication_password(options[:auth_type]),
-      :bearer   => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :hostname    => options[:hostname] || address,
+      :port        => options[:port] || port,
+      :user        => options[:user] || authentication_userid(options[:auth_type]),
+      :pass        => options[:pass] || authentication_password(options[:auth_type]),
+      :bearer      => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
+      :http_proxy  => self.options ? self.options.fetch_path(:proxy_settings, :http_proxy) : nil,
```
cben: why `:proxy_settings, :http_proxy` here and just `:http_proxy` in `kubernetes_connect` ?

enoodle: `kubernetes_connect` gets its options from here, calling `raw_connect` will eventually `kubernetes_connect`. I put them in `http_proxy` so no need for the `proxy_settings`. Openshift calls `kubernetes_connect` as well after adding its own options.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: It is possible to totally unset an option if you set the options without the ones you don't want:
`ems.update(:options => {})` or `ems.update(:options => {:image_inspector_options => {:http_proxy => "my_proxy"}})`. The options are only set from the UI [1] so it will be up to that piece of code to make sure that empty values are not set.
I wonder if it is smart to add tests for the empty option case making sure that they are treated as unset. (then making sure it is so here) - This will prevent us from assigning them the empty value. WDYT?

[1] https://github.com/ManageIQ/manageiq-ui-classic/pull/1652/files#diff-f402e411cd6dc158c724b14d72ead612R445

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -14,6 +14,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
   require_nested :Scanning
 
   include ManageIQ::Providers::Kubernetes::ContainerManagerMixin
+  include ManageIQ::Providers::Kubernetes::ContainerManager::Options
```
moolitayer: :+1:

moolitayer: @enoodle @cben didn't we miss a require nested of the new file here?

PR #46, opened 2017-06-20
Separate data_index lookups from parsing

cben:
We want to formulate graph refresh in a way that can process partial info independently.
Some pieces are still missing to actually achieve that, but it's clear we need to untangle parsing logic from the `@data_index` writes and reads that currently pervade it.

This PR refactors the *current* parser in that direction.
`parse_foo` methods will become pure functions processing their piece of the inventory;
only the `get_foos` methods will read and write `@data_index` to link the resulting hashes
(`get_foos_graph` methods introduced in #30 will do linking differently; I matching changes for them but let's review the non-graph part here.)

- Exception: image & registry parsing still *reads and writes* `@data_index` in-place.  I can live with that for now, will revisit it later.

cc @zakiva @enoodle @simon3z @agrare 
~~WIP because I still need to adapt refresh_parser_spec~~, but please review the refactoring.
(Recommend reviewing by commits)

@miq-bot add-label refactoring

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -532,19 +532,10 @@ module ManageIQ::Providers::Kubernetes
       if new_result[:ems_ref].nil? # Typically this happens for kubernetes services
         new_result[:ems_ref] = "#{new_result[:namespace]}_#{new_result[:name]}"
       end
-      container_groups = []
 
-      endpoint_container_groups = @data_index.fetch_path(
+      container_groups = @data_index.fetch_path(
         :container_endpoints, :by_namespace_and_name, new_result[:namespace],
         new_result[:name], :container_groups)
-      endpoint_container_groups ||= []
-
-      endpoint_container_groups.each do |group|
-        cg = @data_index.fetch_path(
-          path_for_entity("pod"), :by_namespace_and_name, group[:namespace],
-          group[:name])
-        container_groups << cg unless cg.nil?
-      end
 
       labels = parse_labels(service)
       new_result.merge!(
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -103,6 +103,12 @@ module ManageIQ::Providers::Kubernetes
       @data[key].each do |cg|
         node_name = cg.delete(:container_node_name)
         cg[:container_node] = node_name && @data_index.fetch_path(path_for_entity("node"), :by_name, node_name)
+        cg[:project] = @data_index.fetch_path(path_for_entity("namespace"), :by_name, cg[:namespace])
+        replicator_ref = cg.delete(:container_replicator_ref)
+        cg[:container_replicator] = replicator_ref && @data_index.fetch_path(
+          path_for_entity("replication_controller"), :by_namespace_and_name,
+          replicator_ref[:namespace], replicator_ref[:name]
+        )
         @data_index.store_path(key, :by_namespace_and_name,
                                cg[:namespace], cg[:name], cg)
       end
@@ -580,11 +586,8 @@ module ManageIQ::Providers::Kubernetes
         :reason                => pod.status.reason,
         :container_node_name   => pod.spec.nodeName,
         :container_definitions => [],
-        :container_replicator  => nil,
         :build_pod_name        => pod.metadata.try(:annotations).try("openshift.io/build.name".to_sym)
       )
-
-      new_result[:project] = @data_index.fetch_path(path_for_entity("namespace"), :by_name, pod.metadata.namespace)
 
       # TODO, map volumes
       # TODO, podIP
@@ -602,6 +605,7 @@ module ManageIQ::Providers::Kubernetes
         )
       end
 
+      new_result[:container_replicator_ref] = nil
       # NOTE: what we are trying to access here is the attribute:
       #   pod.metadata.annotations.kubernetes.io/created-by
       # but 'annotations' may be nil. The weird attribute name is
@@ -611,9 +615,10 @@ module ManageIQ::Providers::Kubernetes
         # NOTE: the annotation content is JSON, so it needs to be parsed
         createdby = JSON.parse(createdby_txt)
         if createdby.kind_of?(Hash) && !createdby['reference'].nil?
-          new_result[:container_replicator] = @data_index.fetch_path(
-            path_for_entity("replication_controller"), :by_namespace_and_name,
-            createdby['reference']['namespace'], createdby['reference']['name'])
+          new_result[:container_replicator_ref] = {
+            :namespace => createdby['reference']['namespace'],
+            :name      => createdby['reference']['name']
+          }
         end
       end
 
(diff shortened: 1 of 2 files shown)


cben:
Rebased on top of merged #30 and unmerged #48.

@agrare have some matching graph refresh changes in ca158b0a89742aef29e345efb26600b361f8fc13 on cben/graph_refresh branch (needs rebase).
Deliberately keeping them out of this PR for easier review.

@zakiva @enoodle @zeari @moolitayer Please review
(travis broken because I need to rewrite refresh_parser_spec; refresher_spec passed after every commit here)

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -67,6 +68,19 @@ def get_services(inventory)
       key = path_for_entity("service")
       process_collection(inventory["service"], key) { |s| parse_service(s) }
       @data[key].each do |se|
+        se[:container_groups] = @data_index.fetch_path(
+          :container_endpoints, :by_namespace_and_name, se[:namespace], se[:name],
+          :container_groups
+        )
+        se[:project] = @data_index.fetch_path(path_for_entity("namespace"), :by_name, se[:namespace])
+
+        # TODO: this loop only uses last port config - BUG?
```
zeari: I think I remember this. we are filling a `has_one` registry relation so we just need one.

cben: yes but this code always  overwrites with last.  if first port matches a registry and 2nd doesn't, we get nil.
should probably use `.detect {}`

zeari: right, sounds like a bug then...

cben:
Fixed refresh_parser_spec, PTAL.
The only part where it seriously tests `@data_index` linking is for image parsing, luckily I'm not touching that yet ;-)

agrare:
I think this is looking great @cben !

simon3z:
@agrare @zeari @enoodle @moolitayer can you review/approve? Thanks.

moolitayer:
@cben It's great that you separate your commits.
Can you please also squash the rubocop one to the commit that introduced it?
That way someone looking at the log tomorrow will see your actual change which is more interesting.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -532,19 +532,10 @@ def parse_service(service)
       if new_result[:ems_ref].nil? # Typically this happens for kubernetes services
         new_result[:ems_ref] = "#{new_result[:namespace]}_#{new_result[:name]}"
       end
-      container_groups = []
 
-      endpoint_container_groups = @data_index.fetch_path(
+      container_groups = @data_index.fetch_path(
```
moolitayer: Looking at 49d74ab165ce79496bd0c73c7cb131cfb917c004, the data_index lookup is container_endpoints but the variable name is container_groups. Can you please fix that (squash the relevant code from a future commit)?

moolitayer:
@cben I need some context, I see we have ems_inv_to_hashes and ems_inv_to_collections. 
I assume the first is our existing one and the latter is for graph.
How come this change is in the methods called by the old interface?

This change is not related to graph refresh?

cben:
> I see we have ems_inv_to_hashes and ems_inv_to_collections.
> I assume the first is our existing one and the latter is for graph.

Correct.

> How come this change is in the methods called by the old interface? This change is not related to graph refresh?

The `parse_foo` methods are used by both old and graph refresh.
The `get_foo` methods will be distinct from `get_foo_graph` [need better names, for both].

I deliberately touched here only familiar non-graph parts, for easier review.  Graph changes that benefit from this will be in #52.  (Graph refresher test is presently broken anyway, it's `pending` for now...)

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
cben: Note the end of the fetched path takes the endpoint's .container_groups

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -672,16 +681,9 @@ def parse_persistent_volume(persistent_volume)
         :status_phase            => persistent_volume.status.phase,
         :status_message          => persistent_volume.status.message,
         :status_reason           => persistent_volume.status.reason,
-        :persistent_volume_claim => nil
+        :persistent_volume_claim_ref => persistent_volume.spec.claimRef,
```
moolitayer: here also explicit hash would be nice

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -715,12 +717,7 @@ def parse_persistent_volume_claim(claim)
     end
 
     def parse_quota(resource_quota)
```
moolitayer: would you mind renaming this function (different PR)?
There is an inconsistency between `parse_quota` and `get_resource_quota`

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
cben: Ah. Old code did 2 passes building `endpoint_container_groups` then (AFAICT) a no-op pass building same `container_groups`.  New code does 1 pass so the output var is `container_groups`.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
cben: done in 39ab5944c094482d253396fbcc76cb446907c139
also rewrote to handle nil more explicitly, now looks closer to original code.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -638,16 +651,12 @@ def parse_pod(pod)
...
 
       (entity.subsets || []).each do |subset|
         (subset.addresses || []).each do |address|
           next if address.targetRef.try(:kind) != 'Pod'
-          cg = @data_index.fetch_path(
-            path_for_entity("pod"), :by_namespace_and_name,
-            address.targetRef.namespace, address.targetRef.name
-          )
-          new_result[:container_groups] << cg unless cg.nil?
+          new_result[:container_groups_refs] << address.targetRef.to_h
```
cben: done in 58d6182bf4bf583acd4db897ade843833761115e
also rewrote to handle nil more explicitly, now looks closer to original code.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
cben: => #53

cben:
@miq-bot add-label fine/no

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -114,6 +147,11 @@ def get_persistent_volumes(inventory)
       key = path_for_entity("persistent_volume")
       process_collection(inventory["persistent_volume"], key) { |n| parse_persistent_volume(n) }
       @data[key].each do |pv|
+        pvc_ref = pv.delete(:persistent_volume_claim_ref)
+        pv[:persistent_volume_claim] = pvc_ref && @data_index.fetch_path(
+          path_for_entity("persistent_volume_claim"),
+          :by_namespace_and_name, pvc_ref[:namespace], pvc_ref[:name]
+        )
```
moolitayer: So in case there was no ref this code together with line 697,  will fetch_path with `nil` argument.
I think it would be nicer if this was made explicit in the future (don't get path if no ref).

cben:
ping @simon3z, this is ready.

PR #48, opened 2017-06-21
Get rid of unnecessary magic ["table"] RecursiveOpenStruct access

cben:
> *Toi, toi, toi!* (from an old superstition): An exclamation after a statement or declaration in order to prevent a hex from being put on it, accompanied by knocking on a wooden object such as a **table** with one’s knuckles.
> — http://forward.com/culture/15158/spit-your-way-to-safety-toi-toi-toi/ [**emphasis** mine]
> 
> The origin of the custom may be in Germanic folklore, wherein supernatural beings are thought to live in trees, and can be invoked for protection.
> — https://en.wikipedia.org/wiki/Knocking_on_wood

In our case the `obj["table"][:namespace]` incantation summons the tree of hashes that dwells inside a RecursiveOpenStruct, to ward off a monkey patch.
The custom to say `obj["table"][:namespace]` probably arose because `obj[:namespace]` didn't work either until recursive-open-struct 1.0.3 — but it also relies on obscure bug fixed in same 1.0.3!  We happen to be indirectly pegged on 1.0.0 but would break soon.
(See https://github.com/aetherknight/recursive-open-struct/issues/51, `[]` used to invoke `send`, which sees any method including `#namespace` — but also including the normally protected `attr_accessor :table` inherited from ruby's `OpenStruct`...)

There are safer ways to avoid clash with a method (`.to_h[:namespace]`).
But luckily manageiq's `Object#namespace` monkey patch got removed in Aug 2015 (https://github.com/ManageIQ/more_core_extensions/pull/20), and *tfu tfu tfu* is not coming back, so the whole dance is unnecessary, we can just use `.namespace` access :-)

Tested against both r-o-s 1.0.0 and 1.0.4.

@miq-bot add-label bug, technical dept

@simon3z @moolitayer I believe we should backport this.  Alternatively could peg older manageiq releases to r-o-s < 1.0.3 but that might prevent kubeclient / image-inspector-client upgrades.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
@@ -63,7 +63,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EventCatcherMixin
       :timestamp => event.object.lastTimestamp,
       :kind      => event.object.involvedObject.kind,
       :name      => event.object.involvedObject.name,
-      :namespace => event.object.involvedObject['table'][:namespace],
+      :namespace => event.object.involvedObject.namespace,
       :reason    => event.object.reason,
       :message   => event.object.message,
       :uid       => event.object.involvedObject.uid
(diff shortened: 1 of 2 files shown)


cben:
@miq-bot add-label bug, technical debt

simon3z:
@cben please prepare a BZ so that we can backport.

PR #52, opened 2017-06-26
Graph refresh: links between entities

cben:
Followup to #30, implements most entity links in get_*_graph.

Now kubernetes graph refresh passes full refresher tests!
- Except tagging by labels which is skipped.

A few things starting with _ indicate still missing functionality that is not covered by refresher_spec, will add tests later.

Mostly using `lazy_find` by secodary indexes on name or (namespace, name).
(enabled by https://github.com/ManageIQ/manageiq/pull/15447)

- Still using `@data_index` for images & registries.
- Used a local hash similar to data_index for endpoint/service matching.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -36,11 +36,14 @@ module ManageIQ::Providers::Kubernetes
       get_services(inventory)
       get_component_statuses(inventory)
       EmsRefresh.log_inv_debug_trace(@data, "data:")
+
+      # Returning a hash triggers save_inventory_container code path.
       @data
     end
 
     def ems_inv_to_inv_collections(ems, inventory, _options = Config::Options.new)
       initialize_inventory_collections(ems)
+      get_additional_attributes_graph(inventory) # TODO: untested?
       get_nodes_graph(inventory)
       get_namespaces_graph(inventory)
       get_resource_quotas_graph(inventory)
@@ -49,9 +52,14 @@ module ManageIQ::Providers::Kubernetes
       get_persistent_volume_claims_graph(inventory)
       get_persistent_volumes_graph(inventory)
       get_pods_graph(inventory)
-      get_services_graph(inventory)
+      get_endpoints_and_services_graph(inventory)
       get_component_statuses_graph(inventory)
+      # The following use images resulting from parsing pods, so must be called after.
+      # TODO: openshift images parsing will have to plug before this.
+      get_container_images_graph
+      get_container_image_registries_graph
 
+      # Returning an array triggers ManagerRefresh::SaveInventory code path.
       @inv_collections.values
     end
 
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -1153,29 +1142,24 @@ def path_for_entity(entity)
...
-      hash = @data_index.fetch_path(
-        path_for_entity("replication_controller"),
-        :by_namespace_and_name, namespace, name
-      )
-      @inv_collections[:container_replicators].lazy_find(hash[:ems_ref])
+    def lazy_find_replicator(hash)
+      return nil if hash.nil?
+      # Subtle: we can match replicators by project name - instead of project ems_ref -
+      # because when indexed their :container_project a itself lazy_find :by_name,
+      # so its .to_s is the project name.
```
cben: @Ladas @agrare This troubles me ^^
Even without secondary refs, the indirection when manager_ref contains links to other entities, is magic when it works but scared me a bit...

Here I'd prefer to keep `:namespace` in the replicator data, and explicitly use it in indexes.  I need to blacklist it so ActiveRecord .create doesn't complain.  When I tried this before, InventoryCollection didn't let me blacklist things that are parts of manager_ref — can we relax this (or not enforce it for secondary indexes)?

cben:
This now passes referher specs, except tag mapping :-)
But it depends on WIP https://github.com/ManageIQ/manageiq/pull/15447, which is why Travis fails.
@enoodle @zeari @zakiva @moolitayer @yaacov @agrare  please review.

- [ ] There are still some disabled code parts (things starting with `_`), which means they lack tests.  I'll work on this in followup PRs.
- [ ] Endpoints parsing into container_groups<->container_services many-to-many uses a local data_index-like hash.  This probably can be rewritten to directly populate an InventoryCollection for the join table; I'd prefer to tackle this in a later PR.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
Ladas: Yeah we can relax the condition for the secondary indexes. It make sense only for primary index, otherwise you would not be able to identify the same record later.

Yeah having lazy_link as part of the index is funky :-)

cben: cool, switched to index by blacklisted :namespace & name.

simon3z:
@cben can you squash the commits? I think we're ready for merge.

cben:
Squashed & updated PR description.
But wait, another failure on Travis.  => Should be fixed by #55

cben:
@simon3z ready for merge?

- fixed a wrong call in `get_nodes_graph`.
- simplified `lazy_find_*` methods using `lazy_find_by`.
- snuck in a few more `:attributes_blacklist => [:namespace]` so I can consistently `[:namespace]` without having to `.delete(:namespace)` in some cases.

PR #44, opened 2017-06-20
Refresher_spec: Add deletions tests

zakiva:
Extracted k8s entities tests from https://github.com/ManageIQ/manageiq-providers-openshift/pull/18

@simon3z @cben @moolitayer Please review

PR #55, opened 2017-07-05
Update disconnection test to understand soft delete

cben:
Disconnect tests added in #44 were expecting ems_id to become nil,
which doesn't happen after soft delete https://github.com/ManageIQ/manageiq/pull/15251.
Fixed by asking `.archived?` instead.


cben:
@moolitayer @zakiva @simon3z this makes Travis green 👷 🍏 

@miq-bot add-label test

cben:
I see there is discussion on ManageIQ/manageiq#15251 whether it should have been merged;
I believe this PR is good even if it gets reverted for now, as `archived?` method existed before soft delete too.

cben:
Ping @moolitayer @simon3z tiny fix for red master.

PR #57, opened 2017-07-12
Changes needed for openshift graph refresh

cben:
Initializing the openshift inv_collections arguably belongs in openshift
repo, but not bothering with move as @agrare has plans to move them both
to core repo.

RFE: https://bugzilla.redhat.com/show_bug.cgi?id=1470021

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -41,8 +41,21 @@ module ManageIQ::Providers::Kubernetes
       @data
     end
 
-    def ems_inv_to_inv_collections(ems, inventory, _options = Config::Options.new)
+    def ems_inv_to_inv_collections(ems, inventory, options = Config::Options.new)
       initialize_inventory_collections(ems)
+
+      ems_inv_populate_collections(inventory, options)
+
+      # The following take parsed hashes from @data_index, populated during
+      # parsing pods and possibly openshift images, so must be called at the end.
+      get_container_images_graph
+      get_container_image_registries_graph
+
+      # Returning an array triggers ManagerRefresh::SaveInventory code path.
+      @inv_collections.values
+    end
+
+    def ems_inv_populate_collections(inventory, _options)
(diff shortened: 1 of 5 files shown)


cben:
Yay, https://github.com/ManageIQ/manageiq-providers-openshift/pull/34 now passes all tests with this :-)
cc @agrare @Ladas @simon3z

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -277,7 +285,7 @@ def get_namespaces_graph(inv)
 
         container_project = collection.build(h)
 
-        get_custom_attributes_graph(container_project, custom_attrs)
+        get_custom_attributes_graph(container_project, custom_attrs) # TODO: untested
```
simon3z: @cben is this comment still relevant? When do you think this will be tested?

cben: Still relevant, none of our VCR recordings have project labels.  (but I'm confident this works)
I didn't bother Erez with adding those while he re-recorded because there are a few more things we have 0 of, I'm still collecting the list...

cben:
Implemented archiving using `:delete_method => :disconnect_inv`.
Refactored the whole file to be run twice:
```
> rspec .../refresher_spec.rb -f documentation

ManageIQ::Providers::Kubernetes::ContainerManager::Refresher
  graph refresh
    .ems_type
    will perform a full refresh on k8s
    when refreshing an empty DB
      saves the objects in the DB
      when refreshing non empty DB
        removes the deleted objects from the DB
        disconnects objects
        archives objects
  classical refresh
    will perform a full refresh on k8s
    .ems_type
    when refreshing an empty DB
      saves the objects in the DB
      when refreshing non empty DB
        removes the deleted objects from the DB
        archives objects
        disconnects objects
```

@zakiva can you review, especially the specs refactor?
I'll do the same in https://github.com/ManageIQ/manageiq-providers-openshift/pull/34

cben:
I had to take out image labels & docker_labels collections, will init them conditionally in openshift PR.

@moolitayer, @simon3z This is ready.

PR #58, opened 2017-07-12
Let ROS recurse_over_arrays, matching kubeclient

cben:
Kubeclient builds `RecursiveOpenStruct`s that also `recurse_over_arrays`:
https://github.com/abonas/kubeclient/blob/v2.4.0/lib/kubeclient/common.rb#L166
Some of our parser tests contain arrays; in some cases we compensate with nested call `RecursiveOpenStruct.new` calls, in some we pass arrays of hashes instead.
This PR ensure kubeclient-like inputs everywhere.

The helper function name was shamelessly engineered to avoid reindenting ;-) but I can easily change it if someone wants.

@miq-bot add-labels refactoring, test
@moolitayer @enoodle @zakiva please review.  I can do openshift too if you like this.

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
@@ -6,7 +6,7 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser do
   describe "parse_namespace" do
     it "handles simple data" do
       expect(parser.send(:parse_namespace,
-                         RecursiveOpenStruct.new(
+                         array_recursive_ostruct(
                            :metadata => {
                              :name              => "proj2",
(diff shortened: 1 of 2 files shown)


cben:
BTW, I put it in spec_helper because a regular `def` was not accessible from "example group" scopes such as setting `example_component_statuses` — only inside `it`.  I tried turning all these group-level variables into `let` but couldn't because `example_component_statuses` is used in a loop that generates examples.  Sometimes RSpec is a pain...
Doing it in spec_helper with both `c.include` and `c.extend` allows usage both inside and outside examples.

cben:
@moolitayer @simon3z please review/merge.

moolitayer:
:+1: we have some more of those occurrences in `job_spec.rb` and `event_catcher_mixin_spec.rb`.
You might want to change those too, could save us some trouble in the future.

PR #59, opened 2017-07-13
Fix support_metrics function to include prometheus

yaacov:
**Description**

The kubernetes provider supports metrics from prometheus, currently the `supports_metrics?` fails to recognize this endpoint.

This PR fixes that.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1470537

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -5,6 +5,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
   extend ActiveSupport::Concern
 
   DEFAULT_PORT = 6443
+  METRICS_ROLES = %w(prometheus hawkular).freeze
+
   included do
     default_value_for :port do |provider|
       # port is not a column on this table, it's delegated to endpoint.
@@ -15,7 +17,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
   end
 
   def supports_metrics?
-    connection_configurations.hawkular != nil
+    endpoints.where(:role => METRICS_ROLES).exists?
   end
 
   module ClassMethods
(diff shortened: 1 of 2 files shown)


yaacov:
@miq-bot add_label compute/containers

@simon3z @cben @moolitayeri please review

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -15,7 +17,9 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
   end
 
   def supports_metrics?
-    connection_configurations.hawkular != nil
+    roles = endpoints.pluck(:role)
+
+    METRICS_ROLES.any? { |r| roles.include?(r.to_s) }
```
cben: I believe you can write intersection `(METRICS_ROLES & roles).any?`.  Your call if that's prettier.

(If we had tons of endpoints optimal query would be `endpoints.where(role => METRICS_ROLES).exists?` but doesn't matter.)

spec/models/manageiq/providers/kubernetes/container_manager_spec.rb
```diff
@@ -9,6 +9,59 @@
     expect(described_class.raw_api_endpoint("::1", 123).to_s).to eq "https://[::1]:123"
   end
 
+  context "#supports_metrics?" do
+    before(:each) do
+      EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+    end
+
+    it "regular provider has no metrics support" do
+      ems = FactoryGirl.create(
```
cben: can this use `FactoryGirl.build` ?  that would avoid actual DB writes, making tests faster.
(if it doesn't work, and with EMS it frequetly doesn't, no need to fight it, create is OK.)

yaacov: build fails here :-( keeping the create

moolitayer:
@yaacov did you test that https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/20 on top of this PR is indeed collecting metrics?

yaacov:
> @yaacov did you test that #20 on top of this PR is indeed collecting metrics?

No, the scope of this PR is the correctness of the function `supports_metrics?` so the specs only check that

PR #60, opened 2017-07-13
[WIP] match registry to service with >1 port configs

cben:
Fixes a (theoretical) bug in matching registries host:ports we see in images to registry service(s) running in the cluster.  If the service has many ports, match would only succeed for last port (previous iterations of the loop were overwritten).

@miq-bot add-label bug
@simon3z I think this is very theoretical, no point backporting?

This PR only touches old hashes→save_inventory refresh, for now doing graph refresh with last-port logic and a TODO as I don't yet know how to implemet correct logic there.
Also adds tests to this part of the parser.  (With 1 port, it's also already covered in *openshift* repo VCR refersher_spec, which still passes.)

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
@@ -1227,5 +1227,44 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser do
         )
       end
     end
+
+    describe "find_service_registry" do
+      it "handles no ports" do
+        expect(parser.send(:find_service_registry, [], "1.2.3.4")).to be nil
+      end
+
+      it "returns nil for no match" do
+        parser.send(:parse_container_image,
+                    "#10.20.30.40:1010/red-dragon/nodejs-ex@sha256:1234",
+                    "docker://sha256:5678")
+        port_configs = [
+          {:name => "1111-tcp", :protocol => "TCP", :port => 1111, :targetPort => 5000},
+        ]
+        expect(parser.send(:find_service_registry, port_configs, "1.2.3.4")).to be nil
+      end
+
+      it "finds first match" do
+        host = "172.30.185.88"
+        # pretend there exist registries on same host ports 5000 and 6000
+        parser.send(:parse_container_image,
+                    "#{host}:5000/red-dragon/nodejs-ex@sha256:1234",
+                    "docker://sha256:5678")
+        parser.send(:parse_container_image,
+                    "#{host}:6000/red-dragon/ruby-ex@sha256:2345",
+                    "docker://sha256:6789")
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
```diff
@@ -1173,4 +1173,98 @@
...
+                    "docker://sha256:5678")
+        port_configs = [
+          {:name => "1111-tcp", :protocol => "TCP", :port => 1111, :targetPort => 5000},
+        ]
+        expect(parser.send(:find_service_registry, port_configs, "1.2.3.4")).to be nil
+      end
+
+      it "finds first match" do
+        host = "172.30.185.88"
+        # pretend there exist registries on same host ports 5000 and 6000
```
zakiva: @cben is it possible? if yes, is there any special reason to take the first one?

cben: I've not seen this in the wild with openshift's registry, just theoretizing (one could `oc edit` the openshift registry, or run their own registry with custom config...)

In principle, service has many ports and should `has_many` registries.  Currently it `belongs_to` registry so we can only pick one.  I really don't expect multiple matches, didn't want to bother with a migration, just wanted a second unrelated port to not mask the one match...

But let's maybe wait until I figure how I'd do this with graph refresh.  (I mostly opened this PR so I can point a TODO to it).
@Ladas @agrare is there a way to set a field to multiple `find_lazy`, expecting 0 to 1 of them to succeed, and have the successful one win?
I'm getting the feeling I should migrate to reverse the relation...

cben: @simon3z or we can give up, decide to only ever match last port config and drop the non-functional loop?

Ladas: @cben sounds like we should really fix the DB relations :-)

Hm, for what you want, I would probably define a InventoryObjectLazyCollection / InventoryObjectLazyMulti, that would eat several InventoryObjectLazy, so we can correctly model the dependencies created by this. In some cases it could get complicated. :-)

cben: Hmm.  I'd rather reverse the DB relation :-)

Ladas: yeah, the cleanest solution is definitely having the DB modeling right :-)

cben:
I'm going to give up, until someone asks to fix this.

The fix here was partial (only old refresh).
If we really care about such situation, the much cleaner solution is to change schema.

However, I should salvage some tests from here, and delete to useless loop.

PR #61, opened 2017-07-13
remove obsolete pod_health_poll

enoodle:
It is not used anymore since the move to use readinessProbe in https://github.com/ManageIQ/manageiq/pull/14578


enoodle:
@moolitayer

PR #63, opened 2017-07-13
Shouldn't have quotes and commas inside %w(...)

cben:
Those actually become part of the words!
Kudos rubocop for calling this out.
Test worked because we had them in both input and expected output.


moolitayer:
Travis passed in https://travis-ci.org/ManageIQ/manageiq-providers-kubernetes/builds/253291172

moolitayer:
@cben fun with regexp: `grep -r -E "%w\([^\)]*(,|')\)" *`

PR #66, opened 2017-07-19
run provider generator to sync changes from core

durandom:
ran the generator again with

```bash
bundle exec rails g provider <name> --path ../
```

these are the changes.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -4,6 +4,7 @@
     :miq_namespace: management-infra
     :image_inspector_registry: docker.io
     :image_inspector_repository: openshift/image-inspector
+    :blacklisted_event_names: []
     :event_handling:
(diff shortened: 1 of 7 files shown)


config/settings.yml
```diff
@@ -4,6 +4,7 @@
     :miq_namespace: management-infra
     :image_inspector_registry: docker.io
     :image_inspector_repository: openshift/image-inspector
+    :blacklisted_event_names: []
```
cben: this is the only thing that I wanted to check for potential functional change;
it's was unset previously, it's read in https://github.com/ManageIQ/manageiq/blob/fa56592fe6/app/models/manageiq/providers/base_manager.rb#L27-L29
which returns [] both before and with this PR (verified in rails c).

PR #67, opened 2017-07-19
Add parse_service spec

cben:
Split from #60, I'm gonna wait with the change there, but more tests are
always good :-)

@miq-bot add-label test
@zeari @zakiva Please review

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
@@ -1173,4 +1173,59 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser do
         })
     end
   end
+
+  context "services" do
+    let(:service) do
+      array_recursive_ostruct(
+        :metadata => {
+          :name              => "docker-registry",
+          :namespace         => "default",
+          :selfLink          => "/api/v1/namespaces/default/services/docker-registry",
+          :uid               => "13d3bcf7-e6f9-11e6-a348-001a4a162683",
+          :resourceVersion   => "651",
+          :creationTimestamp => "2017-01-30T14:33:33Z",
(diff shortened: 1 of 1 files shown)


spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
```diff
@@ -1173,4 +1173,59 @@
...
+          :uid               => "13d3bcf7-e6f9-11e6-a348-001a4a162683",
+          :resourceVersion   => "651",
+          :creationTimestamp => "2017-01-30T14:33:33Z",
+          :labels            => {:"docker-registry" => "default"},
+        },
+        :spec     => {
+          :ports           => [
+            {:name => "5000-tcp", :protocol => "TCP", :port => 5000, :targetPort => 5000},
+          ],
+          :selector        => {:"docker-registry" => "default"},
```
moolitayer: kubeclient returns string as hash keys doesn't it @cben ?

cben: it returns RecursiveOpenStructs (and so does the test here), so symbol/string keys don't matter.
anyway note the colon before the quotes — this is a symbol.
or did you ask something else?

moolitayer: yeah I was wondering why this is a symbol and not a string. I get your answer you are saying it is the same in a ROS

cben: `selectors` hash indeed has user-supplied free form keys so yes a string could make sense, but physically we get JSON that doesn't make that semantic distinction.
I just didn't stop to think about it, used symbols as everywhere.  IMHO making a disctinction where none exists might be more confusing.

moolitayer:
@zeari can you review ?

Issue #34, opened 2017-06-06
kubernetes capture_context_spec.rb runs openshift code


cben:
In #30 we bumped into https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_spec.rb#L14 in this repo creating an `:ems_openshift` factory and therefore running `Openshift::ContainerManager` code.
Beside conceptually wrong dep, that's annoying because a change that needs simultaneous kubernetes and openshift PRs won't be able to pass Travis on *either* repo.

It's not a trivial change to just use `:ems_kubernetes`, causes VCR errors.  (Didn't look deeper.)

@miq-bot add-label technical debt

- It's a metrics test, which in *practice* is an openshift thing, so perhaps could move to openshift repo?
- But in theory IIUC the code should support metrics in kubernetes (if you run a hawkular)?
- More importantly, `metrics_capture` code lives in kubernetes repo, so it's test must live there (to reduce need for separate PRs).
- It's IMHO important that we do test metrics functionality against openshift classes, one way or the other.  Might have to run similar test in both repos...

@simon3z @yaacov what should we do?

yaacov:
@cben @simon3z 

> It's a metrics test, which in practice is an openshift thing, so perhaps could move to openshift repo?

I'm ok moving this to the openshift repo now and doing a new spec only for kubernetes in a different PR.

simon3z:
> I'm ok moving this to the openshift repo now, and doing a new spec only for kubernetes in a different PR.

OK.

cben:
Fixed by #68 and #72.

cben:
@yaacov there is another case of this in prometheus spec:
https://github.com/manageiq/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context_spec.rb#L10

Simply changing to `:ems_kubernetes` factory fails with:
```
     VCR::Errors::UnusedHTTPInteractionError:
       There are unused HTTP interactions left in the cassette:
         - [get https://capture.context.com:8443/oapi/v1] => [200 "{\"kind\":\"APIResourceList\",\"groupVersion\":\"v1\",\"resources\":[{\"name\":\"appliedclust"]
[...]

PR #68, opened 2017-07-19
Use k8s provider in k8s tests

moolitayer:
Should fix test failures of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/38


moolitayer:
@miq-bot add_label tests

@yaacov please review

moolitayer:
@cben please review

cben:
Yay solves half of #34.
AFAICT we don't have a matching test with ems_openshift in openshift repo.  Would it be worth having?

moolitayer:
@cben AFAIK it tests code that is in this repo.
I think we are better off this way - 
we might change that code together with the tests, and then the openshift repo will start failing.

moolitayer:
> It's definitely better this way, I'm only asking if we also want to test this code against openshift EMS in case the behavior differs.

Yeah that's what I was referring to above. Anyway I don't have a strong opinion on that, if you feel we are better off duplicating the tests to manageiq-providers-openshift as well go ahead.

PR #69, opened 2017-07-19
Convert quotas to numeric values

zakiva:
Parse the quotas as numeric (int or float) values instead of string. 

Depends on https://github.com/ManageIQ/manageiq-schema/pull/44

@simon3z @zeari Please review

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -891,15 +891,15 @@ module ManageIQ::Providers::Kubernetes
       end
 
       resource_quota.spec.hard.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_desired] = quota
+        new_result_h[resource_name][:quota_desired] = parse_quantity(quota)
       end
 
       resource_quota.status.hard.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_enforced] = quota
+        new_result_h[resource_name][:quota_enforced] = parse_quantity(quota)
       end
 
       resource_quota.status.used.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_observed] = quota
+        new_result_h[resource_name][:quota_observed] = parse_quantity(quota)
       end
 
       new_result_h.values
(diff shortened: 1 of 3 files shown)


cben:
@zeari how does this relate to your suffixed->value conversion quest?

zeari:
> @zeari how does this relate to your suffixed->value conversion quest?

Its about the same and relies on mostly the same types of conversions

zakiva:
@moolitayer Please review

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -869,28 +869,40 @@ def parse_resource_quota(resource_quota)
...
 
       resource_quota.status.used.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_observed] = quota
+        add_numeric_quota(new_result_h[resource_name], quota, "quota_observed")
       end
 
       new_result_h.values
     end
 
+    def add_numeric_quota(h, quota, key)
```
moolitayer: `s/key/prefix/`

cben:
@zakiva tests in #22 might help (you probably knew this)

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
zakiva: done

zakiva:
@moolitayer `parse_quantity` method was added through #22 by @zeari with some tests (removed it from here after rebase). I updated the quotas specs and added some other example values.

moolitayer:
@zeari @cben review pls

moolitayer:
The code parsing quotas[1] is a little hard to read already and getting harder (I think I wrote it ¯\_(ツ)_/¯). 

@zakiva maybe you can add in another PR an example for the value of resource_quota.spec.<x>.to_h?

I'm sure you'll find the best way to do it

[1]
```
resource_quota.X.Y.to_h.each do |resource_name, quota|
  new_result_h[resource_name][:quota_desired] = quota
end
```

moolitayer:
@zakiva can you check Travis?

zakiva:
Updated the PR according to the schema changes, we'll probably store all values as floats (makes this PR much simpler).

Fryguy:
Merged the schema, so bouncing this PR.

zakiva:
@moolitayer this should be ready now

PR #72, opened 2017-07-23
Use k8s provider in k8s tests

moolitayer:
- Use k8s provider in k8s tests
- Remove lots of VCR interactions not actually needed in the test (but called by an OpenShift refresh)

PR #73, opened 2017-07-25
Add persistor with configurable saver strategy

Ladas:
Add Persistor with configurable saver strategy. Encapsulating ICs to Persistor and exposing the options to Settings. Testing both ;default and :batch strategies

Check with https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/73/files?w=1 to filter out the whitespace changes

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -131,7 +131,6 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
         )
       )
     # images have custom_attributes but that's done conditionally in openshift parser
-    initialize_custom_attributes_collections(manager.container_images, %w(labels docker_labels))
 
     @collections[:container_groups] =
       ::ManagerRefresh::InventoryCollection.new(
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -7,16 +7,13 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
     nil
   end
 
-  def saver_strategy
-    :default
-  end
-
   def shared_options
-    {
-      :strategy       => strategy,
-      :targeted       => targeted,
-      :saver_strategy => saver_strategy
-    }
+    settings_options = options[:inventory_collections].try(:to_hash) || {}
+
+    settings_options.merge(
+      :strategy => strategy,
+      :targeted => targeted,
+    )
   end
 
   def initialize_inventory_collections
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -227,7 +227,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :model_class => ContainerServicePortConfig,
           :parent      => manager,
           :association => :container_service_port_configs,
-          :manager_ref => [:container_service, :ems_ref, :protocol]
+          :manager_ref => [:ems_ref, :protocol]
         )
       )
 
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -1,247 +1,327 @@
 module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
-  def initialize_inventory_collections(ems, _options)
-    # TODO: Targeted refreshes will require adjusting the associations / arels. (duh)
-    @collections = @inv_collections = {}
-    @inv_collections[:container_projects] = ::ManagerRefresh::InventoryCollection.new(
-      :model_class    => ContainerProject,
-      :parent         => ems,
-      :builder_params => {:ems_id => ems.id},
-      :association    => :container_projects,
-      :secondary_refs => {:by_name => [:name]},
-      :delete_method  => :disconnect_inv,
-    )
-    initialize_custom_attributes_collections(ems.container_projects, %w(labels additional_attributes))
+  def targeted
+    false
+  end
 
-    @inv_collections[:container_quotas] = ::ManagerRefresh::InventoryCollection.new(
-      :model_class          => ContainerQuota,
-      :parent               => ems,
-      :builder_params       => {:ems_id => ems.id},
-      :association          => :container_quotas,
-      :attributes_blacklist => [:namespace],
+  def strategy
+    nil
+  end
+
+  def saver_strategy
+    :default
+  end
+
+  def shared_options
+    {
+      :strategy       => strategy,
+      :targeted       => targeted,
+      :saver_strategy => saver_strategy
+    }
+  end
+
+  def initialize_inventory_collections
+    # TODO: Targeted refreshes will require adjusting the associations / arels. (duh)
+    @collections[:container_projects] = ::ManagerRefresh::InventoryCollection.new(
+      shared_options.merge(
+        :model_class    => ContainerProject,
+        :parent         => manager,
+        :builder_params => {:ems_id => manager.id},
+        :association    => :container_projects,
+        :secondary_refs => {:by_name => [:name]},
+        :delete_method  => :disconnect_inv,
+      )
     )
-    @inv_collections[:container_quota_items] = ::ManagerRefresh::InventoryCollection.new(
-      :model_class => ContainerQuotaItem,
-      :parent      => ems,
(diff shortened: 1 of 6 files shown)


Ladas:
cc @agrare @cben

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -1,256 +1,336 @@
...
-        :parent               => ems,
-        :builder_params       => {:ems_id => ems.id},
-        :association          => :container_routes,
-        :attributes_blacklist => [:namespace],
+        shared_options.merge(
+          :model_class          => ContainerRoute,
+          :parent               => manager,
+          :builder_params       => {:ems_id => manager.id},
+          :association          => :container_routes,
+          :attributes_blacklist => [:namespace, :tags],
```
Ladas: @cben I needed to blacklist tags to have OSE test passing, is there some other change that doesn't require this?

cben: Hmm, didn't notice `parse_*` functions are setting tags.  I'm surprised it worked so far without the blacklist :-)

Ladas: It might have been a bad rebase, because it was there and then I noticed it was missing when I rebased yesterday. Without it, the OpenShift specs were failing for me.

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -41,8 +40,12 @@ def ems_inv_to_hashes(inventory, _options = Config::Options.new)
       @data
     end
 
+    def persister_class
+      ManageIQ::Providers::Kubernetes::Inventory::Persister::ContainerManager
```
Ladas: possibly, but the next step should be to expose persistor, instead of @inv_collections.

spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
```diff
@@ -526,7 +526,26 @@ def assert_disconnected(object)
...
+      end
+
+      # TODO: pending graph tag mapping implementation
+      include_examples "kubernetes refresher VCR tests", :check_tag_mapping => false
+    end
+
+    context "with :batch saver" do
+      before(:each) do
+        stub_settings_merge(
+          :ems_refresh => {:kubernetes => {:inventory_collections => {:saver_strategy => :batch}}}
```
Ladas: Settings can handle symbol, we use in on few places now. But yeah, the interface should probably handle string too. Should be done in coming PRs, where I will be refactoring the interface.

cben:
codeclimate is all about InventoryCollection construction, ignorable.

cben:
@miq-bot add-label performance
(a *lot* of performance :-)

cben:
@moolitayer @enoodle @zeari please review.
(`?w=1` link in PR description helps.)

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -41,8 +40,14 @@ def ems_inv_to_hashes(inventory, _options = Config::Options.new)
...
+    end
+
     def ems_inv_to_inv_collections(ems, inventory, options = Config::Options.new)
-      initialize_inventory_collections(ems, options)
+      persister = persister_class.new(ems)
+      # TODO expose Persistor and use that
```
moolitayer: @Ladas can you please explain this TODO?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -1,247 +1,324 @@
...
+  def targeted
+    false
+  end
+
+  def strategy
+    nil
+  end
+
+  def shared_options
+    settings_options = options[:inventory_collections].try(:to_hash) || {}
```
moolitayer: Where is options coming from?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -1,247 +1,324 @@
...
+
+    settings_options.merge(
+      :strategy => strategy,
+      :targeted => targeted,
```
moolitayer: Is it intentional these two can't be configured?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -1,247 +1,324 @@
 module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
-  def initialize_inventory_collections(ems, _options)
+  def targeted
+    false
+  end
+
+  def strategy
+    nil
```
moolitayer: what is the meaning of a nil strategy ? can we be explicit ?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -1,247 +1,324 @@
...
-        :secondary_refs       => {:by_namespace_and_name => [:namespace, :name]},
-        :attributes_blacklist => [:namespace],
+        shared_options.merge(
+          :model_class          => ContainerService,
+          :parent               => manager,
+          :builder_params       => {:ems_id => manager.id},
+          :association          => :container_services,
+          :secondary_refs       => {:by_namespace_and_name => [:namespace, :name]},
+          :attributes_blacklist => [:namespace],
+          :saver_strategy       => :default # TODO(perf) Can't use batch strategy because of usage of M:N container_groups relation
```
moolitayer: M:N = M:M ? if not what does it mean?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: that is on a base class Persistor, taking it from the advanced settings

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
Ladas: just that we should replace the usage of @inv_collections, that we have in many files by the Persister object, that should be the interface

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: yes, those 2 will be overwritten by the Persistor class used for subgraph saving. The interface itself, I still need to redesign it, trying to figure it out now. :-)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: the relation table container_groups_container_services, we need to model it as a separate InventoryCollection, then the batch saving will work. For both container_groups and container_groups_container_services

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: This is the DB searching strategy, default is nil.  We will redefine it for the sub-graph saving Persistor

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
moolitayer: ahh okay, was a little confused by the fact that here the todo is already done...

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
moolitayer: oh I see it on InventoryCollection now, nice doc!

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
Ladas: yeah, we just need to fix all the other places. :-)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: yeah, I need to refactor those though, since the number of settings to cover all the corner cases has grown a lot :-) So that interface + connection to Settings refactoring is on the roadmap next.

moolitayer:
BTW @Ladas it would be nice to have an empty `inventory_collections` in the settings for visibility

Ladas:
@moolitayer right, I haven't figured out the exact format yet, once i do, I'll push it to the settings. We will probably need to be able to set the option for specific Persistor and specific InventoryCollection.

bdunne:
@moolitayer Any reason there's no milestone here?

moolitayer:
@bdunne see https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/72#issuecomment-321251645

PR #76, opened 2017-07-26
Fix "unused HTTP interactions", make test more robust

cben:
The `.find_by` instead of `.first` will make the tested targets stable and matching the VCR even if refresh record order changes.
It also happens to fix VCR errors following #42, but not because of order.
`.first` returned same record as `.find_by` but behavior differed!

The reason is that without `.reload` after refresh, ActiveRecord caches were stale, with unpredictable results...
(Exact queries during refresh changed in #42, with the result that `pod = @ems.container_groups.first` did not go to DB, and `pod.containers` weirdly returned same container twice.)


cben:
@moolitayer @simon3z @Fryguy @zeari togther with #42 this make tests green.
Can be merged in any order, but you probably want this first, then toggle #42 Travis.

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/16521

PR #77, opened 2017-07-28
Add ContainerImage raise event post processing job

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/15678
- [x] https://github.com/ManageIQ/manageiq/pull/15679

Add ContainerImage raise event post processing job, queuing
the raise event as a background job, since it does not
belong under the refresh.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
@@ -37,7 +37,7 @@ module ManageIQ
             container_images_ids = batch.collect { |x| x[:id] }
             MiqQueue.submit_job(
               :class_name  => "ContainerImage",
-              :method_name => 'raise_creation_event',
+              :method_name => 'raise_creation_events',
               :args        => [container_images_ids],
               :priority    => MiqQueue::HIGH_PRIORITY
             )
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
@@ -39,7 +39,7 @@ module ManageIQ
               :class_name  => "ContainerImage",
               :method_name => 'raise_creation_event',
               :args        => [container_images_ids],
-              :priority    => MiqQueue::MIN_PRIORITY
+              :priority    => MiqQueue::HIGH_PRIORITY
             )
           end
         end
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
@@ -21,6 +21,28 @@ module ManageIQ
             end
           end
         end
+
+        def manager_refresh_post_processing(_ems, _target, inventory_collections)
+          indexed = inventory_collections.index_by(&:name)
+          container_images_post_processing(indexed[:container_images])
+        end
+
+        def container_images_post_processing(container_images)
+          # We want this post processing job only for batches, for the rest it's after_create hook on the Model
+          return unless container_images.saver_strategy == :batch
+
+          # TODO extract the batch size to Settings
+          batch_size = 100
+          container_images.created_records.each_slice(batch_size) do |batch|
+            container_images_ids = batch.collect { |x| x[:id] }
+            MiqQueue.submit_job(
+              :class_name  => "ContainerImage",
+              :method_name => 'raise_creation_event',
+              :args        => [container_images_ids],
+              :priority    => MiqQueue::MIN_PRIORITY
+            )
+          end
+        end
       end
     end
   end


Ladas:
@cben a post processing for raising of the Event. Would be nice to extract it out from the normal graph refresh too. And the same way, we can do the tagging.

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
```diff
@@ -21,6 +21,28 @@ def fetch_entities(client, entities)
...
         end
+
+        def manager_refresh_post_processing(_ems, _target, inventory_collections)
+          indexed = inventory_collections.index_by(&:name)
+          container_images_post_processing(indexed[:container_images])
+        end
+
+        def container_images_post_processing(container_images)
+          # We want this post processing job only for batches, for the rest it's after_create hook on the Model
+          return unless container_images.saver_strategy == :batch
```
Ladas: @cben hm, right now the non batch strategies always use Active Record objects, rails with hooks, validations, etc.

Maybe we could leverage the container_images.use_ar_object? and implement non AR mode for the default saver.

Yeah, it would be nice to get rid of the after_create entirely, but it will be harder

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
```diff
@@ -21,6 +21,28 @@ def fetch_entities(client, entities)
...
+
+          # TODO extract the batch size to Settings
+          batch_size = 100
+          container_images.created_records.each_slice(batch_size) do |batch|
+            container_images_ids = batch.collect { |x| x[:id] }
+            MiqQueue.submit_job(
+              :class_name  => "ContainerImage",
+              :method_name => 'raise_creation_event',
+              :args        => [container_images_ids],
+              :priority    => MiqQueue::MIN_PRIORITY
```
Ladas: Yeah, if there is not enough generic workers, the queue can grow, we've seen that.

Ladas:
@cben @enoodle @simon3z right, so the purpose is to remove any code that generates N+1 queries from the refresh. And for the current queue implementation, the best is to batch the MiqQueue calls. There is still ineffective job hidden inside, but at least those can be processed by many generic workers (while we can't scale the refresh worker now)

https://github.com/Ladas/manageiq/blob/4072ddd5c18a099ddd4a36289a57ac377d772ef9/app/models/miq_event.rb#L22

the raise event job also looks pretty complex, with a power to generate many queries per added Image

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
Ladas: the call_automate job scheduled inside has ```:priority     => MiqQueue::HIGH_PRIORITY```, so the same should probably be here

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
agrare: Yeah I agree it'd be nicer to use this in all cases, not just batch saving, and get rid of the `after_create`

Ladas: for usage in all cases, we would need to implement the Old refresh timestamp based post refresh and then remove it from the Model after_create

Ladas:
@cben given the CI passes, I assume we lack specs for this?

Ladas:
@cben ping, I thought this was already in :-)

cben:
@enoodle @moolitayer Please review.  Dependencies are merged.

moolitayer:
@Ladas so this behavior was creating a strain on the queue and this is coming to solve that?

Ladas:
@moolitayer so the issue is that this stress refresh a lot, each raise_event does like 9 queries. So we are helping the refresh by moving it out while it can be processed by the generic workers in parallel. (in other providers, this kind of job belongs to post refresh)

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
```diff
@@ -21,6 +21,28 @@ def fetch_entities(client, entities)
             end
           end
         end
+
+        def manager_refresh_post_processing(_ems, _target, inventory_collections)
```
moolitayer: Not a show stopper, but the name is confusing.
This happens after save to the database right? (since we have the ids already)
I read processing as post parsing(`manager_refresh_post` would make more sense to me)

I think @cben had a previous commented about this

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
```diff
@@ -21,6 +21,28 @@ def fetch_entities(client, entities)
...
+          indexed = inventory_collections.index_by(&:name)
+          container_images_post_processing(indexed[:container_images])
+        end
+
+        def container_images_post_processing(container_images)
+          # We want this post processing job only for batches, for the rest it's after_create hook on the Model
+          return unless container_images.saver_strategy == :batch
+
+          # TODO extract the batch size to Settings
```
moolitayer: :+1:

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
```diff
@@ -21,6 +21,28 @@ def fetch_entities(client, entities)
...
+          batch_size = 100
+          container_images.created_records.each_slice(batch_size) do |batch|
```
moolitayer: @Ladas in what order are we getting created records? is it guaranteed by inventory collection's interface

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
Ladas: right, I am fine with any sane name :-) we will just need to change it also in the core MIQ

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
Ladas: hum, the order should be the same as we saved them, which should be the same as we parsed them

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
moolitayer: OK no point in waiting for the rename, please rename this when you can.

moolitayer:
@Ladas in the codepath relevant to this PR, the creation events aren't raised in the previous flow?
Is it because we are not creating ActiveRecord objects?

Ladas:
@moolitayer yes, we don't instantiate the AR objects in the :batch strategy, so we don't call the hooks

PR #79, opened 2017-07-31
Move kube/openshift specific template code into its provider directory.

lfu:
Part of ManageIQ/manageiq#15523.
Blocks https://github.com/ManageIQ/manageiq-providers-openshift/pull/28.

@miq-bot add_label refactoring

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_template.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_template.rb
@@ -1,0 +1,3 @@
+class ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplate < ManageIQ::Providers::ContainerManager::ContainerTemplate
+  include ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplateMixin
+end
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
@@ -1,0 +1,69 @@
+module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplateMixin
+  extend ActiveSupport::Concern
+
+  MIQ_ENTITY_MAPPING = {
+    "Route"                 => ContainerRoute,
+    "Build"                 => ContainerBuildPod,
+    "BuildConfig"           => ContainerBuild,
+    "Template"              => ContainerTemplate,
+    "ResourceQuota"         => ContainerQuota,
+    "LimitRange"            => ContainerLimit,
+    "ReplicationController" => ContainerReplicator,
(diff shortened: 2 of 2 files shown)


Fryguy:
@simon3z @moolitayer Please review.  This is part of https://github.com/ManageIQ/manageiq/pull/15523 , of which the data migration change has already been merged in https://github.com/ManageIQ/manageiq-schema/pull/35

simon3z:
@moolitayer @zakiva can you reivew?

app/models/manageiq/providers/kubernetes/shared/container_manager/container_template.rb
```diff
@@ -0,0 +1,69 @@
+module ManageIQ::Providers::Kubernetes::Shared::ContainerManager::ContainerTemplate
```
moolitayer: This file is an exact copy of the code removed from app/models/container_template.rb in ManageIQ/manageiq#15523 so :+1: for the code as well as including this in this repo per https://github.com/ManageIQ/manageiq/pull/15523#discussion_r130370234.

As for the new `shared` path currently we have lots of examples of shared code (ContainerManagerMixin, EventCatcherMixin, some of metrics_capure, scanning/job & entities_mapping all from a quick look)

So while I think the new folder is a good idea, having only ContainerTemplate there will do more to confuse.

moolitayer: @cben @simon3z WDYT ?

cben: I generally think of the openshift provider as a "subclass" of kubernetes provider.
Certainly Openshift the technology *is-a* Kubernetes.
Now technically we don't always subclass, we frequently have both openshift and kubernetes include a shared mixin — but I'm not sure it needs drawing attention to.
(And if it does, I'd rather use "Mixin" suffix)

-0 on need for shared/ folder, and -0.5 on deeper nesting.  For God's sake, we're already working in
```
manageiq/plugins/manageiq-providers-kubernetes/app/models/manageiq/providers/kubernetes/container_manager/
```
:cry:  It's not a veto, if others think it's good I'll play along.

cben: Also, as things are now broken until these 3 PRs land, I'm in favor of merging sooner and figuring out naming later.

moolitayer: I would rather not merge as is(especially since ManageIQ/manageiq#15523 was not yet merged).
Seems `app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb` is more consistent with what we currently have.

moolitayer:
@zakiva @cben another approval is needed

Fryguy:
Since I merged https://github.com/ManageIQ/manageiq/pull/15523, I am going to merge the dependent PRs to avoid things going red.

PR #81, opened 2017-08-03
Fix fallouts from ContainerTemplate STI

cben:
I assumed in this code the model used when defining these collections is same model as the `parent` for which we are looking them up, which is wrong given STI.
This fixes errors in manageiq-providers-openshift repo:
```
    can't save: missing @inv_collections[[:custom_attributes_for, "ContainerTemplate", "labels"]]
    can't save: missing @inv_collections[[:custom_attributes_for, "ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplate", "labels"]]
```
following https://github.com/ManageIQ/manageiq-providers-openshift/pull/28 and related PRs.

I'm still seeing other errors in openshift specs locally, but this gets us closer to green...

@lfu @moolitayer @Ladas please review.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -418,7 +418,7 @@ module ManageIQ::Providers::Kubernetes
 
     # polymorphic, relation disambiguates parent
     def get_container_conditions_graph(parent, hashes)
-      model_name = parent.inventory_collection.model_class.name
+      model_name = parent.inventory_collection.model_class.base_class.name
       hashes.to_a.each do |h|
         h = h.merge(:container_entity => parent)
         @inv_collections[[:container_conditions_for, model_name]].build(h)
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
```diff
@@ -418,7 +418,7 @@ def get_pods_graph(inv)
 
     # polymorphic, relation disambiguates parent
     def get_container_conditions_graph(parent, hashes)
-      model_name = parent.inventory_collection.model_class.name
+      model_name = parent.inventory_collection.model_class.base_class.name
```
moolitayer: we don't need base class here right? (since the classes use in collections are already basic)

cben:
Added missing `require_nested` (see https://github.com/ManageIQ/manageiq/pull/15523/files#r131175697 for *interesting* way in which autoloading failed; doesn't matter much, we want explicit load all files)

app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
cben: You're right we shouldn't need it, but experimentally we do :-\  Without it, in https://github.com/ManageIQ/manageiq-providers-openshift/pull/28 refresher specs,
I get weird `Invalid single-table inheritance type: ManageIQ::Providers::Openshift::ContainerManager::ContainerTemplate is not a subclass of ManageIQ::Providers::ContainerManager::ContainerTemplate`

This `parent`-sniffing magic is too fragile, I'm gonna replace it with something more explicit (plus give inventory collections unique `.name`s for debugging).  But for now let's get back to green.

PR #82, opened 2017-08-07
Refresher: Split collection from parsing methods

cben:
Concrete benefit is separate `realtime_block` timing.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresher.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresher.rb
@@ -3,14 +3,18 @@ module ManageIQ::Providers::Kubernetes
     include ::EmsRefresh::Refreshers::EmsRefresherMixin
     include ManageIQ::Providers::Kubernetes::ContainerManager::RefresherMixin
 
-    def parse_legacy_inventory(ems)
-      entities = ems.with_provider_connection { |client| fetch_entities(client, KUBERNETES_ENTITIES) }
-      EmsRefresh.log_inv_debug_trace(entities, "inv_hash:")
+    # Full refresh. Collecting immediately. Don't have separate Collector classes.
+    def collect_inventory_for_targets(ems, _targets)
+      inventory = ems.with_provider_connection { |client| fetch_entities(client, KUBERNETES_ENTITIES) }
+      EmsRefresh.log_inv_debug_trace(inventory, "inv_hash:")
+      [[ems, inventory]]
+    end
 
+    def parse_targeted_inventory(ems, _target_is_ems, inventory)
       if refresher_options.inventory_object_refresh
-        ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser.ems_inv_to_inv_collections(ems, entities, refresher_options)
+        ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser.ems_inv_to_inv_collections(ems, inventory, refresher_options)
       else
-        ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser.ems_inv_to_hashes(entities, refresher_options)
+        ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser.ems_inv_to_hashes(inventory, refresher_options)
       end
     end
   end


cben:
similar to https://github.com/ManageIQ/manageiq-providers-openshift/pull/43
I believe they're independent can be merged in either order.

@miq-bot add-label refactoring

app/models/manageiq/providers/kubernetes/container_manager/refresher.rb
```diff
@@ -3,14 +3,18 @@ class ContainerManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
     include ::EmsRefresh::Refreshers::EmsRefresherMixin
     include ManageIQ::Providers::Kubernetes::ContainerManager::RefresherMixin
 
-    def parse_legacy_inventory(ems)
-      entities = ems.with_provider_connection { |client| fetch_entities(client, KUBERNETES_ENTITIES) }
-      EmsRefresh.log_inv_debug_trace(entities, "inv_hash:")
+    # Full refresh. Collecting immediately. Don't have separate Collector classes.
+    def collect_inventory_for_targets(ems, _targets)
```
moolitayer: where is this new function used?

cben: oh sorry forgot to give context.  both are called from:
https://github.com/durandom/manageiq/blob/f9b9bd5fbc/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb#L71-L94
[https://github.com/ManageIQ/manageiq/pull/13728 laid the framework but most of its explanation doesn't apply as I'm not using "Collector", "Target" classes, just defining these 2 methods.]

cben: Previously we were using the naive default implementations where `collect_inventory_for_targets` did nothing
and `parse_targeted_inventory` called `parse_legacy_inventory` that both collected and parsed.

moolitayer: Sorry I still don't get it. Looking at[1] it looks like parse_legacy_inventory is called unless ``collector.kind_of? ManagerRefresh::Inventory::Collector`` which I understand this refresher is not. What am I missing?

[1] https://github.com/manageiq/manageiq/blob/f9b9bd5fbc/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb#L146

moolitayer: ahh I get it now, you implemented the caller of that code :+1:

moolitayer:
LGTM :+1:

moolitayer:
@agrare @Ladas please reivew

app/models/manageiq/providers/kubernetes/container_manager/refresher.rb
cben: right, I replaced that implementation.
(took me an embarassingly long time to realize I can just do it, for a day or two I thought I'll have to subclass Inventory::Collector...)

moolitayer: :+1: I think I even saw a todo there about implementing it in subclasses

Ladas: yeah, this is good, later we will be refactoring this into a common code in the base refresher.

cben:
> logging around fetching each entity type with a count

Will do, fetching is still relatively fast, but it can be helpful debug production logs...
Meanwhile, see https://github.com/ManageIQ/manageiq/pull/15746 for per-collection saving timing,
and https://github.com/ManageIQ/manageiq-gems-pending/pull/255 for not losing timings on errors.

moolitayer:
> Will do, fetching is still relatively fast, but it can be helpful debug production logs...
> Meanwhile, see ManageIQ/manageiq#15746 for per-collection saving timing,
> and ManageIQ/manageiq-gems-pending#255 for not losing timings on errors.

:+1: @cben do we need an issue for the logging item?

moolitayer:
in my comment I was referring to https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/82#pullrequestreview-54679029 . Not a general logging issue.
The purpose is tracking TODO.

PR #84, opened 2017-08-08
Adding test for number_of_containers method for ContainerImage

nimrodshn:
Adding test for `number_of_containers` method for `ContainerImage`.

cc: @enoodle 
@simon3z 

related to: https://github.com/ManageIQ/manageiq/pull/15741, https://github.com/ManageIQ/manageiq-ui-classic/pull/1840


nimrodshn:
Travis fails is depending core PR.

moolitayer:
Why did we choose to add the test only here given that the code is in core?

The reason I'm asking is if someone breaks the code in  ManageIQ/manageiq#15741 Travis will not fail
(at least with the way things are setup currently).

I'm not against having this code here too. (even better, having it in openshift testing a openshift container_image)

But It is more important to have this in ManageIQ/manageiq#15741

nimrodshn:
@moolitayer will move to ManageIQ/ManageIQ :+1:

PR #87, opened 2017-08-08
Use EntitiesMapping in Container Template

zakiva:
We can now use the [entities mapping](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/entities_mapping.rb)  to map the objects kind to a miq class after instantiation. 
Moved from https://github.com/ManageIQ/manageiq/pull/13619
@moolitayer @simon3z Please review

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
@@ -1,18 +1,6 @@
 module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplateMixin
   extend ActiveSupport::Concern
-
-  MIQ_ENTITY_MAPPING = {
-    "Route"                 => ContainerRoute,
-    "Build"                 => ContainerBuildPod,
-    "BuildConfig"           => ContainerBuild,
-    "Template"              => ContainerTemplate,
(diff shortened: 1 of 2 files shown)


moolitayer:
@lfu @zeari can you please review?

app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
```diff
@@ -1,18 +1,10 @@
...
-    "BuildConfig"           => ContainerBuild,
-    "Template"              => ContainerTemplate,
-    "ResourceQuota"         => ContainerQuota,
-    "LimitRange"            => ContainerLimit,
-    "ReplicationController" => ContainerReplicator,
-    "PersistentVolumeClaim" => PersistentVolumeClaim,
-    "Pod"                   => ContainerGroup,
-    "Service"               => ContainerService,
-  }.freeze
+  def model_for_entity(entity)
```
lfu: Just wondering where this method belongs to, here or ```EntitiesMapping```?

zakiva: probably better to put it in `EntitiesMapping` so we can reuse it in other places in the future, I'll move it there.

PR #88, opened 2017-08-11
Remove use_ar_object to speedup the saving

Ladas:
Remove use_ar_object to speedup the saving, it's not needed
anymore due to the latest changes in the saver engine (https://github.com/ManageIQ/manageiq/pull/15761).

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -127,7 +127,6 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           # TODO: should match on digest when available
           :manager_ref    => [:image_ref],
           :delete_method  => :disconnect_inv,
-          :use_ar_object  => true, # several serialized attributes
         )
       )
     # images have custom_attributes but that's done conditionally in openshift parser
@@ -251,7 +250,6 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :builder_params => {:ems_id => manager.id},
           :association    => :container_component_statuses,
           :manager_ref    => [:name],
-          :use_ar_object  => true, # TODO(perf) :message has #<Kubeclient::ComponentStatus health="true"> stringified
         )
       )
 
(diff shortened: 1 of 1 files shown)


Ladas:
@miq-bot add_label enhancement
cc @cben we can avoid using AR in all of these cases now, the core can handle it faster then

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -250,8 +249,8 @@ def initialize_inventory_collections
           :parent         => manager,
           :builder_params => {:ems_id => manager.id},
           :association    => :container_component_statuses,
+          :use_ar_object  => true, # TODO somehow :message attributes gets #<Kubeclient::ComponentStatus health="true">
           :manager_ref    => [:name],
-          :use_ar_object  => true, # TODO(perf) :message has #<Kubeclient::ComponentStatus health="true"> stringified
```
cben: I can't reproduce, in 2 clusters + QE env VCR, all I see are normal strings.
I suppose you'd see this when instead of message there is a nested hash, though [API docs][1] claim it's always a string.
If you have a repro, please let me know.

[1]: https://kubernetes.io/docs/api-reference/v1.7/#componentcondition-v1-core

Ladas: I see this in the current k8s VCR spec

cben: Thanks.  Solved the mystery: I have strong evidence it's always a string and this was a manual mistake in this cassette (unfortunately it's one we had no easy way to rerecord and kept editing :-( ). Fixing cassette in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/90
=> No need handle this specially in refresh.
(We actually don't really need to collect and store this info, it's broken in openshift and we had to hide it in UI...)

Ladas: cool, I'll rebase once the https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/90 is merged

cben:
@agrare can you review?  we want 2 LGTMs per PR so this needs someone who knows what use_ar_object did :)

agrare:
@Ladas can you link to which PRs removed the need for this setting?

Ladas:
@agrare linked it in the description

PR #89, opened 2017-08-11
Make ems_ref of service_port_config unique

Ladas:
Make ems_ref of service_port_config unique by adding :protocol.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -1148,7 +1148,7 @@ module ManageIQ::Providers::Kubernetes
 
     def parse_service_port_config(port_config, service_id)
       {
-        :ems_ref     => "#{service_id}_#{port_config.port}_#{port_config.targetPort}",
+        :ems_ref     => "#{service_id}_#{port_config.port}_#{port_config.targetPort}_#{port_config.protocol}",
         :name        => port_config.name,
         :protocol    => port_config.protocol,
         :port        => port_config.port,


cben:
Sounds good, this is what we wanted to do, but I want to reproduce this in real life.
(Please do ping me if I neglect this!)

simon3z:
@Ladas @cben why not just using the `name`? It's the [unique identifier](https://github.com/kubernetes/kubernetes/blob/db928095a04dbb502898b1d57a8a2940d3ac9d09/pkg/api/types.go#L2776-L2781).

Also, I think that for correctness technically we'd need a migration (maybe we can get away without it, but at least let's consider it).

Ladas:
@cben can we leverage the name here as @simon3z suggests?

cben:
So just `"#{service_id}_#{port_config.name}"`?  Sounds good.

@simon3z in https://github.com/ManageIQ/manageiq/pull/15574 Ladislav concluded we're (sometimes) doing unnecessary deletes/inserts anyway so we can do them one last time, don't require a migration.
Port configs are not a user-visible collection, deletes/inserts should not cause problem.
By that logic, this is also backportable.

I keep thinking "why not first change refresh to ignore ems_ref and use individual parts", then drop ems_ref.  Problem is that `service_id` string passed into parse_service_port_config() is not stored anywhere except in the ems_ref.
Wait, we don't need it, the port config is linked to specific parent service anyway!
Sounds like we can identify in old refersh just by `[:name]` and in graph refresh by `:manager_ref =>
 [:container_service, :name]`?

simon3z:
> @simon3z in ManageIQ/manageiq#15574 Ladislav concluded we're (sometimes) doing unnecessary deletes/inserts anyway so we can do them one last time, don't require a migration.

:+1: 

> Sounds like we can identify in old refersh just by [:name]

Correct me if I'm wrong but it seems what I was suggesting in my previous comment, so :+1:

Ladas:
replaced by:
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/168
https://github.com/ManageIQ/manageiq/pull/16454

PR #90, opened 2017-08-14
Manually edit cassette that I believe was manually misedited

cben:
This cassette shows 1 of 3 componentstatus.message field containing a sub-*hash* `{"health":"true"}` rather than a string.  I can't reproduce that on any live cluster nor cassette, it's always a string.
Evidence it was actually the *string* `'{"health":"true"}'`:

- refresh_parser_spec from [same commit](https://github.com/ManageIQ/manageiq/pull/4891) has `'{...}'` string!
- https://github.com/kubernetes/kubernetes/issues/16721#issue-114767348 by dkorn from same time shows `'{...}'` string
- https://github.com/openshift/origin/issues/5570#issuecomment-153020811 by another person from same time shows `'{...}'` string
- api/types.go had `message String` ever since it was added
  https://github.com/kubernetes/kubernetes/commit/951a125751d5985750c82e919ab151a117bc5a94
- I see other signs this cassette was manually edited, multiple times.

If anyone has doubts, I could do a one-liner parser change to tolerate a hash better.

This does not really affect anything, componentstatus is broken in openshift and [hidden in UI](https://github.com/ManageIQ/manageiq/issues/8572).

@miq-bot add-label test

@yaacov @zakiva Please review.

### DIFF
--- a/spec/vcr_cassettes/manageiq/providers/kubernetes/container_manager/refresher.yml
+++ b/spec/vcr_cassettes/manageiq/providers/kubernetes/container_manager/refresher.yml
@@ -1,3 +1,5 @@
+# Note: this cassette has been manually edited, probably more than once,
+# and doesn't necessarily represent ground truth.
 ---
 http_interactions:
 - request:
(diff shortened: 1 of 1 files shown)


Ladas:
@cben so, is it possible to fetch a new VCR with real data?

cben:
> cben so, is it possible to fetch a new VCR with real data?


Fair question!
Thanks to Zahi's work, we're gradually shifting to new openshift cassettes
with reproduction instructions.
We have a folklore fear of *kubernetes* cassettes as nobody has run MIQ
against a real kubernetes (non openshift) in a long time, we're not

In any case the reason I investigated here, that a new recording won't
easily answer,
was suspicion that some *old* k8s version could returning a non-string.

cben:
@moolitayer ready for review/merge.

cben:
ping @moolitayer this is tiny and blocks another PR.
I'm confident the cassette here was wrong, but it's also a question of principle — is it OK to "discard evidence"?  If you have any concerns, it's trivial to tolerate non-strings better.

moolitayer:
@cben I had all my GH notifications go to spam yesterday (since a bot subscribed me to just about any openshift repository) taking a look now

PR #92, opened 2017-08-15
A spec/support/ directory to be reused by openshift repo

cben:
Specifically will allow
https://github.com/ManageIQ/manageiq-providers-openshift/pull/35 to
reuse `array_recursive_ostruct` helper from here, instead of duplicating it.

@miq-bot add-label test

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -8,17 +8,6 @@ VCR.configure do |config|
   config.cassette_library_dir = File.join(ManageIQ::Providers::Kubernetes::Engine.root, 'spec/vcr_cassettes')
 end
 
-# Helps constructing inputs similar to kubeclient results
-module ArrayRecursiveOpenStruct
-  def array_recursive_ostruct(hash)
-    RecursiveOpenStruct.new(hash, :recurse_over_arrays => true)
-  end
-end
-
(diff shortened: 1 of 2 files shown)


cben:
@Fryguy @bdunne please review. I tried to make this match how core spec/shared/ helpers are reused.

bdunne:
@cben I would expect this to live in `spec/support/` since it's a test helper, not a shared example.

cben:
Aaah, I assumed it's this-repo-vs-used-by-others distinction.
Moved to support/, PTAL.

cben:
@moolitayer can you review/merge?

spec/spec_helper.rb
```diff
@@ -8,17 +8,6 @@
...
-  end
-end
-
-RSpec.configure do |c|
-  c.include ArrayRecursiveOpenStruct
-  c.extend ArrayRecursiveOpenStruct
-end
-
 Dir[Rails.root.join("spec/shared/**/*.rb")].each { |f| require f }
+# This repo's spec/support/ is also used by manageiq-providers-openshift.
```
moolitayer: :+1:

bdunne:
@moolitayer It's an rspec convention

PR #93, opened 2017-08-15
Adding initial changelog with sprint 66 info

JPrause:


moolitayer:
@JPrause do you have automation around generating these?
Is the generation related to milestones?
Asking because we only recently added milestones to all PRs.

moolitayer:
Oh I see mentioned in a link you provided[1] that we should not dump git log into there - so probably same goes for pr messages.

@simon3z Do we have an approach for maintaining this file (and in openshift).
It seems to me we can either
- identify user impacting PRs and request they add a changelog entry
- fetch PR titles, edit and save manually

First option seems less time consuming.

[1] http://keepachangelog.com/en/1.0.0/

JPrause:
@moolitayer yes,..I have a script (actually two) that collects this data for each sprint. However, I see you understand the dilemma of keeping good changelogs.

I like, and would encourage each group to follow your first point about each user identifying which PRs are of interest and have impact for a user reading the changelog. But,...I'm not sure we're there yet,..so I will try to move in that direction by being the evangelist,..and continue to process these changelogs (where and as necessary).

PR #97, opened 2017-08-21
Container Template: convert params to hashes

zakiva:
Depends on https://github.com/ManageIQ/manageiq/pull/15863

Converts the Container Template Parameters that are passed to instantiate method to hashes before passing them to openshift for template processing. The new hashes include only relevant attributes for instantiation instead of the whole active record object that has some irrelevant data. 

Moved from https://github.com/ManageIQ/manageiq/pull/15587

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
@@ -10,7 +10,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplateMixin
                                             :namespace => project
                                           },
                                           :objects    => objects,
-                                          :parameters => params)
+                                          :parameters => params.collect(&:instantiation_attributes))
     create_objects(processed_template['objects'], project)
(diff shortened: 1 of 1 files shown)


zakiva:
@cben @moolitayer @lfu

app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
```diff
@@ -54,4 +54,10 @@ def rollback_object(obj)
       _log.error("Unexpected Exception while deleting object: #{e}")
     end
   end
+
+  private
+
+  def params_to_hashes(params)
+    params.collect { |param| param.attributes.slice("name", "value", "generate", "from", "required") }
```
moolitayer: I would expect this to be encapsulated in ContainerTemplateParameters since objects should encapsulate their properties(and not have to pass params as parameter). 

However I see this was not mentioned in the long discussion in https://github.com/ManageIQ/manageiq/pull/15587 and I would not like for this to wait any longer so @zakiva let me know if this is not working out and we can skip this.

zakiva: done thanks!

PR #98, opened 2017-08-22
Container Template: Allow instantiation with object_labels

zakiva:
Extracted from: https://github.com/ManageIQ/manageiq/pull/15406

Allowing the `instantiate` method to receive labels as an argument so that each created object will be tagged with these labels during instantiation.

cc @moolitayer @simon3z

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
@@ -2,15 +2,17 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplateMixin
   extend ActiveSupport::Concern
   include ManageIQ::Providers::Kubernetes::ContainerManager::EntitiesMapping
 
-  def instantiate(params, project = nil)
+  def instantiate(params, project = nil, labels = nil)
     project ||= container_project.name
+    labels  ||= object_labels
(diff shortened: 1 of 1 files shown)


moolitayer:
@lfu @zeari @cben review please?

app/models/manageiq/providers/kubernetes/container_manager/container_template_mixin.rb
```diff
@@ -2,15 +2,16 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerTemplateMixin
   extend ActiveSupport::Concern
   include ManageIQ::Providers::Kubernetes::ContainerManager::EntitiesMapping
 
-  def instantiate(params, project = nil)
+  def instantiate(params, labels, project = nil)
```
zakiva: @lfu in the original PR (https://github.com/ManageIQ/manageiq/pull/15406#discussion_r125641015) you suggested to add the labels as an optional argument, now with the current changes merged on automation side,  will you prefer that?  I can set here a default value

lfu: @zakiva A default value works for me. So a user who does not want to add more labels won't be forced to specify the label.

zakiva: sounds good

PR #100, opened 2017-08-23
Scanning job will read oscap erros from image-inspector

enoodle:
bz: https://bugzilla.redhat.com/show_bug.cgi?id=1448323

This will change the Job status to "warn" when OpenSCAP fails and set the message that is received from image-inspector to be displayed:

warning_oscap_failure

Now depends on https://github.com/ManageIQ/manageiq-schema/pull/57 `last_scan_results` table
Depends on https://github.com/ManageIQ/manageiq/pull/15967 for the `LastScanResult` model and association with `ContainerImage`

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
@@ -139,13 +139,24 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
       :guest_os      => IMAGES_GUEST_OS
     }
 
-    verify_error = verify_scanned_image_id(image_inspector_client.fetch_metadata)
+    begin
+      inspector_metadata = image_inspector_client.fetch_metadata
+    rescue ImageInspectorClient::InspectorClientException => e
+      _log.error("analyzing image-inspector metadata for #{options[:docker_image_id]} failed with error: #{e}")
+    end
+
+    verify_error = verify_scanned_image_id(inspector_metadata)
     if verify_error
       _log.error(verify_error)
       return queue_signal(:abort_job, verify_error, 'error')
     end
 
-    collect_compliance_data(image)
+    if inspector_metadata.OpenSCAP.Status == 'Error'
+      update!(:options => options.merge(:scan_status         => inspector_metadata.OpenSCAP.Status,
+                                        :scan_result_message => inspector_metadata.OpenSCAP.ErrorMessage))
+    else
+      collect_compliance_data(image)
+    end
 
     image.scan_metadata(SCAN_CATEGORIES,
                         "taskid" => jobid,
@@ -155,7 +166,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
 
   def collect_compliance_data(image)
     unless OpenscapResult.openscap_available?
-      _log.warn "OpenSCAP Binary missing, skipping scan"
+      _log.warn("OpenSCAP Binary missing, skipping scan")
       return nil
     end
     _log.info "collecting compliance data for #{options[:docker_image_id]}"
@@ -230,12 +241,32 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
 
     delete_pod
 
+    set_image_scan_status unless %w(aborting canceling).include?(self.state)
+
   ensure
     case self.state
     when 'aborting' then process_abort(*args)
     when 'canceling' then process_cancel(*args)
-    else queue_signal(:finish, 'image analysis completed successfully', 'ok')
+    else
+      queue_signal(:finish,
+                   target_entity.last_scan_result.scan_result_message,
+                   target_entity.last_scan_result.scan_status)
     end
+  end
+
+  def set_image_scan_status
+    return unless target_entity
+    target_entity.update(
+      :last_scan_result => ScanResult.create(
+        :scan_type           => "OpenSCAP",
+        :scan_status         => options[:scan_status] == 'Error' ? 'warn' : 'ok',
+        :scan_result_message => if options[:scan_status] == 'Error'
+                                  options[:scan_result_message]
+                                else
+                                  'image analysis completed successfully'.freeze
+                                end
+      )
+    )
   end
 
   def finish(*args)
--- a/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
@@ -57,17 +57,38 @@ class MockImageInspectorClient
   end
 
   def fetch_metadata(*_args)
-    if @repo_digest
-      OpenStruct.new('Id' => @for_id, 'RepoDigests' => ["123456677899987765543322", @repo_digest])
-    else
-      OpenStruct.new('Id' => @for_id)
-    end
+    meta = if @repo_digest
+             OpenStruct.new('Id' => @for_id, 'RepoDigests' => ["123456677899987765543322", @repo_digest])
+           else
+             OpenStruct.new('Id' => @for_id)
+           end
+    meta["OpenSCAP"] = OpenStruct.new("Status" => @status)
+    meta
   end
 
   def fetch_oscap_arf
     File.read(
       File.expand_path(File.join(File.dirname(__FILE__), "ssg-fedora-ds-arf.xml"))
     ).encode("UTF-8")
+  end
+end
+
+class MockFailedImageInspectorClient < MockImageInspectorClient
+  def initialize(status, msg, *args)
+    super(*args)
+    @status = status
+    @msg = msg
+  end
+
+  def fetch_metadata(*_args)
+    os = super
+    os["OpenSCAP"] = OpenStruct.new("Status"       => @status,
+                                    "ErrorMessage" => @msg)
+    os
+  end
+
+  def fetch_oscap_arf
+    raise ImageInspectorClient::InspectorClientException.new(404, "test error message")
   end
 end
 
@@ -347,6 +368,38 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
         expect(@job.status).to eq 'error'
         expect(@job.message).to eq "cannot analyze image #{IMAGE_NAME} with id #{IMAGE_ID[0..11]}:"\
                                    " detected ids were #{MODIFIED_IMAGE_ID[0..11]}"
+      end
+    end
+
+    context 'reading openscap messages' do
+      OSCAP_ERROR_MSG = 'Unable to run OpenSCAP: Unable to get RHEL dist number'.freeze
+      OSCAP_SUCCESS_MSG = 'image analysis completed successfully'.freeze
+
+      before(:each) do
+        # Expecting to raise from MockFailedImageInspectorClient before getting to use openscap binary
+        allow(OpenscapResult).to receive_messages(:openscap_available? => true)
+      end
+
+      it 'set the ok status from image-inspector OSCAP' do
+        allow_any_instance_of(described_class).to receive_messages(
+          :image_inspector_client => MockFailedImageInspectorClient.new("Success", "", IMAGE_ID)
+        )
+        @job.signal(:start)
+        expect(@job.state).to eq 'finished'
+        expect(@job.status).to eq 'ok'
+        expect(@job.message).to eq OSCAP_SUCCESS_MSG
+        expect(@image.last_scan_result.scan_result_message).to eq OSCAP_SUCCESS_MSG
+      end
+
+      it 'set the warn status from image-inspector OSCAP' do
+        allow_any_instance_of(described_class).to receive_messages(
+          :image_inspector_client => MockFailedImageInspectorClient.new("Error", OSCAP_ERROR_MSG, IMAGE_ID)
+        )
+        @job.signal(:start)
+        expect(@job.state).to eq 'finished'
+        expect(@job.status).to eq 'warn'
+        expect(@job.message).to eq OSCAP_ERROR_MSG
+        expect(@image.last_scan_result.scan_result_message).to eq OSCAP_ERROR_MSG
       end
     end
 


enoodle:
for some reason the values I assign to `@status` and `@message` are overwritten, but not in the spec runs, only when running through the workers. I am still trying to figure out why.

enoodle:
@roliveri Any idea how I can save information inside a job between signals? I want to save the status of the OpenSCAP report but  to display when finished.

moolitayer:
@enoodle we already save things like the image's id:
https://github.com/enoodle/manageiq-providers-kubernetes/blob/44276daf6b5e72eeaf23ac92dd9ec80fc339561d/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb#L52

There is also a mechanism for error messages - let me know if that's what you need.

enoodle:
The tests are now failing even though they pass on my local machine so to understand I will upload a few versions to try to understand.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -139,7 +140,8 @@ def analyze
       :guest_os      => IMAGES_GUEST_OS
     }
 
-    verify_error = verify_scanned_image_id(image_inspector_client.fetch_metadata)
+    @image_inspector_metadata = image_inspector_client.fetch_metadata
+    verify_error = verify_scanned_image_id(@image_inspector_metadata)
```
moolitayer: I don't think @ is the best mechanism for us here. It works but only because we aren't going through the state machine. maybe we can pass a variable?

In cases were @ is needed in this class, we wrap it in a memoization method that ensure initialization.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -233,7 +243,10 @@ def cleanup(*args)
     case self.state
     when 'aborting' then process_abort(*args)
     when 'canceling' then process_cancel(*args)
-    else queue_signal(:finish, 'image analysis completed successfully', 'ok')
+    else
+      message = options[:oscap_message] || 'image analysis completed successfully'
+      status = options[:oscap_status] || 'ok'
```
moolitayer: I think the status list has to have certain values e.g 'ok', 'error', etc. It might have implications on the job screen. Can you check that?

It makes sense to me to return an error here - partially failed processes don't usually return success.
(cc @cben @simon3z)

enoodle: @moolitayer I will look into those error code. This is intended to return an error here.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -164,6 +166,14 @@ def collect_compliance_data(image)
     openscap_result.save
   rescue ImageInspectorClient::InspectorClientException => e
     _log.error("collecting compliance data for #{options[:docker_image_id]} with error: #{e}")
+    if @image_inspector_metadata.OpenSCAP.Status == "Error"
```
moolitayer: I'm not sure we shouldn't just set `update_message`.
Anyway I think this would look nicer if it was similar to ems_image_inspector_options.
(stored in options like your doing)

enoodle: when `finish` calls `process_finish` it will rewrite the status and message. So changing here will be nice but not enough. I am not sure I under stand the bit about the options. what should be saved there?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: @moolitayer I changed the status to "warn", look at the top comment for how it looks like in the UI

simon3z:
@enoodle is it possible to report this information (e.g. result of last scan) also in the image page?

enoodle:
@simon3z 
This is how it looks in the container image page:
container_image_openscap_not_available

We are currently not saving the last scan status, so if we want to add a row in the UI for that we should add this here.

simon3z:
> We are currently not saving the last scan status, so if we want to add a row in the UI for that we should add this here.

@enoodle yes please let's have a better error reporting in the container image page as well.
Also consider to add an extra column in the list with whether the scan succeeded or failed (see https://github.com/ManageIQ/manageiq-ui-classic/pull/1840)

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
moolitayer: > when finish calls process_finish it will rewrite the status and message. So changing here will be nice but not enough.

We could not set options, set status and message right away on the job, and later on set them only if not set (we would have to reimplement `process_finished`). @enoodle @simon3z  wdyt?

The other suggestion(something cosmetic) is relevant if we go down the other patch of using options.

simon3z: @moolitayer sadly I can't see how the old code looked like here. I am not sure if I'd like to use `options`, I feel that we have several columns already where to report the status `state`, `status`, `message`, `results`. Can you check what other providers are doing?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -139,23 +140,27 @@ def analyze
...
                         "args"   => [YAML.dump(scan_args)])
   end
 
-  def collect_compliance_data(image)
+  def collect_compliance_data(image, inspector_metadata)
     unless OpenscapResult.openscap_available?
-      _log.warn "OpenSCAP Binary missing, skipping scan"
+      _log.warn(OSCAP_BINARY_MISSING_MSG)
+      update!(
+        :options => options.merge(:oscap_status => "warn", :oscap_message => OSCAP_BINARY_MISSING_MSG)
```
simon3z: @enoodle @moolitayer  I am not sure if I'd like to use `options`, I feel that we have several columns already where to report the status `state`, `status`, `message`, `results`. Can you check what other providers are doing? I have the feeling that they need to report some specific error as well.
Anyway please check with the core team, @roliveri @kbrock etc.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: Updating the status and the message is done when the Job changes states and it helps the user see the progress, So I dont think we should update, at least not the message until the Job is finished.  I will answer about the options in the next comment bellow

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: Job has only two other subclasses/descendants: `VmScan` and `ManageIQ::Providers::NativeOperationWorkflow`. They both seem to be using the `options` column to save arguments [1][2][3][4]. 
`VmScan` is updating the status message as it changes states and receives signals, but we are doing it through other functions that we are calling like `image.scan_metadata`. In this PR we are just saving the status of the OpenSCAP scan to display it at the end to the user (otherwise it will be overwritten)

[1]https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_scan.rb#L76
[2]https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_scan.rb#L97
[3]https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_scan.rb#L116
[4]https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/native_operation_workflow.rb#L72

enoodle:
Adding `last_oscap_status` and `last_oscap_message`:
1. Adding columns to ContainerImage
- Pros: Easy to access from both the job and the ContainerImage UI screen, 
- Cons: Schema changes take time to bubble up to releases, another Text field per image might make them take much more space, not available for `VmScan`
2.  Adding `has_one :last_scan_job :class => Job`, preferably to the `ScanningMixin` and using the `options` column the same way as in the first commit.
- Pros: Same easy access, one but with one more scope through the job
- Cons: Will be harder to sync the status for success.


I have updated this PR with the first option (Still have to add code to take the status/message for success). It should be accompanied with https://github.com/enoodle/manageiq-schema/commit/7196117190e44dce3435e812e0a23ef71371c340
@simon3z Is this what you had in mind? I am working on updating this on the UI repo as well.

ilackarms:
@simon3z do we want to pull this into the 4.6 deployment work for testing purposes?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -14,6 +14,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   IMAGE_INSPECTOR_SA = 'inspector-admin'
   INSPECTOR_ADMIN_SECRET_PATH = '/var/run/secrets/kubernetes.io/inspector-admin-secret-'
   PROXY_ENV_VARIABLES = %w(no_proxy http_proxy https_proxy)
+  OSCAP_BINARY_MISSING_MSG = "OpenSCAP Binary missing, skipping scan".freeze
```
ilackarms: why would OpenSCAP be missing?

enoodle: I think it is a travis thing. anyway it is possible and there is a test case for it https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/100/files#diff-0a5ccb8a9fe991f71c0d38d1783ac03eR161

ilackarms: i wonder why it's tested? is this problem limited to travis? I feel like we can assume that OpenSCAP will always exist inside the scanner pod

moolitayer: exactly, this was the topic of discussion back when OpenSCAP was introduced. While the binary was added to the appliance I could not have been added to the Travis machines so it was agreed that tests should run if we have the binary and not run if we don't. Up until now the job was oblivious of this, @enoodle can we keep it this way?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -139,7 +140,10 @@ def analyze
       :guest_os      => IMAGES_GUEST_OS
     }
 
-    verify_error = verify_scanned_image_id(image_inspector_client.fetch_metadata)
+    inspector_metadata = image_inspector_client.fetch_metadata
+    image.update(:last_oscap_status  => inspector_metadata.OpenSCAP.Status,
+                 :last_oscap_message => inspector_metadata.OpenSCAP.ErrorMessage)
```
moolitayer: @enoodle do we have a guard against two ssa running on the same image simultaneously?
This assumes we do

enoodle: Yes, this seem like a race problem. It saves in the image the status of the **last** oscap status and then it will probably display it for both jobs.
If we want to save the openscap status per job, we will have to user the `options`.
I don't think the results can be different. Both jobs will run the same pod at the same time with the same image.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -155,7 +159,9 @@ def analyze
 
   def collect_compliance_data(image)
     unless OpenscapResult.openscap_available?
-      _log.warn "OpenSCAP Binary missing, skipping scan"
+      _log.warn(OSCAP_BINARY_MISSING_MSG)
+      target_entity.update(:last_oscap_status => "Error",
+                           :last_oscap_message => OSCAP_BINARY_MISSING_MSG)
```
moolitayer: Is there any reason this is `last_oscap_message|status` and not `last_scan_message|status`
seems more useful. Idea is status is ok if both openscap & package scan completed, if not then message must be set.
This way we can put there whatever message we need in the future without changing the column name.
If `last_scan_message` is set it should always take precedence over `image analysis completed successfully`.
@enoodle makes sense?

simon3z:
> @simon3z do we want to pull this into the 4.6 deployment work for testing purposes?

@ilackarms no, it's not related to ops.

simon3z:
> @simon3z Is this what you had in mind? I am working on updating this on the UI repo as well.

@enoodle personally I like it but I think this should be reviewed and approved by @roliveri (is this already available with VMs? will VM need something similar in the future?).

I don't want containers to be special in this regard. Maybe the VM implementation will be missing but we should be on the same page with the core team on this.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -233,7 +236,10 @@ def cleanup(*args)
     case self.state
     when 'aborting' then process_abort(*args)
     when 'canceling' then process_cancel(*args)
-    else queue_signal(:finish, 'image analysis completed successfully', 'ok')
+    else
+      message = target_entity.last_scan_message || 'image analysis completed successfully'
+      status = target_entity.last_scan_status == "Error" ? 'warn' : 'ok'
```
moolitayer: Can you pass `warn` last_scan_status and then:
`status = target_entity.last_scan_status || 'ok'`

enoodle: @moolitayer and if the status is not error we would leave it `nil`?

moolitayer: If `target_entity.last_scan_status` is nil then job.status should be 'ok'
```
> nil || 'ok'
 => "ok" 
```

enoodle: @moolitayer `target_entity.last_scan_status` will be either `nil` or `warn` ?

moolitayer: Yes, like you did it in `last_scan_message`. I've realized `last_scan_status` could still be set from a previous job,
so regardless of this comment we would need to handle that in some fashion. One option is that once the job starts we initialize the values.

There are several ways to handle this but I'm thinking we should use job.options while the job is running and only set image.last_X after the job completes. This will avoid hairiness related to: 
- two scans on the same image
- intermittent image.last_x values while the job is running

@enoodle @cben @simon3z thoughts?

enoodle: :+1:

enoodle:
@moolitayer PTAL

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -229,14 +232,30 @@ def cleanup(*args)
 
     delete_pod
 
+    set_image_scan_status
```
moolitayer: reminder we agreed to only do this unless 'aborting' or 'canceling'

moolitayer:
Looks good, one minor comment. @cben ptal

@jrafanie @kbrock can one of you please review given you knowledge of the job infrastructure?

simon3z:
> @enoodle personally I like it but I think this should be reviewed and approved by @roliveri (is this already available with VMs? will VM need something similar in the future?).

@moolitayer @enoodle did we get any feedback by @roliveri here?
cc @chessbyte

roliveri:
@enoodle @simon3z This is not currently available for VMs. It would be useful to support this for VMs too, but we haven't looked into it as of yet.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -229,14 +232,29 @@ def cleanup(*args)
...
+
   ensure
     case self.state
     when 'aborting' then process_abort(*args)
     when 'canceling' then process_cancel(*args)
-    else queue_signal(:finish, 'image analysis completed successfully', 'ok')
+    else
+      queue_signal(:finish,
+                   target_entity.last_scan_message,
+                   target_entity.last_scan_status)
```
jrafanie: I can't tell where this "last_scan_status" ultimately ends up going but is there a validation on possible values?  It looks like the value will be whatever `metadata.OpenSCAP.Status` is.  Is this guaranteed to pass any validations we do on the last_scan_status?

jrafanie:
@moolitayer the job part of this seems fine.  I had a question above @enoodle about the status but looks good so far.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: @jrafanie it was previously set in `set_image_scan_status` to be either `warn` or `ok`. The message though can be arbitrary from `metadata.OpenSCAP.ErrorMessage`

cben:
do you still need me to review?  You folks have a lot of context I don't...

Fryguy:
I merged dependent PR ManageIQ/manageiq#15967 ( @moolitayer @simon3z I didn't see an approval on the other side, but merged anyway since it seemed straightforward.  There is an approval here, but the approval is from Sep 11 and it was just un-WIPed today, so i'm not sure if I should merge this one or not. )

enoodle:
@simon3z @ilackarms Can you take a look? One more approval is needed to merge this.

simon3z:
> @simon3z @ilackarms Can you take a look? One more approval is needed to merge this.

Reassigning this to @moolitayer.
@moolitayer you can merge when you're OK with the PR and the reviews.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -139,30 +139,33 @@ def analyze
...
+    collect_compliance_data(image, inspector_metadata)
 
     image.scan_metadata(SCAN_CATEGORIES,
                         "taskid" => jobid,
                         "host"   => MiqServer.my_server,
                         "args"   => [YAML.dump(scan_args)])
   end
 
-  def collect_compliance_data(image)
+  def collect_compliance_data(image, metadata)
```
moolitayer: current flow:
1. collect metadata with scan result
2. try to parse openscap, if we have an exception we use the collected metadata

We will probably have unexpected results if the exception happens from other reasons (e.g network)
Also we should probably skip the fetch if  `metadata.OpenSCAP.Status` is bad

moolitayer:
@ilackarms @cben @zeari @yaacov needs another reviewer, care to give this one a look?

chessbyte:
@enoodle why did you reset this to WIP?

enoodle:
Mooli's comment made me understand that the first use of
`image_inspector_client ` should also be protected with rescue block.
I am not able to complete this now, so but I am confidant that we will have
it ready on Sunday.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: Image-Inspector is trying not to throw other exceptions https://github.com/moolitayer/image-inspector-client/blob/master/lib/image-inspector-client.rb#L61
I will change to read `metadata.OpenSCAP.Status` before and skip trying to collect OpenSCAP results if there was an error. Will also add a begin-rescue clause around `image_inspector_client.fetch_metadata`

PR #102, opened 2017-08-28
Add prometheus authentication and refactored prometheus client

nimrodshn:
Add Prometheus authentication using prometheus client.
Plus, some refactoring.

cc: @yaacov @cben 
@simon3z

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
@@ -28,7 +28,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Promet
   end
 
   def prometheus_endpoint
-    @ext_management_system.endpoints.find_by(:role => "prometheus")
+    @ext_management_system.connection_configurations.prometheus.endpoint
   end
 
   def prometheus_uri
@@ -52,6 +52,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Promet
   end
 
   def prometheus_try_connect
-    prometheus_client.get("query").kind_of?(Hash)
+    body = prometheus_client.get("query", :query => "ALL").body
+    data = JSON.parse(body)
+    data.kind_of?(Hash)
+  rescue Faraday::SSLError => err
+    raise MiqException::MiqUnreachableError, err.message, err.backtrace
+  rescue StandardError => err
+    raise MiqException::MiqInvalidCredentialsError, err.message, err.backtrace
   end
 end
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -72,6 +72,11 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     client.hawkular_try_connect
   end
 
+  def verify_prometheus_credentials
+    client = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::PrometheusClient.new(self)
+    client.prometheus_try_connect
+  end
+
   # UI methods for determining availability of fields
   def supports_port?
     true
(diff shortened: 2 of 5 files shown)


simon3z:
@nimrodshn @yaacov @moolitayer @cben anything that we can test here?

moolitayer:
> @nimrodshn @yaacov @moolitayer @cben anything that we can test here?

Adding tests for verify_credentials would be beneficial, I don't think we have any.
They can use a cassette or client mocks.

yaacov:
> @nimrodshn @yaacov @moolitayer @cben anything that we can test here?

we can do a test for "prometheus_try_connect" like here:
https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_client_spec.rb#L27

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -28,7 +28,7 @@ def prometheus_client_new(uri, credentials, options)
   end
 
   def prometheus_endpoint
-    @ext_management_system.endpoints.find_by(:role => "prometheus")
+    @ext_management_system.connection_configurations.prometheus.endpoint
```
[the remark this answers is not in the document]
nimrodshn: added tests that will fail when using a saved ems.

nimrodshn:
@moolitayer added tests.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -52,6 +52,10 @@ def prometheus_options
   end
 
   def prometheus_try_connect
-    prometheus_client.get("query").kind_of?(Hash)
+    body = prometheus_client.get("query", :query => "ALL").body
+    data = JSON.parse(body)
+    data.kind_of?(Hash)
+  rescue StandardError
+    false
```
moolitayer: @nimrodshn if you take a look at `verify_credentials` you will see we throw different exceptions for things that can got wrong. You can throw them here and then the user can get a message if the server is unreachable, credentials are bad, etc. (`MiqException::MiqUnreachableError`, `RestClient::Unauthorized`). I'm not sure how ssl errors should be handled, @cben probably knows.

nimrodshn:
@moolitayer PTAL :smiley_cat:

moolitayer:
> @moolitayer PTAL

Let's talk about this

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -52,6 +52,12 @@ def prometheus_options
   end
 
   def prometheus_try_connect
-    prometheus_client.get("query").kind_of?(Hash)
+    body = prometheus_client.get("query", :query => "ALL").body
+    data = JSON.parse(body)
+    data.kind_of?(Hash)
+  rescue Faraday::SSLError => err
+    raise MiqException::MiqUnreachableError, err.message, err.backtrace
+  rescue StandardError => err
```
cben: It's great you're catching everything, we were bitten with SSL and network errors raising all kinds of unexpected exception classes (though it was with rest-client, maybe faraday is saner...).
Perhaps however the catch-all should be "Unreachable", and "InvalidCredentials" should only represent errors server returns over successful HTTP connection?
Did you test what exception you actually get when you pass wrong credentials?

I don't care strongly about this.

yaacov: > maybe faraday is saner...

Nope :-) `Faraday::SSLError` is the error when no one is listening ... ( "unreachable" )
The `StandardError` is actually not a http error, we get an answer but it is not a parsable json, so JSON.parse fail ...

moolitayer:
We can't merge this one. Closing in favor of  #107

PR #104, opened 2017-08-29
add authentication to image inspector requests

ilackarms:
this PR leverages a new security feature in Image Inspector, which allows a custom authentication check using a shared secret.

The shared secret is initialized by passing the pod environment variable `INSPECTOR_AUTH_TOKEN`. HTTP requests send to the image inspector pod should be signed with the header `"X-Auth-Token"` whose value is the shared secret in order to authenticate.

Note that if Image Inspector does not find `INSPECTOR_AUTH_TOKEN` in the environment, it will skip performing this authentication check on HTTP requests.

This PR is a retry of #6, which failed Travis tests for unknown reasons.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
@@ -3,7 +3,7 @@ require 'kubeclient'
 
 class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   PROVIDER_CLASS = ManageIQ::Providers::Kubernetes::ContainerManager
-  INSPECTOR_IMAGE_TAG = '2.1'.freeze
+  INSPECTOR_IMAGE_TAG = '2.2'.freeze
   INSPECTOR_PORT = 8080
   DOCKER_SOCKET = '/var/run/docker.sock'
   SCAN_CATEGORIES = %w(system software)
@@ -14,6 +14,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   IMAGE_INSPECTOR_SA = 'inspector-admin'
   INSPECTOR_ADMIN_SECRET_PATH = '/var/run/secrets/kubernetes.io/inspector-admin-secret-'
   PROXY_ENV_VARIABLES = %w(no_proxy http_proxy https_proxy)
+  INSPECTOR_AUTH_TOKEN = "INSPECTOR_AUTH_TOKEN".freeze
 
   def load_transitions
     self.state ||= 'initializing'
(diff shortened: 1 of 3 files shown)


ilackarms:
@moolitayer @enoodle ptal when you have a chance. thanks :smile:

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -391,13 +396,17 @@ def pod_definition(inspector_admin_secret_name)
...
             :readinessProbe  => {
               "initialDelaySeconds" => 15,
               "periodSeconds"       => 5,
               "httpGet"             => {
-                "path" => "/healthz",
-                "port" => options[:pod_port]
+                "path"        => "/healthz",
+                "port"        => options[:pod_port],
+                "httpHeaders" => {
+                  "name"  => "X-Auth-Token",
```
simon3z: @ilackarms remind me, `X-Auth-Token` is passing through the master proxy?

simon3z:
@moolitayer @ilackarms don't we want to enforce the image-inspector version that supports this? If not let's make explicit the reason for that.

moolitayer:
Maybe we should bump INSPECTOR_IMAGE_TAG accordingly?
(I'm not sure what that means here)

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
ilackarms: yes; this is exactly why we need to use a custom header

ilackarms:
@simon3z that's a good point. i believe `2.2` was the version where this change was released in image inspector?

ilackarms:
we are waiting forhttps://github.com/openshift/origin/pull/15298 for this pr to become relevant

simon3z:
> @simon3z that's a good point. i believe 2.2 was the version where this change was released in image inspector?

@ilackarms check the git history, the releases and dockerhub for availability.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -3,7 +3,7 @@
 
 class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   PROVIDER_CLASS = ManageIQ::Providers::Kubernetes::ContainerManager
-  INSPECTOR_IMAGE_TAG = '2.1'.freeze
+  INSPECTOR_IMAGE_TAG = '2.2'.freeze
```
ilackarms: @simon3z is this the correct tag? or should it be `2.2.0`?

ilackarms:
@simon3z confirmed that `2.2` is the correct tag

simon3z:
@cben can you take a look at this as well?

ilackarms:
@simon3z I can test this with 4.6 deployment

simon3z:
> @simon3z I can test this with 4.6 deployment

@ilackarms OK but remember that we're mostly focused on Prometheus for that effort so give precedence to PRs that are related to that.

cben:
cc @enoodle @bazulay @ourfali

cben:
cc @oourfali

cben:
Closing in favor of rebased #239.  The diff there is exactly the same, except this PR bumps INSPECTOR_IMAGE_TAG to 2.2 and #239 doesn't touch it; we'll first bump the tag directly to 2.4 in #234.

PR #106, opened 2017-08-29
Add parent inventory collections to ensure proper disconnect order

Ladas:
Add parent inventory collections to ensure proper disconnect order
in the case that collection is empty and we can't compute
dependencies.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -28,7 +28,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
         :delete_method  => :disconnect_inv,
       )
     )
-    initialize_custom_attributes_collections(manager.container_projects, %w(labels additional_attributes))
+    initialize_custom_attributes_collections(manager, :container_projects, %w(labels additional_attributes))
 
     @collections[:container_quotas] = ::ManagerRefresh::InventoryCollection.new(
       shared_options.merge(
@@ -73,8 +73,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
         :secondary_refs => {:by_name => [:name]},
       )
     )
-    initialize_container_conditions_collection(manager.container_nodes)
-    initialize_custom_attributes_collections(manager.container_nodes, %w(labels additional_attributes))
+    initialize_container_conditions_collection(manager, :container_nodes)
+    initialize_custom_attributes_collections(manager, :container_nodes, %w(labels additional_attributes))
 
     # polymorphic child of ContainerNode & ContainerImage,
     # but refresh only sets it on nodes.
@@ -143,8 +143,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :delete_method        => :disconnect_inv,
         )
       )
-    initialize_container_conditions_collection(manager.container_groups)
-    initialize_custom_attributes_collections(manager.container_groups, %w(labels node_selectors))
+    initialize_container_conditions_collection(manager, :container_groups)
+    initialize_custom_attributes_collections(manager, :container_groups, %w(labels node_selectors))
     @collections[:container_volumes]      =
       ::ManagerRefresh::InventoryCollection.new(
         shared_options.merge(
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -321,7 +321,8 @@ def initialize_inventory_collections
   end
 
   # ContainerCondition is polymorphic child of ContainerNode & ContainerGroup.
-  def initialize_container_conditions_collection(relation)
+  def initialize_container_conditions_collection(manager, association)
```
moolitayer: Are you using `manager` here?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -347,9 +349,10 @@ def initialize_custom_attributes_collections(relation, sections)
       @collections[[:custom_attributes_for, relation.model.base_class.name, section.to_s]] =
         ::ManagerRefresh::InventoryCollection.new(
           shared_options.merge(
-            :model_class => CustomAttribute,
-            :arel        => query,
-            :manager_ref => [:resource, :section, :name],
+            :model_class                  => CustomAttribute,
+            :arel                         => query,
+            :manager_ref                  => [:resource, :section, :name],
+            :parent_inventory_collections => [association]
```
moolitayer: I'm a little confused, shouldn't we be passing manager here?

cben: After we had a problem here with STI, I've wanted to make these collections creation/use less magic and more explicit — though I forgot what exactly I planned :-/

Here, using `[association]` assumes the `foos` in `ems.foos` will matches that inventory collection's `.name`.  Collection names are currently all implicitly derived from the model, so this is like 2 assumptions lining up...
=> I'd rather this function took the parent collection as argument and used `[parent.name]` here.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
cben: Yes, line below.  Instead of taking `ems.foos` the function now takes separte `ems, :foos` and then calls `ems.foos`.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
cben: If you look at https://github.com/ManageIQ/manageiq/pull/15903, `parent_inventory_collections` expects symbols — names of other InventoryCollections being saved.
It's just a coincidence by convention that our `ems.foos` rails association names match model names like `Foo` and InventoryCollection of Foo defaults to having `.name == :foos`.  So `association` seems to work here, but as I say in my other comment, I'd rather use something more explicit.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: right, and I should Right and I should use it also for parent_inventory_collections

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: right, that would make more sense

cben: friendly ping, I'd like to land these fixes.  I can make the changes if you prefer.

moolitayer: @cben is this ready? can you review?

Ladas: @cben gah, I totally forgot, I am in the middle of rollup testing now and we have public holiday tomorrow, and Fri PTO, so I will be back on Monday. So if you can make the change in the meantime, great :-) Otherwise I will tackle it somewhere next week. :-)

Ladas: @cben damn, do you know how to get name from the association?

```
2.4.2 :017 > pp container_ems.containers.name
"Container"
```

cben: See #134 — what do you think?

Ladas:
replaced by https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/134

PR #113, opened 2017-09-15
Move log setting to provider

moolitayer:
Needed for https://github.com/ManageIQ/manageiq/pull/15978


moolitayer:
@simon3z please review/merge

PR #118, opened 2017-09-26
Fix net_usage_rate_average units

yaacov:
**Description**

We collect `average kb usage per second` from the external metrics data base.
When we insert data in the internal metrics data base we declare it to be 0.5 of `average datagrams per second`.

This may lead to wrong data presented to the end user, for example, if we collect 80 kb/sec the user may see 160 datagrams/sec [the *2 is for 0.5 to value conversion]

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1495733

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -42,9 +42,9 @@ module ManageIQ::Providers
         :counter_key           => "net_usage_rate_average",
         :instance              => "",
         :capture_interval      => INTERVAL.to_s,
-        :precision             => 2,
+        :precision             => 1,
         :rollup                => "average",
-        :unit_key              => "datagramspersecond",
+        :unit_key              => "kilobytespersecond",
         :capture_interval_name => "realtime"
       }
     }


yaacov:
@miq-bot add-labels bug

@zeari @cben @moolitayer @ilackarms @simon3z @gshefer please review

yaacov:
@Ladas please review

yaacov:
@cben please review

cben:
If one already had metrics and then upgrades to this change, will it fix
them retroactively, or result in mixed history?
Does this affect collection/processing or only display/reporting?

yaacov:
> If one already had metrics and then upgrades to this change, will it fix them retroactively, or result in > mixed history?

> Does this affect collection/processing or only display/reporting?

We are not sure it even affect reporting :-) @gshefer and @zeari are researching this issues, @gshefer is searching the processing/reporting side, and @zeari the customer / history issues.

> ... result in mixed history?

@cben - This is a good question, do you suggest we should not fix this BZ for backward comparability  reasons ?

Ladas:
@cben mixed history, the samples and rollups are already stored in the db. So we should have a data migration that will fix the data. (though we can't backport it)

yaacov:
@ladas hi, from the mails, I understand that the `:precision` field is used on data that originate from `vmware` collector, do we know that it is used on data originating from `containers` collector ?

Ladas:
@yaacov @cben best if you guys would wrote integrations specs to be sure (by collecting know set of samples, we would check we store expected values into the DB)

I am 95% sure it hits containers, since this code is common to all providers. ;-)

yaacov:
@Ladas answering myself :-) 

> do we know that it is used on data originating from containers collector ?

yes, it affect all metrics [1]

[1] https://github.com/ManageIQ/manageiq/blob/master/app/models/metric/ci_mixin/processing.rb#L190

yaacov:
> best if you guys would wrote integrations specs to be sure (by collecting know set of samples, we would check we store expected values into the DB)

@gshefer is working on that

yaacov:
@cben @Ladas -

a. The migration discussion is irrelevant for this PR, we should continue it in the mailing list.
b. We need this fix regardless of way we will offer/suggest to affected customers a way to migrate, and warn them about bad data in previous chargeback reports, this requires PM intervantion.

cben:
Hmm, I'm afraid it's very relevant.
IIUC once user runs a version including this PR, new metrics collected mean
something different from previously existing, and there is no way to know
which was which.
After that, migration is impossible!

- If you ignore people running master, you could say it's OK as long as we
come up with migration before next major release...
  I'd rather we waited and merged together with migration but oh well.

- This should not be backported, old version need another approach.

yaacov:
@cben hi, this is my 2 cents on the matter:

> This should not be backported, old version need another approach.

  - Mistake in reports should be transparent to users and made public immediately,

    We did this by reporting the BZ:  https://bugzilla.redhat.com/show_bug.cgi?id=1495733

  - Mistake in reports should be fixed immediately:
    a. New users will have correct data.
    b. Current users could have the option to upgrade to a correct system, in the time they chose.

    We did this by merging the fix as soon as possible into the master branch of our development repository.

> I'd rather we waited and merged together with migration but oh well.

  - Reports should be un-mutable and trusted, same as paper reports. Two reports made by the system in different times should have the same data, even if it include a mistake.

    a. We should inform users that reports made by a pre-patched system contain an error.
    b. We should not migrate data, as this will violate the un-mutable property of reports.

yaacov:
https://bugzilla.redhat.com/show_bug.cgi?id=1511142 

target 5.8.3

@miq-bot add-labels fine/yes

simaishi:

```
$ git log -1
commit f51bbb92b6f8ad1559c908884ec99068ff38f49d
Author: Federico Simoncelli <federico.simoncelli@gmail.com>
Date:   Wed Sep 27 09:40:36 2017 +0200

    Merge pull request #118 from yaacov/fix-net_usage_rate_average-units
    
    Fix net_usage_rate_average units
    (cherry picked from commit 851fb594922bda6a4daf676da5e4ebfea340fb9c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1511142
```

PR #119, opened 2017-09-27
Fail on creating scanning job with image instance

zakiva:
Depends on https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/105. 

This PR removes the backward compatibility kept in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/105. In case of creating a scanning job with image instance (instead of class) an explicit exception will be raised.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -174,12 +174,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
   end
 
   def raw_scan_job_create(target_class, target_id = nil, userid = nil, target_name = nil)
-    # maintain backward compatibility pre https://github.com/ManageIQ/manageiq/pull/13722
-    if target_class.kind_of?(ContainerImage)
-      target_id = target_class.id
-      target_name = target_class.name
-      target_class = target_class.class.name
-    end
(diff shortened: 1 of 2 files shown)


zakiva:
@simon3z @moolitayer @cben

zakiva:
@miq-bot add_label fine/no

simon3z:
@zakiva @moolitayer where are the `message-buffer` travis failure coming from?

zakiva:
@simon3z I don't think it's related to this PR, it fails on master too.

zakiva:
@simon3z I don't know but I think it was introduced in https://github.com/ManageIQ/manageiq/pull/16024. 
cc @jrafanie

jrafanie:
> @simon3z I don't know why exactly, but I think it was introduced in ManageIQ/manageiq#16024.
> Should we just require 'application_helper' in this repo?
> cc @jrafanie

@zakiva If you need application_helper in this repo, please require it here.  We want to detangle 🍝  dependencies in the main repo so we can't put it back there.

jrafanie:
@zakiva in the main repo, you can now run `BUNDLER_GROUPS=manageiq_default bundle exec rspec path/to/spec` and it won't include the manageiq-classic-ui code and won't parse the enormous routes file... same with rails console.  The idea is we want to break apart dependencies even further so we need to specify dependencies correctly at the places we use them.

PR #120, opened 2017-09-27
adding cve_url and image_tag to global image_inspector settings

enoodle:
This will use thes settings as global options when no per-provider options are set

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1459555

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/options.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/options.rb
@@ -33,12 +33,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::Options
             :repository  => {
               :label          => N_('Image-Inspector Repository'),
               :help_text      => N_('Image-Inspector Repository. example: openshift/image-inspector'),
-              :global_default => Settings.ems.ems_kubernetes.image_inspector_repository,
+              :global_default => ::Settings.ems.ems_kubernetes.image_inspector_repository,
             },
             :registry    => {
               :label          => N_('Image-Inspector Registry'),
               :help_text      => N_('Registry to provide the image inspector repository. example: docker.io'),
-              :global_default => Settings.ems.ems_kubernetes.image_inspector_registry,
+              :global_default => ::Settings.ems.ems_kubernetes.image_inspector_registry,
             },
             :image_tag   => {
               :label          => N_('Image-Inspector Tag'),
@@ -46,11 +46,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::Options
               :global_default => ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job::INSPECTOR_IMAGE_TAG,
             },
             :cve_url     => {
-              :label     => N_('CVE location'),
-              :help_text => N_('Enables defining a URL path prefix for XCCDF file instead of accessing the default location.
+              :label          => N_('CVE location'),
+              :help_text      => N_('Enables defining a URL path prefix for XCCDF file instead of accessing the default location.
   example: http://my_file_server.org:3333/xccdf_files/
   Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
               # Future versions of image inspector will extend this.
+              :global_default => ::Settings.ems.ems_kubernetes.image_inspector_cve_url,
             },
           }
         }
--- a/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
@@ -445,8 +445,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job < Job
   end
 
   def add_cve_url(pod_def)
-    if ems_image_inspector_options.key?(:cve_url)
-      pod_def[:spec][:containers][0][:command].append("--cve-url=#{ems_image_inspector_options[:cve_url]}")
-    end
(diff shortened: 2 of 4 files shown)


config/settings.yml
```diff
@@ -8,6 +8,8 @@
     :miq_namespace: management-infra
     :image_inspector_registry: docker.io
     :image_inspector_repository: openshift/image-inspector
+    :image_inspector_image_tag: 2.1
```
simon3z: @enoodle I have mixed feelings about exposing this one. The image tag is so tied to the code we're developing and shipping that I am not sure it should be configurable at all.
I understand that once you give the `image_inspector_registry` and `image_inspector_repository` flexibility there is already lot of space for mistakes, but I don't want to give the impression here that the version change is something we can cope with (because we can't).
cc @cben

enoodle: @simon3z I removed it then. I think having it per-provider will be enough for most users anyway.

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -445,8 +445,7 @@ def inspector_proxy_env_variables
   end
 
   def add_cve_url(pod_def)
-    if ems_image_inspector_options.key?(:cve_url)
-      pod_def[:spec][:containers][0][:command].append("--cve-url=#{ems_image_inspector_options[:cve_url]}")
-    end
+    cve_url = ems_image_inspector_options.fetch_path(:cve_url) || ::Settings.ems.ems_kubernetes.image_inspector_cve_url
+    pod_def[:spec][:containers][0][:command].append("--cve-url=#{cve_url}") unless cve_url.blank?
```
moolitayer: @enoodle so we always append `--cve-url` now?
What if the value of both options is nil ?

enoodle: unless the value of `cve_url` is blank. If the value of both options will be empty it won't be added. I will add a test for this case.

app/models/manageiq/providers/kubernetes/container_manager/options.rb
```diff
@@ -46,11 +46,12 @@ def advanced_settings
...
             },
             :cve_url     => {
-              :label     => N_('CVE location'),
-              :help_text => N_('Enables defining a URL path prefix for XCCDF file instead of accessing the default location.
+              :label          => N_('CVE location'),
+              :help_text      => N_('Enables defining a URL path prefix for XCCDF file instead of accessing the default location.
   example: http://my_file_server.org:3333/xccdf_files/
   Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
               # Future versions of image inspector will extend this.
+              :global_default => Settings.ems.ems_kubernetes.image_inspector_cve_url,
```
moolitayer: `::Settings`

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
moolitayer: ahh right, missed that.

BTW do we have a test using the provider setting? I can only see one using the global one

moolitayer:
@enoodle I think we should have two tests:
- setting the global value => should use global
- set global and per provider => should use provider.

BTW we are setting the per provider value to the global one by default?
so that means usually the code will return the default from the per provider one right?
(seems fine)

enoodle:
> BTW we are setting the per provider value to the global one by default?
so that means usually the code will return the default from the per provider one right?
(seems fine)

No, the per provider option is empty by default, so that we know we should take the "Settings" default. (Other wise updating `::Settings`  would not update default..)

moolitayer:
ahh thanks. So what is the meaning of the `global_default` directive?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
enoodle: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/120/files#diff-676c4ca56bd15c7755ac3a9b6037d3eeR210

enoodle:
The `advanced_options` function is used to describe the options but the globals are not used through it. The options themselves are saved per provider while this function is a class method.
Maybe we should be using the globals from this function but currently we are not.

app/models/manageiq/providers/kubernetes/container_manager/options.rb
enoodle: @moolitayer Why is this important? with the rest of the options I didn't use `::` but maybe I should do this with all of them?

moolitayer: IT's not that important, we should get used to having all references absolute.

spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
```diff
@@ -207,13 +207,39 @@ def create_pod_definition
         expect(pod[:spec][:containers][0][:env][0][:value]).to eq(my_value)
       end
 
-      it 'should send cve_url from options' do
+      it 'should send cve_url from options over global' do
+        stub_settings_merge(:ems => {:ems_kubernetes => {:image_inspector_cve_url => "from_global" }})
```
moolitayer: :+1:

simon3z:
@moolitayer can you approve this if it's OK with you?

moolitayer:
@enoodle there seems to be a styling problem

enoodle:
@moolitayer @simon3z This seems ready now

PR #121, opened 2017-09-27
Drop Container Component Statuses

zakiva:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1485431 
Schema: https://github.com/ManageIQ/manageiq-schema/pull/79
Core: https://github.com/ManageIQ/manageiq/pull/16055
UI: https://github.com/ManageIQ/manageiq-ui-classic/pull/2246

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
@@ -5,9 +5,7 @@ module ManageIQ
         KUBERNETES_ENTITIES = [
           {:name => 'pods'}, {:name => 'services'}, {:name => 'replication_controllers'}, {:name => 'nodes'},
           {:name => 'endpoints'}, {:name => 'namespaces'}, {:name => 'resource_quotas'}, {:name => 'limit_ranges'},
-          {:name => 'persistent_volumes'}, {:name => 'persistent_volume_claims'},
-          # workaround for: https://github.com/openshift/origin/issues/5865
-          {:name => 'component_statuses', :default => []}
+          {:name => 'persistent_volumes'}, {:name => 'persistent_volume_claims'}
         ]
 
         def fetch_entities(client, entities)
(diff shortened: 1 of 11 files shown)


simon3z:
@zakiva IIRC there's a BZ to link, right?

zakiva:
> @zakiva IIRC there's a BZ to link, right?

Just added it :)

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
```diff
@@ -5,9 +5,7 @@ module ContainerManager::RefresherMixin
         KUBERNETES_ENTITIES = [
           {:name => 'pods'}, {:name => 'services'}, {:name => 'replication_controllers'}, {:name => 'nodes'},
           {:name => 'endpoints'}, {:name => 'namespaces'}, {:name => 'resource_quotas'}, {:name => 'limit_ranges'},
-          {:name => 'persistent_volumes'}, {:name => 'persistent_volume_claims'},
-          # workaround for: https://github.com/openshift/origin/issues/5865
-          {:name => 'component_statuses', :default => []}
+          {:name => 'persistent_volumes'}, {:name => 'persistent_volume_claims'}
```
cben: nit: do you mind keeping the trailing comma? (not everyone likes them, your call)

moolitayer:
AFAIU this patch should be independant

simon3z:
> AFAIU this patch should be safe to merge

@moolitayer feel free to merge (sync with @zakiva ) when ready if decoupled from the other PRs.

zakiva:
@moolitayer yeah this should be safe to merge now, actually, it has to be merged before we remove the class/table in the main and schema repos. 
(UI also should be safe).

app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
zakiva: I don't like it :) 
(and it wasn't there in the first place)

moolitayer:
:scissors: :fire: :scissors: :fire: :scissors:

PR #122, opened 2017-09-27
dont return an empty set when something goes wrong

durandom:
an empty set means there is no data and the platform will not
try to collect the data again. Unless a manual gap detection is
performed.

see https://github.com/ManageIQ/manageiq/blob/91d3ef4189e8b9e9d2225b758082d52f688926da/app/models/metric/ci_mixin/capture.rb#L193-L195
and
https://github.com/ManageIQ/manageiq/blob/91d3ef4189e8b9e9d2225b758082d52f688926da/app/models/metric/ci_mixin/capture.rb#L56-L58

fixes
https://bugzilla.redhat.com/show_bug.cgi?id=1411894

@yaacov please have a look if that makes sense for you
cc @agrare

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -68,7 +68,7 @@ module ManageIQ::Providers
         ems.try(:update_attributes,
                 :last_metrics_error       => :invalid,
                 :last_metrics_update_date => Time.now.utc)
-        return [{}, {}]
+        raise
       end
 
       Benchmark.realtime_block(:collect_data) do
@@ -78,7 +78,7 @@ module ManageIQ::Providers
           _log.error("Hawkular metrics service unavailable: #{e.message}")
           ems.update_attributes(:last_metrics_error       => :unavailable,
                                 :last_metrics_update_date => Time.now.utc) if ems
-          return [{}, {}]
+          raise
         end
       end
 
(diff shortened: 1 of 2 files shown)


durandom:
https://github.com/ManageIQ/manageiq/pull/16042 was my first attempt to address this BZ

yaacov:
The fail test are exactly what we want to change :-) so we just need to change the tests to "not failing silently" instead of "fail silently"

yaacov:
^^

the lines [1] of the three fail tests:
`expect(@node.perf_collect_metrics('interval_name')).to eq([{}, {}])`
should be:
`expect(@node.perf_collect_metrics('interval_name')).to raise_error`

and the name should also change to "not silently" :-)

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb#L30

yaacov:
p.s.

the reason for the fail quietly was to allow our internal testing scripts we used to populate metrics to run without failing on missing metrics. so no problem to rais error here from our side.

durandom:
@agrare can you give a quick 👍 as well?
@simon3z can you merge afterwards?

simon3z:
> @simon3z can you merge afterwards?

@durandom definitely. thanks!

agrare:
ping @simon3z

durandom:
FINE BZ is  https://bugzilla.redhat.com/show_bug.cgi?id=1496938

simaishi:

```
$ git log -1
commit f5c8cfb27642a8ab4dd4c3ed77d0c8f420034822
Author: Federico Simoncelli <federico.simoncelli@gmail.com>
Date:   Thu Sep 28 00:05:23 2017 +0200

    Merge pull request #122 from durandom/perf_collect_raise
    
    dont return an empty set when something goes wrong
    (cherry picked from commit 9cf161bf3af9867737eaee4deb026252934d2321)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1496938
```

PR #123, opened 2017-09-29
Prefix ems_ref with object type in event hash

agrare:
The event handler expects the ems_ref to be prefixed by the object type,
not using the raw ems_ref.  This is causing issues adding a real ems_ref
for an event because the process_container_entities_in_event! method is
clearing the key.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/event_parser_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/event_parser_mixin.rb
@@ -4,7 +4,16 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EventParserMixin
   included do
     def self.event_to_hash(event, ems_id = nil)
       _log.debug("ems_id: [#{ems_id}] event: [#{event.inspect}]")
-      {
+      ems_ref_key = case event[:kind]
+                    when 'Node'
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/container_manager/event_parser_mixin.rb
```diff
@@ -4,7 +4,16 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EventParserMixin
   included do
     def self.event_to_hash(event, ems_id = nil)
       _log.debug("ems_id: [#{ems_id}] event: [#{event.inspect}]")
-      {
+      ems_ref_key = case event[:kind]
```
moolitayer: Consider moving this above line 30

moolitayer:
@agrare The fix makes sense. What I don't get is what Is a real ems_ref?
Is that an ems_ref for the event (not the event target)? Where is that value stored?

Do we have a bug or issue for this or is it for something under development?

cben:
https://github.com/ManageIQ/manageiq/issues/16074

cben:
IIUC this better be merged simultaneously with https://github.com/ManageIQ/manageiq/pull/16076

PR #125, opened 2017-10-01
Require 'faraday_middleware' in monitoring_manager_mixin.rb

zakiva:
Fixes https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/131

Test in `spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb` fails since the require statement was removed from the main repo in https://github.com/ManageIQ/manageiq/pull/16024, so we should require it in this repo.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
@@ -1,3 +1,5 @@
+require 'faraday_middleware'
+
 module ManageIQ::Providers::Kubernetes::MonitoringManagerMixin
   extend ActiveSupport::Concern
   ENDPOINT_ROLE = :prometheus_alerts
--- a/spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb
@@ -41,6 +41,7 @@ describe ManageIQ::Providers::Kubernetes::MonitoringManager do
         described_class.name.underscore,
         # :record => :new_episodes,
       ) do
+        # in case of error try to check monitoring_manager.authentication_check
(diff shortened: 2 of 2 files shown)


zakiva:
@simon3z @moolitayer @cben

moolitayer:
@zakiva Application_helper is coming from the UI repo right? if so we should avoid creating that dependency
Do we know why we need the helper?

Fryguy:
Yeah, this should not really be necessary and if it is, we need to decouple it... cc @blomquisg @durandom

Fryguy:
cc @jrafanie

jrafanie:
yeah, the reason we removed the application_helper from the main repo is because it wasn't used anymore.  I can't imagine any backend provider code needing application_helper.

@zakiva can you create an issue with the error that happens when you run the test without application_helper?  That should be decoupled.

zakiva:
> @zakiva can you create an issue with the error that happens when you run the test without application_helper? That should be decoupled.

@jrafanie opened https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/131

zakiva:
As @cben wrote in the related issue we were missing `faraday`, so requiring only that (instead of all application_helper) will make the test pass.

app/models/manageiq/providers/kubernetes/monitoring_manager_mixin.rb
```diff
@@ -1,3 +1,5 @@
+require 'faraday_middleware'
```
cben: cc @moolitayer is this the best place ?  maybe this file only needs `faraday` but `faraday_middleware` belongs in message_buffer_client.rb ?

To me this is Good Enough :+1:

zakiva:
This was fixed in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/40, closing.

PR #128, opened 2017-10-12
Name custom attributes ICs nicely

Ladas:
Name custom attributes ICs nicely, so it shows up nicely in log and the graph.


app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -338,6 +338,7 @@ def initialize_custom_attributes_collections(manager, association, sections)
       @collections[[:custom_attributes_for, relation.model.base_class.name, section.to_s]] =
         ::ManagerRefresh::InventoryCollection.new(
           shared_options.merge(
+            :name                         => "#{association}_#{section}".to_sym,
```
cben: can you also add name in initialize_container_conditions_collection?

Ladas: yes, good idea

Ladas:
@moolitayer @cben both custom attributes and conditions have name now

moolitayer:
cc @cben

PR #129, opened 2017-10-13
Add an Inventory collector worker for update-driven pod collection

agrare:
This adds a new worker which uses update-driven `client.watch_pods()` and `ManagerRefresh::Target` to send pod updates with inventory to the `RefreshWorker` to be saved, ensuring that no pods will be missed by `EmsRefresh` if the pod is created and deleted between refreshes.

Depends on: ~~https://github.com/ManageIQ/manageiq/pull/16198~~


agrare:
This is still very WIP, I think we need the watch in its own thread so that we still heartbeat even if no changes are happening.

Also I had to connect to `ems.connect(:service => "kubernetes", :path => '/api')` otherwise I didn't have a `watch_pods` method and `watch_entities('pods')` threw a 404.

agrare:
cc @Ladas

agrare:
> we will need the same for OpenShift repo

One step ahead of you ;) https://github.com/ManageIQ/manageiq-providers-openshift/pull/52

Ladas:
@cben if you are interested, this is first wip that will be able to save these targets

https://github.com/Ladas/manageiq-providers-kubernetes/commit/63a3be0e7adebd11595c31a9caf51eb8264e7dae

https://github.com/Ladas/manageiq-providers-openshift/commit/da91bdfd56c5d292bb33552f115ab4ccddfa2e74

It uses hardcoded data, that @agrare will be sending, but it is able to save them. For now I am fetching also all nodes and projects (we should be able to filter only referenced from the API?), so we can connect the pods to them.

And it only creates&updates, so we don't need to limit the queries, the price is more queries to the DB.

But in the end, we can have a working version pretty fast I think, then we can optimize it to do minimal API requests and queries to the DB.

Also I am reusing the same parser, since we can convert the watches data into format we got from the API.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker/runner.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  private
+
+  def connection
+    @connection ||= ems.connect(:service => "kubernetes", :path => '/api')
+  end
+
+  def parse_notice_pod_ems_ref(pod)
+    pod.metadata.uid
```
cben: is uid only part you parse?  or will you replace it with calling parser later?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker/runner.rb
```diff
@@ -0,0 +1,40 @@
...
+        :association => :container_groups,
+        :manager_ref => ems_ref,
+        :options     => {
+          :payload => notice.object,
+        },
+      )
+
+      EmsRefresh.queue_refresh(target)
+
+      heartbeat
```
cben: this will heartbeat irregularly, can pass arbitrarily long time until next notice arrives — is that ok?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker/runner.rb
cben: ah, I see, `:payload => notice.object` smuggles the whole object, and the parsing meat is the [other branch](https://github.com/Ladas/manageiq-providers-kubernetes/commit/63a3be0e7adebd11595c31a9caf51eb8264e7dae#diff-0324981fdb3019ce6d98f9c86d97f2bbR59)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker/runner.rb
agrare: Yeah that's what I'm worried about, if there is no way to specify a timeout to `#watch` we'll need a heartbeat thread which also checks if this thread is running, similar to what the event catcher does.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker/runner.rb
agrare: Yeah exactly, the goal is to use this to just skip the collect step and the parse+save is still done in the refresh worker

agrare:
@cben do you know what's up with this? [ref](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/129#issuecomment-336482354)

> Also I had to connect to ems.connect(:service => "kubernetes", :path => '/api') otherwise I didn't have a watch_pods method and watch_entities('pods') threw a 404.

cben:
`ems.connect(:service => "kubernetes", :path => '/api')` is correct (though I think one of those should be enough).

We have a Liskov violation where an openshift EMS .connect defaults to `/oapi`, differently from kubernetes EMS :-(  Better always specify it.

agrare:
Okay thanks :D I thought the openshift service would have all the methods from the kube service also.
I'll use just `:service => "kubernetes"` instead of specifying the path directly, :+1: thanks!

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -0,0 +1,96 @@
...
+    while @queue.length > 0
+      object = queue.deq
+      ems_ref = parse_notice_pod_ems_ref(object)
+
+      targets << ManagerRefresh::Target.new(
+        :manager     => ems,
+        :association => :container_groups,
+        :manager_ref => ems_ref,
+        :options     => {
+          :payload => BinaryBlob.create!(:binary => object.to_json).id,
```
Ladas: ah, you do it here, I thought that would be in the generic code...

Ladas: I would rather separate this and solve ```effective saving of payload``` as another problem

agrare:
Not sure why codeclimate says `exit_requested` is an unused variable...it is an instance variable defined by `attr_accessor`

Fryguy:
rubocop is correct...you are setting a local variable and not the accessor.  You need `self.exit_requested =` to set via the accessor.

agrare:
:man_facepalming: thanks

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    sleep_poll_normal
+  end
+
+  def before_exit(message, _exit_code)
+    safe_log("#{message} Stopping Watch Thread.")
+
+    stop_watch_thread
+
+    unless @watch_thread_tid.nil?
```
cben: tid is typo? (also below)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -0,0 +1,127 @@
+require 'thread'
+
+module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollectorMixin
+  attr_reader   :queue
+  attr_accessor :exit_requested
+  private :queue, :exit_requested
```
cben: I dont remember seeing any private accessors in MIQ before.  Consider just using instance vars, for consistency.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -0,0 +1,127 @@
...
+    _log.info("Started watch thread")
+
+    tid
+  end
+
+  def watch_thread
+    pod_watch_stream.each do |notice|
+      break if exit_requested
+
+      _log.info("EMS [#{ems.id}] Pod: #{parse_notice_pod_ems_ref(notice.object)}")
```
cben: If this goes to evm.log, there is no implied context that this is from watch thread (unless you look up thread #).
Please add some words eg. "received change in pod"

cben:
Can you add setting to not even start this worker? (Or did I miss it elsewhere?)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
Ladas: I use them all the time. It's always better to use object interface over accessing it directly. ;-)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
Ladas: thread id? :-)

cben: right, but you called same var/attr `watch_thread_id` elsewhere?
(btw, it's not really an id, it's a Thread object.  however the comparable [base EventCatcher](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/event_catcher/runner.rb) uses `tid` terminology so whatever...)

agrare: Good catch, was supposed to be `watch_thread_id`, `tid` was a holdover from the event catcher

agrare:
@cben it is disabled by default in the base class (see [ManageIQ/manageiq/16198 config/settings.yml](https://github.com/ManageIQ/manageiq/pull/16198/files#diff-646ad6b10917e4385cbcc8c8524e24a2R1062) and it is enabled in the openshift PR https://github.com/ManageIQ/manageiq-providers-openshift/pull/52/files#diff-646ad6b10917e4385cbcc8c8524e24a2R30
I'll add it as enabled here as well for consistency.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
agrare: :+1: done, the class and method names are already there but I made it more obvious that it was catching an update.
I also changed it to debug loglevel since a highly active system could flood the logs with changes, thoughts?

Ladas: So e.g. in AWS, we log every event (that will be stored to DB). So it is quite verbose for log, but also handy for debugging. :-)

agrare:
@cben addressed all of your comments, PTAL :)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
cben: I think each notice will trigger many lines for the queuing and doing the refresh, so info is also ok.  (and at this stage of maturity, IMHO debugging is important :-)

agrare: info it is then :+1:

agrare:
@moolitayer can you take a look?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -0,0 +1,127 @@
...
+    _log.info("Started watch thread")
+
+    tid
+  end
+
+  def watch_thread
+    pod_watch_stream.each do |notice|
+      break if exit_requested
+
+      _log.info("EMS [#{ems.id}] Received change for pod [#{parse_notice_pod_ems_ref(notice.object)}]")
```
moolitayer: maybe this should be debug?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    tid = Thread.new do
+      begin
+        watch_thread
+      rescue => err
+        _log.warn("Watch thread exception: #{err}")
+      end
+    end
+
+    _log.info("Started watch thread")
```
moolitayer: log thread id?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
agrare: We discussed this and @cben wanted this to be debug.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
agrare: :+1: done

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
cben: IIRC someone else suggested debug and I encouraged this to stay info...
My thought was that this is very fresh functionality and logging by default will come in handy.
LGTM from me either ways.

agrare:
@moolitayer all dependent PRs merged, PTAL

Issue #126, opened 2017-10-01
tag mapping with graph refresh


cben:
Label->Tag mapping currently implemented only with save_inventory refresh.
This is last blocker for feature parity of graph refresh with old refresh (#27, https://bugzilla.redhat.com/show_bug.cgi?id=1470021).

### merged

- [x] https://github.com/ManageIQ/manageiq/pull/16098 (merged, ignore github showing it closed)
  - refactoring, extracted Mapper class
  - Core: collect all mapped tags, batch API to create/resolve them all
    (still inefficient unbatched under the hood)
  - save_inventory_containers: use batch create/resolve
  - save_inventory_cloud: use batch create/resolve
- [x] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/130 (merged, ignore github showing it closed)
   Containers refresh parser: use new Mapper class, pass it to save_inventory_containers
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/316
   Amazon refresh parser: use new Mapper class, pass it to save_inventory_cloud

- [x] https://github.com/ManageIQ/manageiq/pull/16259 Core: expose mapped tags as InventoryObjects, create tags via InventoryCollection (even for old refresh!)

- [x] https://github.com/ManageIQ/manageiq/pull/16453 Test improvements 
- [x] ~~https://github.com/ManageIQ/manageiq/pull/16449 reformulate `ContainerLabelTagMapping.controls_tag?` as a scope~~ REVERTED
- [x] https://github.com/ManageIQ/manageiq/pull/16499 Fix idiotic mistake I made in save_tags_inventory spec that wasted me several days 
- [x] https://github.com/ManageIQ/manageiq/pull/16501 reformulate `ContainerLabelTagMapping.controls_tag?` as a scope, Take 2
- [x] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/170 Test improvements
- [x] https://github.com/ManageIQ/manageiq-ui-classic/pull/2788 change UI spec to use new factory
- [x] #162 kubernetes: Use the InventoryObjects/Collections to implement mapping in graph refresh
- [x] https://github.com/ManageIQ/manageiq-providers-openshift/pull/64 openshift: Use the InventoryObjects/Collections to implement routes, builds mapping in graph refresh, test improvements.
[...]

PR #130, opened 2017-10-15
Use new ContainerLabelTagMapping::Mapper interface

cben:
Overview/plan issue: ManageIQ/manageiq-providers-kubernetes#126

To be merged **simultaneously** with https://github.com/ManageIQ/manageiq/pull/16098 & https://github.com/ManageIQ/manageiq-providers-amazon/pull/316.  Can't pass Travis until the core PR lands.

Expose mapper to save_inventory_container which will use it to find/create all referenced tags.


cben:
Oops, still rewriting this a bit to fix a test failure.

djberg96:
@cben Any reason we can't backport this to Fine and Gaprindashvili?

PR #132, opened 2017-10-17
Improve Prometheus metrics collection

yaacov:
**Description**

A. We want to update our collector to use _rate endpoints like we use in Hawkular after #159. 

~B. We want to leverage the new Prometheus API gem, for metrics collection.~

BZs:
https://bugzilla.redhat.com/show_bug.cgi?id=1461969 - Improve Metrics collection from Prometheus
~https://bugzilla.redhat.com/show_bug.cgi?id=1468251 - Use Prometheus Metrics Ruby client~

[EDIT] BZ # 1468251 will be handled in a separate PR


yaacov:
@miq-bot add_label bug, metrics

manageiq-providers-kubernetes.gemspec
```diff
@@ -13,12 +13,12 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "hawkular-client", "~> 4.1"
-  s.add_runtime_dependency "image-inspector-client",  "~>1.0.3"
-  s.add_runtime_dependency "kubeclient",      "~>2.4.0"
-  s.add_runtime_dependency "prometheus-alert-buffer-client", "~>0.2.0"
```
moolitayer: where did this dependency go?

moolitayer:
@yaccov can you please split the new client from the c-advisor change? that would make review easier

manageiq-providers-kubernetes.gemspec
yaacov: Where did it come from ? :-) I will add it :-)

yaacov:
> @yaacov can you please split the new client from the c-advisor change? that would make review easier

sure, but it's only 3 lines, with very clear comment ?
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/132/files#diff-31fcb751e61ede5300676fc6f43811ccR9

yaacov:
> @yaacov can you please split the new client from the c-advisor change? that would make review easier

@moolitayer:

Split this PR to two PR's
a. refactor the prometheus collector to use the _rate endpoints. ( This PR )
b. move to use the prometheus-api-client instead of faraday. ( A new PR after this one is merged )

manageiq-providers-kubernetes.gemspec
yaacov: The move to prometheus-api-client is moved to a new PR, so no change here.

yaacov:

Split into two PR's for @moolitayer request, I will add the usage of the `prometheus-api-client` after this is merged in a new PR.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -6,6 +6,9 @@ def initialize(target, start_time, end_time, interval)
       @ends = end_time.to_i.in_milliseconds if end_time
       @interval = interval.to_i
       @tenant = target.try(:container_project).try(:name) || '_system'
+      # on k8s 1.7 and above @prometheus_job is "kubernetes-cadvisor"
```
cben: is this a fact about upstream k8s, or about how we config prometheus in openshift 3.7?

yaacov: > about how we config prometheus in openshift 3.7

@cben: it's how we config it,  @zgalor is maintaining the `.yaml` code for our prometheus.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -3,69 +3,72 @@ class PrometheusCaptureContext
...
     include CaptureContextMixin
 
-    def collect_node_metrics
-      # TODO: This function should be replaced to use utilization and rate endoints
-
-      # prometheus field is in sec, multiply by 1e9, sec to ns
-      # FIXME: we must update this labels to 3.7 labeling scheme and make sure it's uniqe (using type, id, and namespace labeiing)
-      cpu_resid = "sum(container_cpu_usage_seconds_total{container_name=\"\",id=\"/\",instance=\"#{@target.name}\",job=\"kubernetes-cadvisor\"}) * 1e9"
-      process_cpu_counters_rate(fetch_counters_rate(cpu_resid))
+    # avg span of 1/sec sampels
```
cben: s/sampels/samples/g

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -38,92 +41,12 @@ def ts_values
...
-        # we use:
-        # (T = start of window timestamp, V = min value of window samples)
-        # because the min value is the value of the sample closest to start of window.
-        {
-          'start' => prv['start'],
-          'end'   => n['start'],
-          'min'   => n['min'] - prv['min']
-        }
-      end
-    end
```
moolitayer: :+1: I see these were all moved as is to `hawkular_legacy_capture_context.rb`

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -38,92 +41,12 @@ def ts_values
...
 
-    private
-
-    CPU_NANOSECONDS = 1e09
-
-    def target_name
-      "#{@target.class.name.demodulize}(#{@target.id})"
-    end
-
     def validate_target
```
moolitayer: we want to keep this one private

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -6,6 +6,9 @@ def initialize(target, start_time, end_time, interval)
       @ends = end_time.to_i.in_milliseconds if end_time
       @interval = interval.to_i
       @tenant = target.try(:container_project).try(:name) || '_system'
+      # on k8s 1.7 and above @prometheus_job is "kubernetes-cadvisor"
+      # on older deploys @prometheus_job is "kubernetes-nodes"
```
moolitayer: And we don't support OpenShift < 3.7 for Prometheus right?

zgalor: right

cben:
Code LGTM.  Functionality — @yaacov is looking into how `rate(...[5m])` may affect reports, I trust him to decide what to do.

One corner case that may help decide: imagine original sample intervals are very non-uniform, some missing to you have short & long intervals.  I'm pretty sure `irate` (last 2 samples dy/dt) would be even worse then, changes would be divided by different intervals and then summing/averaging *slopes* with equal weight would be wrong.
IIUC, the problem is inherent in schema that deals in *usage per second* samples.  We must store either cumulative totals or `increase` over segment, not divided by time.  Increases can be added/averaged safely, they are inherently weighted proportional to time covered.

Usage samples *can* be approximately OK iff samples really have fixed frequency, like 60s.  But for that, when we get gaps we *can't set missing points as neither nil nor 0*!  We'd have to carefully spread the single rate/second over the gap :hole:.  This is horrendous, let's please not go there :pray:.

TLDR: :chart_with_upwards_trend: :bar_chart: !

What now?

- What did previous ruby derivative code do if given missing points?
  I now suspect recent "fixes" we did to divide by correct *dt* actually made things worse :-)
  Assuming that suffered from similar deep problem, let's merge this and open separate bigger wrong-schema issue...
- rate[5m] will smooth occasional missing samples.  It's not the *right* approximation, but it's probably an improvement at this stage :-)

yaacov:
> IIUC, the problem is inherent in schema that deals in usage per second samples. We must store either cumulative totals or increase over segment, not divided by time. Increases can be added/averaged safely, they are inherently weighted proportional to time covered.

@cben I agree, we need to store the usage raw data (or `increase`) + raw data of the `node_capacity` and do cosumption calculations as far down the chain as we can.

yaacov:
@cben @moolitayer pleese ping if we can squash this.
not squashing to make it easy to see change done to fix the "POD" pod issue.

yaacov:
@moolitayer ping me when you want to squash

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -3,69 +3,74 @@ class PrometheusCaptureContext
...
-    def collect_node_metrics
-      # TODO: This function should be replaced to use utilization and rate endoints
-
-      # prometheus field is in sec, multiply by 1e9, sec to ns
-      # FIXME: we must update this labels to 3.7 labeling scheme and make sure it's uniqe (using type, id, and namespace labeiing)
-      cpu_resid = "sum(container_cpu_usage_seconds_total{container_name=\"\",id=\"/\",instance=\"#{@target.name}\",job=\"kubernetes-cadvisor\"}) * 1e9"
-      process_cpu_counters_rate(fetch_counters_rate(cpu_resid))
+    # avg span of 1/sec samples
+    # value depend on prometheus scraping rate, we scrape every 30s
+    # min avg span of 1/sec samples for this scrape rate is about ~5m
```
moolitayer: I don't understand this comment and the 5m value is the last value you had here.
Can you rephrase?
You explained yesterday this is the size of the sliding window we use for the rate function.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -3,69 +3,74 @@ class PrometheusCaptureContext
...
+      @metrics = %w(cpu_usage_rate_average mem_usage_absolute_average)
+      collect_metrics_for_labels(labels)
     end
 
     def collect_group_metrics
-      # TODO: This function should be replaced to use utilization and rate endoints
+      # set pod labels
+      # NOTE: we do not use container_name = "POD"
+      #       "POD" is a system container attached to each pod that
+      #       holds to the namespace so it will not get recycled
```
moolitayer: Could you write this positive rather than negative, example:
```pod_name="X" willl yield metrics for all the containers belonging to X as well as the internal POD container (OpenShift's equivalent of kubernetes  'pause' )```

**Comment by moolitayer** [comment]
@yaacov please squash

File: app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb (hunk 8 above)
Comment (yaacov) [reply app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb]: 👍 nice

**Comment by yaacov** [comment]
> @yaacov please squash

@moolitayer squashed  :paw_prints:

**Comment by moolitayer** [comment]
Labeling as Gaprindashvili/yes per BZ 1461969

**Comment by simaishi** [comment]

```
$ git log -1
commit d226f8656912f49a143e9a034da5826a2482b3cc
Date:   Wed Dec 27 12:58:48 2017 +0200

    Merge pull request #132 from yaacov/use-cadvisor-job-for-prometheus-metrics
    
    Improve Prometheus metrics collection
    (cherry picked from commit 63c6681fc93388f4c66fefb1e55cca6bbe2fff03)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1530734
```

PR #134, opened 2017-10-18
Add parent inventory collections to ensure proper disconnect order

cben:
Proposed replacement for #106 — tweaked the initialize_custom_attributes_collections interface per review discussion.

Add parent inventory collections to ensure proper disconnect order in the case that collection is empty and we can't compute dependencies.

Fixes bug with `save_unused_images: false` mode deleting existing child custom_attributes records (https://github.com/ManageIQ/manageiq-providers-openshift/pull/46, which also needs a tiny tweak to work with this)

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -28,7 +28,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
         :delete_method  => :disconnect_inv,
       )
     )
-    initialize_custom_attributes_collections(manager.container_projects, %w(labels additional_attributes))
+    initialize_custom_attributes_collections(@collections[:container_projects], %w(labels additional_attributes))
 
     @collections[:container_quotas] = ::ManagerRefresh::InventoryCollection.new(
       shared_options.merge(
@@ -73,8 +73,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
         :secondary_refs => {:by_name => [:name]},
       )
     )
-    initialize_container_conditions_collection(manager.container_nodes)
-    initialize_custom_attributes_collections(manager.container_nodes, %w(labels additional_attributes))
+    initialize_container_conditions_collection(manager, :container_nodes)
+    initialize_custom_attributes_collections(@collections[:container_nodes], %w(labels additional_attributes))
 
     # polymorphic child of ContainerNode & ContainerImage,
     # but refresh only sets it on nodes.
@@ -143,8 +143,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :delete_method        => :disconnect_inv,
         )
       )
-    initialize_container_conditions_collection(manager.container_groups)
-    initialize_custom_attributes_collections(manager.container_groups, %w(labels node_selectors))
+    initialize_container_conditions_collection(manager, :container_groups)
+    initialize_custom_attributes_collections(@collections[:container_groups], %w(labels node_selectors))
     @collections[:container_volumes]      =
       ::ManagerRefresh::InventoryCollection.new(
         shared_options.merge(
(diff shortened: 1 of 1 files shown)


cben:
@Ladas please review (we can continue here or on your PR, whatever you prefer)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -326,7 +327,8 @@ def initialize_container_conditions_collection(relation)
   end
 
   # CustomAttribute is polymorphic child of many models
-  def initialize_custom_attributes_collections(relation, sections)
+  def initialize_custom_attributes_collections(parent_collection, sections)
+    relation = parent_collection.full_collection_for_comparison
```
Ladas: hm, we probably do not need the relation here? 

```relation.model.base_class.name```
can be
```parent_collection.model.class.base_class.name```

right?

Comment (cben) [reply app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb]: We need something to build query on, see `:resource_id   => relation` below.
In your version this was `manager.public_send(association)` but I figured if I have the actual collection can use its actual query.

Oh, yes, resource_type can be simplified :+1:

Comment (cben) [reply app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb]: Done, PTAL.
Can you update https://github.com/ManageIQ/manageiq-providers-openshift/pull/46 to
```
      initialize_custom_attributes_collections(@collections[:container_images], %w(labels docker_labels))
```

BTW, should we do same for initialize_container_conditions_collection?  Set parent_collection?
I'm inclined to refactor it to same interface anyway.

File: app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb [hunk 2]
```diff
@@ -326,19 +327,22 @@ def initialize_container_conditions_collection(relation)
   end
 
   # CustomAttribute is polymorphic child of many models
-  def initialize_custom_attributes_collections(relation, sections)
+  def initialize_custom_attributes_collections(parent_collection, sections)
+    type = parent_collection.model_class.base_class.name
+    relation = parent_collection.full_collection_for_comparison
```
cben: Also I wasn't sure if I should use `full_collection_for_comparison` or `db_relation` or something, went with the one that looked simplest ;-)

Ladas: Right, those are picked based on a set strategy

Ladas:
@cben ah, fun CI fail, it's because we create OpenShift manager in k8s https://github.com/Ladas/manageiq-providers-kubernetes/blob/a98199b2bee2aa079d177e802067bf12d6770ce0/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context_spec.rb#L10

so initialize_custom_attributes_collections, because 1 IC is still passing relation there. So the fis should be to use kubernetes ems factory

cben:
Yes, using k8s factory there is tricky (https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/34).  We gave up for now, and the price is occasional circular dep when *neither* PR can pass Travis until both are merged :-(

If you can update https://github.com/ManageIQ/manageiq-providers-openshift/pull/46 to new signature, we'll test both locally we'll be able to merge.
You can just fetch and push from my branch https://github.com/cben/manageiq-providers-openshift/commits/enable_check_metadata_not_deleted_specs

cben:
Hmm, #34 was supposed to be fixed, this is new dep on openshift, checking if I can easily fix it...

simon3z:
> If you can update ManageIQ/manageiq-providers-openshift#46 to new signature, we'll test both locally we'll be able to merge.

@cben @Ladas let me know when you tested. AFAIU I should merge both this PR and https://github.com/ManageIQ/manageiq-providers-openshift/pull/46 once verified. Right?

Ladas:
@cben @simon3z when testing locally with https://github.com/ManageIQ/manageiq-providers-openshift/pull/46 the specs are passingh

PR #138, opened 2017-10-24
Update metrics to cAdvisor endpoint, and re-record metrics capture tests using k8s v1.7

yaacov:
**Description**

We currently use Openshift provider for the tests in the Kubernetes provider, this PR fixes that.

a. Use k8s for metrics capture tests.
b. Re-record Prometheus metrics from a k8s provider.
c. Use the cAdvisor endpoint for metrics collection.


yaacov:
@cben @moolitayer  please review

Related:
https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/34
https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/72

yaacov:
cc @simon3z

yaacov:
p.s.

The change from `kubernetes-nodes` to `kubernetes-cadvisor` is just because my provider is 1.7 ...
We are working on a bigger refactor here: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/132

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context_spec.rb
```diff
@@ -31,17 +32,17 @@
         EmsRefresh.refresh(@ems)
         @ems.reload
 
-        @node = @ems.container_nodes.find_by(:name => "capture.context.com")
-        pod = @ems.container_groups.find_by(:name => "docker-registry-1-4svbw")
+        @node = @ems.container_nodes.find_by(:name => "mas-1.example.com")
+        pod = @ems.container_groups.where("name like ?", "docker-registry%").last
```
cben: There is exactly one such pod (I just run this to check).
So `where(...).last` == `where(...).first` == `find_by(...)`.  I recommend using `find_by` as idiomatic for "I expect there will be exactly one.

cben:
LGTM

moolitayer:
> c. Use the cAdvisor endpoint for the new recording.

Not only for the new recording, for metrics collection...

yaacov:
> Not only for the new recording, for metrics collection...

@moolitayer 👍 see updated description

moolitayer:
@yaacov can you please change the PR title (tests change seems like the minor change here)?

@yaacov @cben Sorry for reiterating, just to make 100% sure;
After this change we will:
- get metrics support for the new Prometheus job configuration (cc @zgalor)
- loose metrics support for the old conf
And that is what we want for metrics?

(should we think about warning users if they try to add an unsupported Prometheus?)

yaacov:
> @yaacov can you please change the PR title (tests change seems like the minor change here)?

done 👍 

>@yaacov @cben Sorry for reiterating, just to make 100% sure;
>After this change we will:
>
>    get metrics support for the new Prometheus job configuration (cc @zgalor)
    loose metrics support for the old conf
    And that is what we want for metrics?

@moolitayer 

Yes and No :-)

This will get metrics from Prometheus installed in the k8s > 1.7 configuration.
This will not work with k8s <= 1.6, but we did not officially support it before [1].
This is what we decided with @simon3z : only support k8s >= 1.7

> (should we think about warning users if they try to add an unsupported Prometheus?)

No, we did not support Prometheus before this PR [1] and we *do not* support it now.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1461969 - on_dev

moolitayer:
> (should we think about warning users if they try to add an unsupported Prometheus?)
> No, we did not support Prometheus before this PR [1] and we do not support it now.

If the metrics endpoint is supported only on from a specific k8s/ocp version it would be helpful to give the user feedback when adding the endpoint

PR #140, opened 2017-10-26
Collect ems targeted Prometheus events

moolitayer:
- Enable alerts on the new events
- Dropped the timestamp hack (no longer needed after https://github.com/ManageIQ/manageiq/pull/16104)

Part of Prometheus Alert integration:
https://github.com/ManageIQ/manageiq/issues/14238

Related to
https://github.com/ManageIQ/manageiq/pull/16310

### DIFF
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
@@ -79,29 +79,43 @@ module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerM
     event[:url] = annotations["url"]
     event[:severity] = parse_severity(annotations["severity"])
     labels = event["labels"]
+    # TODO(mtayer): remove after https://github.com/ManageIQ/manageiq/pull/16339
     event[:ems_ref] = incident_identifier(event, labels, annotations)
     event[:resolved] = event["status"] == "resolved"
-    timestamp = event["timestamp"]
-
-    target = find_target(labels)
+    timestamp = event[:resolved] ? event["endsAt"] : event["startsAt"]
     {
-      :ems_id              => @cfg[:ems_id],
-      :source              => "DATAWAREHOUSE",
-      :timestamp           => timestamp,
-      :event_type          => "datawarehouse_alert",
-      :target_type         => target.class.name,
-      :target_id           => target.id,
-      :container_node_id   => target.id,
-      :container_node_name => target.name,
-      :message             => annotations["message"],
-      :full_data           => event.to_h
-    }
+      :source     => "DATAWAREHOUSE",
+      :timestamp  => timestamp,
+      :event_type => "datawarehouse_alert",
+      :message    => annotations["message"],
+      :ems_ref    => incident_identifier(event, labels, annotations),
+      :full_data  => event.to_h
+    }.merge(
+      find_target(
+        annotations,
+        labels
+      )
+    )
   end
 
-  def find_target(labels)
-    instance = ContainerNode.find_by(:name => labels["instance"], :ems_id => @target_ems_id)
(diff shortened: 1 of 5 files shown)


moolitayer:
@simon3z @cben @enoodle please review

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -14,7 +14,7 @@ def reset_event_monitor_handle
   end
 
   def stop_event_monitor
-    @event_monitor_handle.stop unless @event_monitor_handle.nil?
+    @event_monitor_handle.try(:stop)
```
cben: I mildly prefer original, or `@event_monitor_handle&.stop`.
`.try` also tolerates any non-nil object that does not respond to `.stop`.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -58,18 +62,21 @@ def fetch
...
-    timestamp = alert["status"] == 'resolved' ? alert["endsAt"] : alert["startsAt"]
-    Time.zone.at((Time.parse(timestamp).to_f + (0.000001 * indent)))
+  def process_alert!(alert, generation, group_index)
+    alert['generationID'] = generation
+    alert['index'] = group_index
+    alert
+  end
+
+  def alert_for_miq?(alert)
+    %w(ContainerNode ExtManagementSystem).include?(
```
cben: duplicates cases in `find_target`.  is it possible to always attempt `find_target` and skip if that fails?

cben:
but note Travis failure.

enoodle:
@moolitayer seems like travis is failing due to changed interface of `find_target`. Other than that LGTM

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
moolitayer: I could do something like 
```
if annotations['miqTarget'] == 'ContainerNode'
  # handle node
else
  # handle ext
end
```
But i'm not sure that is better. WDYT @cben?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
moolitayer: That was actually do to a review comment[1]. It should be ok now to use safe navigation right?
maybe we can use that. Reverted this change until we will settle this.

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/40#discussion_r144874951

moolitayer:
@cben @enoodle could you please give this another look?
I believe Travis will be happy now

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -79,29 +79,42 @@ def extract_event_data(event)
...
+      # TODO: we must do the db query here unless we get Prometheus to emit ems_ref
+      node = ContainerNode.find_by(:ems_id => @target_ems_id, :name => labels["instance"])
+      $cn_monitoring_log.error("Could not find node from labels: [#{labels}]") unless node
+      {
+        :container_node_name => labels["instance"],
+        :container_node_id   => node.try(:id),
+        :target_type         => node.try(:class).try(:name),
+        :target_id           => node.try(:id),
+      }
+    when 'ExtManagementSystem' then {
```
cben: nit: the brace after `then` opens a hash, not a block right?
in this position it really feels like a block visually.  please put on next line without `then`.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
cben: never mind then, what you have LGTM.

PR #135, opened 2017-10-18
Targeted refresh for pods, using watches

Ladas:
Has to be merged together with 
https://github.com/ManageIQ/manageiq-providers-openshift/pull/54 to pass the CI (cyclical dependency of the prometheus spec)

Targeted refresh for pods, using watches

Issue #85, opened 2017-08-08
Relationship to all entities that are archivable should use an active context


moolitayer:
I think these are:
- container.rb
- container_group.rb
- container_image.rb
- container_project.rb

See 
- https://github.com/ManageIQ/manageiq/pull/15251/files#diff-055b42a9cb7c02e4d91fb07e80ca49fdR16
- https://github.com/ManageIQ/manageiq/pull/15251/files#diff-268a58d46a171293f40496890d650b72R5

cc @Ladas @cben @agrare

cben:
and container_node.rb.

cc @zakiva @zeari 
I agree we want `active` on the associations by default.  But should be careful whether this changes meaning of any historical/chargeback reporting — might need to adjust those queries to still include archived objects.

zeari:
> meaning of any historical/chargeback reporting — might need to adjust those queries to still include archived objects.

@cben we use `archived?` so it should be fine. https://github.com/manageiq/manageiq/blob/master/app/models/chargeback_container_image.rb#L100

Ladas:
Also the relations using :active should be the same as before, when the foreign key was nullified. In both cases, the relation did not return disconnected objects. So chargeback was usually using something else, like Model.where(...), instead of the associations.

cben:
We're talking about entity-entity association here, not ems-entity.  Those
are not and were not nullified on disconnect (except image-registry), so
AFAICT return both active and archived
(example that started this: image.containers. container may be archived
while image is still in use.)

Chargeback: just asking archived? is not enough, if we add active scoping,
youll stop seeing archived ones; will need the both-kinds query to preserve
behavior.

zeari:
> Chargeback: just asking archived? is not enough, if we add active scoping,
> youll stop seeing archived ones; will need the both-kinds query to preserve
> behavior.

We always take all Containers\projects into consideration regardless if theyre currently active or not.
`archived?` is used to show in the specific line of the report if the container\project is currently alive or not.

cben:
leftover: the pod-project relation is disconnected but kept track of using `old_container_project`:
`ContainerGroup belongs_to :old_container_project`
`ContainerProject has_many :archived_container_groups, :foreign_key => "old_container_project_id"`

In the spirit of soft delete, should stop doing that, retain container group's `container_project_id` intact but put `active` scope on `project.container_groups`.

mfeifer:
Please re-open if this is still a bug.

PR #142, opened 2017-10-26
Fix kubernetes refresher spec tests

agrare:
There was a test added which assumed we do not nil out the container_project from a pod, but this is not committed yet.
This reverts the commit that changed the test expectations.

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
@@ -519,7 +519,7 @@ shared_examples "kubernetes refresher VCR tests" do |check_tag_mapping: true|
 
         [pod0, pod1].each do |pod|
           assert_disconnected(pod)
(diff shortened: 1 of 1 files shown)


agrare:
Ref: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/135#issuecomment-339736987

agrare:
cc @moolitayer @cben

cben:
Thanks, LGTM.  This reverts commit 3685aa60c4ddcb1e6f312714574aead2db2e821c.

Is there a PR that's going to stop clearing `archived_pod.container_project_id`?
Should link that as addressing part of #85.  And per [last comment there](https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/85#issuecomment-328715130), should scope `project.container_groups` relations to `active`.

agrare:
@cben yes https://github.com/ManageIQ/manageiq/pull/16302

moolitayer:
Thanks for the fix @agrare

PR #144, opened 2017-10-30
Fix spec since we do not delete project foreign_key

Ladas:
Fix spec since we do not delete project foreign_key when
we disconnect pod.

Needed for fixing CI after the https://github.com/ManageIQ/manageiq/pull/16302 is merged


agrare:
Kicking tests

agrare:
> Hmm, not sure above sentence adds the clarity I hoped for ;-)

Hahaha

agrare:
Hm travis seems to be stuck

Issue #103, opened 2017-08-28
Should re-connect container images that are used again


cben:
https://bugzilla.redhat.com/show_bug.cgi?id=1488072

(extracting from https://github.com/ManageIQ/manageiq/pull/14808 which was an unacceptable solution, so we can track the bug)
If refresh archives a container image record and later refresh encounters same image again, we create a duplicate record.  This can be confusing in reports.

We don't see this much when images come from openshift registry, as these stay around for a long time.
Scenarios where we may disconnect an image not currently used in any container:

- a Kubernetes provider
- Openshift provider that run images from external registries
- Openshift dropping the image from /oapi/v1/images e.g. by [pruning](https://docs.openshift.org/latest/admin_guide/pruning_resources.html#pruning-images)
- Openshift configured get_container_images: false (https://github.com/ManageIQ/manageiq/pull/14606)
- Openshift configured store_unused_images: false (#11 + https://github.com/ManageIQ/manageiq-providers-openshift/pull/9)
- Openshift fetching used images one by one (if https://github.com/ManageIQ/manageiq/pull/14628 gets revived)

simon3z:
@cben we probably need better tracking of this RFE, if you file a BZ it wont go unnoticed (please **do** file a BZ :smile: )

cben:
Should be fixed for graph refresh by #145.

Left unfixed in save_inventory refresh as we never found an acceptably efficient fix.  (This is our first deliberate behavior divergence between old vs graph refresh.)

PR #145, opened 2017-10-31
Make sure targeted refresh does not duplicate entities

Ladas:
Make sure targeted refresh does not duplicate entities by implementing the reconnect logic for ContainerGroup, Container and ContainerImage models.

Fixes BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1507899

Fixes issue:
https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/103

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -127,7 +127,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           # TODO: should match on digest when available
           :manager_ref            => [:image_ref],
           :delete_method          => :disconnect_inv,
-          :custom_reconnect_block => custom_reconnect_block(:image_ref)
+          :custom_reconnect_block => custom_reconnect_block
         )
       )
     # images have custom_attributes but that's done conditionally in openshift parser
@@ -355,16 +355,17 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
     end
   end
 
-  def custom_reconnect_block(manager_ref = :ems_ref)
+  def custom_reconnect_block
     # TODO(lsmola) once we have DB unique indexes, we can stop using manual reconnect, since it adds processing time
     lambda do |inventory_collection, inventory_objects_index, attributes_index|
       relation = inventory_collection.model_class.where(:ems_id => inventory_collection.parent.id).archived
 
       # Skip reconnect if there are no archived entities
       return if relation.archived.count <= 0
+      raise "Allowed only manager_ref size of 1, got #{inventory_collection.manager_ref}" if inventory_collection.manager_ref.count > 1
 
       inventory_objects_index.each_slice(1000) do |batch|
-        relation.where(manager_ref => batch.map(&:second).map(&:manager_uuid)).each do |record|
+        relation.where(inventory_collection.manager_ref.first => batch.map(&:second).map(&:manager_uuid)).each do |record|
           index = inventory_collection.object_index_with_keys(inventory_collection.manager_ref_to_cols, record)
 
           # We need to delete the record from the inventory_objects_index and attributes_index, otherwise it
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -116,17 +116,18 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :manager_ref    => [:host, :port],
         )
       )
-    @collections[:container_images]           =
+    @collections[:container_images] =
       ::ManagerRefresh::InventoryCollection.new(
         shared_options.merge(
-          :model_class    => ContainerImage,
-          :parent         => manager,
-          :builder_params => {:ems_id => manager.id},
-          :association    => :container_images,
+          :model_class            => ContainerImage,
+          :parent                 => manager,
+          :builder_params         => {:ems_id => manager.id},
+          :association            => :container_images,
           # TODO: old save matches on [:image_ref, :container_image_registry_id]
           # TODO: should match on digest when available
-          :manager_ref    => [:image_ref],
-          :delete_method  => :disconnect_inv,
+          :manager_ref            => [:image_ref],
+          :delete_method          => :disconnect_inv,
+          :custom_reconnect_block => custom_reconnect_block(:image_ref)
         )
       )
     # images have custom_attributes but that's done conditionally in openshift parser
@@ -134,13 +135,14 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
     @collections[:container_groups] =
       ::ManagerRefresh::InventoryCollection.new(
         shared_options.merge(
-          :model_class          => ContainerGroup,
-          :parent               => manager,
-          :builder_params       => {:ems_id => manager.id},
-          :association          => :container_groups,
-          :secondary_refs       => {:by_namespace_and_name => [:namespace, :name]},
-          :attributes_blacklist => [:namespace],
-          :delete_method        => :disconnect_inv,
+          :model_class            => ContainerGroup,
+          :parent                 => manager,
+          :builder_params         => {:ems_id => manager.id},
+          :association            => :container_groups,
+          :secondary_refs         => {:by_namespace_and_name => [:namespace, :name]},
+          :attributes_blacklist   => [:namespace],
+          :delete_method          => :disconnect_inv,
+          :custom_reconnect_block => custom_reconnect_block
         )
       )
     initialize_container_conditions_collection(manager, :container_groups)
@@ -157,12 +159,13 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
     @collections[:containers]             =
       ::ManagerRefresh::InventoryCollection.new(
         shared_options.merge(
-          :model_class    => Container,
-          :parent         => manager,
-          :builder_params => {:ems_id => manager.id},
-          :association    => :containers,
+          :model_class            => Container,
+          :parent                 => manager,
+          :builder_params         => {:ems_id => manager.id},
+          :association            => :containers,
           # parser sets :ems_ref => "#{pod_id}_#{container.name}_#{container.image}"
-          :delete_method  => :disconnect_inv,
+          :delete_method          => :disconnect_inv,
+          :custom_reconnect_block => custom_reconnect_block
         )
       )
     @collections[:container_port_configs] =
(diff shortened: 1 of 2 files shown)


Ladas:
cc @agrare @simon3z @moolitayer

cben:
Cool!
The BZ it solves is about watch worker but I think this also does reconnections in full refresh, right?  Sounds like it solves #103?

Does this depart from save_inventory<->graph refresh equivalence?
(BTW, do we care about watch worker + save_inventory, or do we consider that an unsupported combination of settings?)

Do you have any numbers on how much this slows down things?

About the BZ scenario this addresses:
IIUC it's a race condition, where pod watch notifications still arrive after full refresh GET /pods already saw it deleted.  Did I get that right?
I wonder if we should discard the watch data then, because it's stale?
OK, in this case while it's stale in sense the pod was deleted, it's fresher data on *pod's state*, so should not be discarded.
I'm just trying to build some mental model how to think about this kind of issues...

Ladas:
@cben yes this solves https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/103

```Does this depart from save_inventory<->graph refresh equivalence?
(BTW, do we care about watch worker + save_inventory, or do we consider that an unsupported combination of settings?)``` I think yes. If used with old refresh, we open to possibility of duplicates. With the fact that duplicates should be possible only for ContainerImage, in old refresh without the collector worker.

```about the perf impact```, it should be low, since majority of creations should be done by small batches in targeted refresh. But I haven't do exact measurements.

```About the BZ scenario this addresses:```
Yes, you get it right, if you delete the pod after full refresh starts, but the API query will not fetch it, then we disconnect it, while accumulating watch data about the pod in the queue. 

We do not store the resource version, so we cannot really ignore it. For now refreshing the watch events accumulated in queue should get us to the latest version. I am just thinking how to effectively disconnect by update, could be done by detecting the finished_at date in event, we can just set deleted_on date.

For 5.0, the resource version (or timestamp) will be used for atomic checks in DB, to avoid updating new data by the old.

**Comment by moolitayer** [comment]
- There seems to be some additional logic in disconnection we are not addressing here? for example in `container_image.disconnect_inv | container_group.disconnect_inv`

- in the scenario described in bz 1507899, it seems the object won't be duplicated but it will be marked as a live object instead of a deleted one?

**Comment by Ladas** [comment]
@moolitayer the relations deleted by disconnect_inv methods should be put back by refresh itself. But it's for other PRs, no remove the nullifying of the foreign_keys.

Yes, for now we will mark it as live one. The next full refresh should disconnect it again (if we were hitting the race). I will look into optimizing this, using the finished_at dates from watch data (that we have for Pod and Container), we could decide to just update the archived entity (so it has the latest resource version data) but not making it active.

**Comment by cben** [comment]
:tada: We do store it (parsed by `parse_base_item`):
```
> ContainerGroup.last.resource_version
"2174456"
```
does this help?

**Comment by cben** [comment]
> > Does this depart from save_inventory<->graph refresh equivalence? 
> > (BTW, do we care about watch worker + save_inventory, or do we consider that an unsupported combination of settings?) 

> I think yes. 

just to be clear, did you mean yes it departs / yes we care / yes unsupported combination ?

> If used with old refresh, we open to possibility of duplicates. With the fact that duplicates should be possible only for ContainerImage, in old refresh without the collector worker.

**Comment by Ladas** [comment]
@cben maybe :-) TBH I haven't thought about resource version much for 4.6 scope. We can check it out later as an optimization. Anyway for the race, if the full refresh did the disconnect, then there has to be already queued targets with resource version >= than the actual version. 

There are like 3 events generated for pod delete and I think the full refresh can't give us the last (containing the finished_at date), so maybe resource version doesn't help here at all.

For 5.0, the resource version is essential, since it will all be in parallel.

**Comment by Ladas** [comment]
@cben ```just to be clear, did you mean yes it departs / yes we care / yes unsupported combination ?```

em, yes it departs, yes unsupported combination (given we do not like duplicates). Not sure if we care. :-) The switch to old refresh should be in a very critical case given now we would lose features too.

**Comment by Ladas** [comment]
@cben hm, maybe we can set deleted_on = nil only for ContainerImage, we are talking with @agrare that we might fetch only ```type [DELETED]``` events for g-release, since everything else will be fetched by a full refresh.

**Comment by agrare** [comment]
@Ladas what about using non-graph refresh would lead to duplicates here?  It is still running in only the single context

**Comment by moolitayer** [comment]
@Ladas we need projects too right?
What about Nodes? while they are not part of the watch it seems like we would benefit from being able to reconnect them (removing a node and then adding it again seems like a common use case)

**Comment by Ladas** [comment]
@moolitayer @cben @agrare https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/151 disabling possible combination of old refresh + targeted refresh, cause I think that would start to delete most of the inventory. 

@moolitayer we can add projects and nodes in this PR if you want, I was focusing solely on the BZ described race condition. Is there a case when remove&add of Project or Node can result into the same ems_ref? (cc @cben)

File: app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb [hunk 1]
```diff
@@ -351,4 +354,36 @@ def initialize_custom_attributes_collections(parent_collection, sections)
         )
     end
   end
+
+  def custom_reconnect_block(manager_ref = :ems_ref)
```
agrare: @Ladas this looks like a pretty generic "customer" reconnect block, is this something that could be at the `ManagerRefresh` level?

Ladas: yeah, maybe we could move it up later

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -351,4 +354,36 @@ def initialize_custom_attributes_collections(parent_collection, sections)
...
+        relation.where(manager_ref => batch.map(&:second).map(&:manager_uuid)).each do |record|
+          index = inventory_collection.object_index_with_keys(inventory_collection.manager_ref_to_cols, record)
+
+          # We need to delete the record from the inventory_objects_index and attributes_index, otherwise it
+          # would be sent for create.
+          inventory_object = inventory_objects_index.delete(index)
+          hash             = attributes_index.delete(index)
+
+          # Make the entity active again, otherwise we would be duplicating nested entities
+          hash[:deleted_on] = nil
```
cben: idea: what if you don't do this?  could you update the archived record in-place without unarchiving it?

(not blocker for this PR, but I think that's the correct merge of "an update to pod/foo" with "pod/ no longer contains foo")

BTW, remind me what we do with DELETED notices?  can they archive the pod, or is that only done by full refreshes?

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
```diff
@@ -46,6 +46,12 @@
     full_refresh_test(:targeted_refresh_all_nodes_and_namespaces)
   end
 
+  it "will perform a targeted refresh not duplicating archived Pods and Containers" do
+    create_archived_entities
+
+    full_refresh_test(:targeted_refresh_referenced_nodes_and_namespaces)
```
cben: does this actually test the problematic scenario?  don't you also need some `queue_target!`?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: right, so that would cause duplication of the nested tables (container_condition, container_volume, custom_attribute, security_context), since there is a query that goes through active container_group I think. So we would have to solve reconnect for those too...

Best to wait for unique indexes, so we don't over-complicate the refresh for reconnect that covers 1% of edges cases. :-)

Only full refresh can disconnect right now. We use the simplest version of targeted refresh, that can only create&update records, for g-release. :-)

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
Ladas: yes, it uses the same queued targets as the other specs, that is inside full_refresh_test method.

So what we do here is that ```create_archived_entities``` is mocking DB with archived entities and we check that we reconnect&update those entities instead of duplicating them.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -116,31 +116,33 @@ def initialize_inventory_collections
...
+          :parent                 => manager,
+          :builder_params         => {:ems_id => manager.id},
+          :association            => :container_images,
           # TODO: old save matches on [:image_ref, :container_image_registry_id]
           # TODO: should match on digest when available
-          :manager_ref    => [:image_ref],
-          :delete_method  => :disconnect_inv,
+          :manager_ref            => [:image_ref],
+          :delete_method          => :disconnect_inv,
+          :custom_reconnect_block => custom_reconnect_block(:image_ref)
```
moolitayer: isn't this always `inventory_collection.manager_ref` ?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
moolitayer: @Ladas I would really prefer to have this in the main repo with the other infrastructure components.
I fear I'm missing the knowledge to review this code

cc @simon3z

Ladas: for the sake of backport, it's safer to first have it here, since for g-release, we do not use it elsewhere

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: right, yeah, we can use that, rather than passing argument. Will fix it

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: Also, this is not really a generic code applicable e.g. to a cloud, cloud ticks a bit different, so I would have to modify it. And also the plan is to get rid of this ruby post processing and rather use unique indexes in the DB.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -364,7 +364,7 @@ def custom_reconnect_block(manager_ref = :ems_ref)
       return if relation.archived.count <= 0
 
       inventory_objects_index.each_slice(1000) do |batch|
-        relation.where(manager_ref => batch.map(&:second).map(&:manager_uuid)).each do |record|
+        relation.where(inventory_collection.manager_ref.first => batch.map(&:second).map(&:manager_uuid)).each do |record|
```
cben: please assert manager_ref consists of exactly one column.
(fine assumption for current needs, but if tomorrow someone reuses this block for more tables, I prefer an error to wrong reconnections)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -351,4 +354,37 @@ def initialize_custom_attributes_collections(parent_collection, sections)
...
   end
+
+  def custom_reconnect_block
+    # TODO(lsmola) once we have DB unique indexes, we can stop using manual reconnect, since it adds processing time
+    lambda do |inventory_collection, inventory_objects_index, attributes_index|
+      relation = inventory_collection.model_class.where(:ems_id => inventory_collection.parent.id).archived
+
+      # Skip reconnect if there are no archived entities
+      return if relation.archived.count <= 0
+      raise "Allowed only manager_ef size 1, got #{inventory_collection.manager_ref}" if inventory_collection.manager_ref.count > 1
```
moolitayer: s/manager_ef size 1/manager_ref of size 1/

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -351,4 +354,37 @@ def initialize_custom_attributes_collections(parent_collection, sections)
...
+  def custom_reconnect_block
+    # TODO(lsmola) once we have DB unique indexes, we can stop using manual reconnect, since it adds processing time
+    lambda do |inventory_collection, inventory_objects_index, attributes_index|
+      relation = inventory_collection.model_class.where(:ems_id => inventory_collection.parent.id).archived
+
+      # Skip reconnect if there are no archived entities
+      return if relation.archived.count <= 0
+      raise "Allowed only manager_ref size of 1, got #{inventory_collection.manager_ref}" if inventory_collection.manager_ref.count > 1
+
+      inventory_objects_index.each_slice(1000) do |batch|
```
moolitayer: Seems like it would be helpful to have configurable

simaishi:

```
$ git log -1
commit ba092974f99f4439feff5e2b567fcaaeb061b66f
Date:   Tue Nov 7 15:03:21 2017 +0200

    Merge pull request #145 from Ladas/make_sure_targeted_refresh_does_not_duplicate_entities
    
    Make sure targeted refresh does not duplicate entities
    (cherry picked from commit 61765c1e95295efd1d0c466fdf5ea2e0e736a098)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1510475
```

PR #149, opened 2017-10-31
Prometheus Alert flow changes

moolitayer:
Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1507990

Currently ManageIQ collects alerts with annotation `MiqTarget=ContainerNode|ExtManagementSystem` and uses the annotation to find the target object.
Suggested: ManageIQ collects all generated alerts[1] and always attaches incoming alerts to the ems

Note: a user can still specify `MiqTarget=ContainerNode` to attach a specific alert to a node.
- If the node cannot be found for some reason,  the alert will fallback on the ext.
- There should never be a mismatch, even if we have a Node alerts with several firing instances defined on the same node (e.g one instance for each container in the node).
  The screen would show them as duplicated alerts.

[1] you can still ignore alerts by annotating them with `miqIgnore=true`

### DIFF
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
@@ -80,7 +80,7 @@ module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerM
     event[:severity] = parse_severity(annotations["severity"])
     labels = event["labels"]
     # TODO(mtayer): remove after https://github.com/ManageIQ/manageiq/pull/16339
-    event[:ems_ref] = incident_identifier(event, labels, annotations)
+    event[:ems_ref] = incident_identifier(labels, annotations, event["startsAt"])
     event[:resolved] = event["status"] == "resolved"
     timestamp = event[:resolved] ? event["endsAt"] : event["startsAt"]
     {
@@ -88,7 +88,7 @@ module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerM
       :timestamp  => timestamp,
       :event_type => "datawarehouse_alert",
       :message    => annotations["message"],
-      :ems_ref    => incident_identifier(event, labels, annotations),
+      :ems_ref    => incident_identifier(labels, annotations, event["startsAt"]),
       :full_data  => event.to_h
     }.merge(
       find_target(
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -99,34 +99,39 @@ def extract_event_data(event)
...
-        :target_type         => node.try(:class).try(:name),
-        :target_id           => node.try(:id),
-      }
-    when 'ExtManagementSystem'
-      {
-        :target_type => 'ExtManagementSystem',
-        :target_id   => @target_ems_id,
-      }
+      if node
+        ret = {
```
cben: cosmetic: consider early return here.  then you don't need `ret` variable and `ret ||` at the end, if you got to the end at all you just return the EMS target.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -99,34 +99,39 @@ def extract_event_data(event)
...
-      }
+      if node
+        return {
+          :container_node_name => labels["instance"],
+          :container_node_id   => node.try(:id),
+          :target_type         => node.try(:class).try(:name),
+          :target_id           => node.try(:id),
+        }
+      else
+        $cn_monitoring_log.error("Could not find node from labels: [#{labels}] defaulting to ems") unless node
```
ilackarms: is this `unless` necessary? we only reach this block if `!node`

moolitayer: Lost in refactor, thanks!

moolitayer:
@ilackarms @enoodle bump

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -80,15 +80,15 @@ def extract_event_data(event)
...
+    event[:ems_ref] = incident_identifier(labels, annotations, event["startsAt"])
     event[:resolved] = event["status"] == "resolved"
     timestamp = event[:resolved] ? event["endsAt"] : event["startsAt"]
     {
       :source     => "DATAWAREHOUSE",
       :timestamp  => timestamp,
       :event_type => "datawarehouse_alert",
       :message    => annotations["message"],
-      :ems_ref    => incident_identifier(event, labels, annotations),
+      :ems_ref    => incident_identifier(labels, annotations, event["startsAt"]),
```
enoodle: nit, why do you cal `incident_identifier` twice? isn't it clearer to use `event[:ems_ref]` here?

moolitayer: mostly because I want to get rid of the upper one when https://github.com/ManageIQ/manageiq/pull/16339 merges and that would allow just dropping that line.

moolitayer:
Updated the default target to be `ext_management_system`, added tests

moolitayer:
@ilackarms please review

ilackarms:
@moolitayer this PR has been built in to custom image builds which are currently working and appear to be showing the correct number of alerts. @joelddiaz can you confirm that alerts mismatch has disappeared since using recent (within past 1 week) images?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
moolitayer: @enoodle FYI I'm removing this line as  ManageIQ/manageiq#16339 is merged and backported

moolitayer:
> @moolitayer this PR has been built in to custom image builds which are currently working and appear to be showing the correct number of alerts. @joelddiaz can you confirm that alerts mismatch has disappeared since using recent (within past 1 week) images?

@ilackarms I plan to merge this one by end of today so we can have it in the beta build done later today.
Any feedback from @joelddiaz will be addressed.

moolitayer:
@ilackarms the beta build is going to be the base for testing starting next week.

simaishi:

```
$ git log -1
commit 5536f1e387864b826ca4c31efb2c802dad52b52a
Date:   Tue Nov 7 15:50:59 2017 +0200

    Merge pull request #149 from moolitayer/prevent_mismatch
    
    Prometheus Alert flow changes
    (cherry picked from commit c202228c2b44af89ad2ee90615e45cd6bcf33575)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1510481
```

PR #151, opened 2017-11-01
Ignore ManagerRefresh::Target unless Graph Refresh is configured

Ladas:
Ignore ManagerRefresh::Target unless Graph Refresh is configured. In current form, the targeted refresh would try to run through old refresh code, resulting in disconecting not sent entities.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresher_mixin.rb
@@ -10,7 +10,7 @@ module ManageIQ
             ems_targets, sub_ems_targets = targets.partition { |x| x.kind_of?(ExtManagementSystem) }
             all_targets = []
 
-            if sub_ems_targets.present?
+            if sub_ems_targets.present? && refresher_options.inventory_object_refresh
               # We can disable targeted refresh with a setting, then we will just do full ems refresh on any event
               ems_event_collection = ManagerRefresh::TargetCollection.new(:targets    => sub_ems_targets,
                                                                           :manager_id => ems_id)


moolitayer:
@Ladas Gaprindashvili/yes right? we are going to need a bz for everything going into Gaprindashvili please

cben:
LGTM :+1:
I expect we'll go into more and more targetted refreshes in future, and we never intended to implement them with old refresh.

Should we log something?  If this keeps happening it indicates you have unnecessary watch worker running (and putting strain on MiqQueue).
But it can also happen for a short while after you properly turned off watch worker and graph refresh, due to existing queued refresh.

Ladas:
@cben I think it should be enough to document, that if we switch to old refresh, then we should disable the collector worker? @agrare or should the CollectorWorker startup condition require refresher_options.inventory_object_refresh ?

moolitayer:
@Ladas I prefer not start the worker with the old refresh, (why document something that we can do our self)

agrare:
> we are going to need a bz for everything going into Gaprindashvili please

This isn't required _yet_, we're still good to backport PRs without a BZ as long as they're reasonable bug fixes.

agrare:
> I prefer not start the worker with the old refresh, (why document something that we can do our self)

I'll push a commit that @Ladas can pick which will not start the collector worker unless inventory_object_refresh is enabled.

agrare:
@moolitayer see https://github.com/ManageIQ/manageiq-providers-openshift/pull/60 and https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/153

Unfortunately since the workers don't inherit from each other and this isn't always going to be true at the base manager level I had to define in both repos.

simaishi:

```
$ git log -1
commit e4c4061b9f15faeb0f3817d698b3a80e8033ce62
Date:   Thu Nov 9 18:13:37 2017 +0200

    Merge pull request #151 from Ladas/run_targeted_refresh_only_for_graph_refresh
    
    Ignore ManagerRefresh::Target unless Graph Refresh is configured
    (cherry picked from commit 47158f544397da5c97b8a1670e52c6d03c643cfd)
```

PR #152, opened 2017-11-01
Make sure we save last occurrence of watch pod

Ladas:
Make sure we save last occurrence of watch pod. The last occurrence  is having the finished_at date. Right now, we were saving the first occurrence, thanks to using .build method (which returns existing InventoryObject if there are duplicates)

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
@@ -37,7 +37,9 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::TargetCollectionMixin
   end
 
   def pod_list
-    target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    targets = target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    # Return only the latest occurrence of each pod
+    targets.index_by { |x| x.metadata.uid }.values
   end
 
   private
--- a/spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
@@ -71,6 +71,8 @@ shared_examples "openshift refresher VCR targeted refresh tests" do
   def normal_refresh(suffix)
     queue_target!('watch_pod_stress_4.json')
     queue_target!('watch_pod_stress_5.json')
+    queue_target!('watch_pod_stress_6_modified.json')
+    queue_target!('watch_pod_stress_6_deleted.json')
 
     VCR.use_cassette(described_class.name.underscore + "_#{suffix}",
                      :match_requests_on => [:path,]) do # , :record => :new_episodes) do
@@ -141,38 +143,38 @@ shared_examples "openshift refresher VCR targeted refresh tests" do
   def expected_table_counts_targeted_refresh_all_nodes_and_namespaces
     base_counts.merge(
       :computer_system          => 9,
-      :container                => 2,
-      :container_condition      => 42,
-      :container_group          => 2,
+      :container                => 3,
+      :container_condition      => 45,
+      :container_group          => 3,
       :container_image          => 1,
       :container_image_registry => 1,
       :container_node           => 9,
       :container_project        => 14,
-      :container_volume         => 2,
-      :custom_attribute         => 58,
+      :container_volume         => 3,
+      :custom_attribute         => 61,
       :ext_management_system    => 1,
       :hardware                 => 9,
       :operating_system         => 9,
-      :security_context         => 2,
+      :security_context         => 3,
     )
   end
 
   def expected_table_counts_targeted_refresh_referenced_nodes_and_namespaces
     base_counts.merge(
       :computer_system          => 1,
-      :container                => 2,
-      :container_condition      => 10,
-      :container_group          => 2,
+      :container                => 3,
+      :container_condition      => 13,
+      :container_group          => 3,
       :container_image          => 1,
       :container_image_registry => 1,
       :container_node           => 1,
       :container_project        => 1,
-      :container_volume         => 2,
-      :custom_attribute         => 12,
+      :container_volume         => 3,
+      :custom_attribute         => 15,
       :ext_management_system    => 1,
       :hardware                 => 1,
       :operating_system         => 1,
-      :security_context         => 2
+      :security_context         => 3
     )
   end
 
@@ -220,10 +222,13 @@ shared_examples "openshift refresher VCR targeted refresh tests" do
   end
 
   def assert_specific_container
-    @container = Container.find_by(:name => "stress4")
+    # Test added Container
+    @container = Container.find_by(:ems_ref => "1b641a4f-aa70-11e7-8a08-001a4a162711_stress4_docker.io/fsimonce/stress-test")
     expect(@container).to have_attributes(
       :name          => "stress4",
       :restart_count => 0,
+      :started_at    => "2017-10-06 08:27:00.000000000 +0000",
+      :finished_at   => nil,
     )
     expect(@container[:backing_ref]).not_to be_nil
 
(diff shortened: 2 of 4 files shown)


Ladas:
@cben @agrare so the index_by relies on having the last version of object in the end. Maybe we should sort it by resource_version first. Also, not sure if we want to do some sort of merging of data, instead of taking the last? Anyway, we can fill the finished_at date for the Container, and we miss column for that in Pod (check the test data, there is deletionTimestamp for Pod cc @moolitayer )

agrare:
@Ladas why not save each one assuming they are all delivered in order?  It doesn't do anything if all we care about is having a record for each pod but if we want to capture the whole history of the pods and e.g. run policies if someone changes the limits to something huge then deletes it, that history would be lost.  Thoughts?

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
```diff
@@ -71,6 +71,8 @@ def parse_notice_pod_ems_ref(pod)
   def normal_refresh(suffix)
     queue_target!('watch_pod_stress_4.json')
     queue_target!('watch_pod_stress_5.json')
+    queue_target!('watch_pod_stress_6_modified.json')
+    queue_target!('watch_pod_stress_6_deleted.json')
```
cben: :heart:

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
```diff
@@ -220,10 +222,13 @@ def assert_ems
   end
 
   def assert_specific_container
-    @container = Container.find_by(:name => "stress4")
+    # Test added Container
+    @container = Container.find_by(:ems_ref => "1b641a4f-aa70-11e7-8a08-001a4a162711_stress4_docker.io/fsimonce/stress-test")
```
cben: can you explicitly assert `.where(:ems_ref...).count == 1`?
(also checked by counts above but that's impossible to figure out)

app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
```diff
@@ -37,7 +37,9 @@ def pods_references(pods)
   end
 
   def pod_list
-    target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    targets = target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    # Return only the latest occurrence of each pod
+    targets.index_by { |x| x.metadata.uid }.values
```
cben: hmm.  neither activesupport docs nor [specs](https://github.com/rails/rails/blob/v5.1.4/activesupport/test/core_ext/enumerable_test.rb#L169-L180) guarantee this clearly.  I'd add a unit test for this function, or better write the hash-filling loop ourselves.

but I'd feel we better sort by resource version.
consider watch worker dying, starting again and getting some older version in the watch initial flood, all queued into one refresh => we may get them out of order here (though last one would probably still be last?).
This *shouldn't* happen, I think, if watching starting `resourceVersion=0`, but there were several bugs around that in kubernetes (and depended on etcd 2 vs 3 at some point), I think it's safer to treat watch starting point as a wish, not a guarantee.

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
Ladas: :+1: right, the full counts are easier in this case, since we have 3 unique ContainerGroup and Container watch events. But I will add the check just for sure.

Ladas:
@agrare we can't really save it all, that would require the sequential target saving. But if we would have examples like you say, with Limits e.g. We can start merging the target data we have. So e.g. we would store 5 different limits of 5 different targets, then next full refresh would disconnect the old? @cben thoughts?

Ladas:
@cben "Ladas is right about build taking first, it calls push aka << which works unless data_index[inventory_object.manager_uuid]. Didn't know that. Why have both build and find_or_build then?I think build still happily returns an InvObj that's not gonna get saved, a risky API."

Right, so the it would make more sense for build to return the last, but I think that could cause dead references. (would work in 5.0 where we turn all links to lazy_links before going through queue). The usual workflow in other providers is find_or_build(uid).assign_attributes(hash), which takes data of the last. 

Anyway, next time we sit together, I sense we should do graph refresh interface 1.0 :-D For now .build takes only first data, find_or_build(uid).assign_attributes(hash) always updates with latest data. (assign attributes does check for defined inventory_object_attributes)

@agrare @cben for storing duplicate version, history, etc. I think we need to redesign the unique indexes. It should be enough to have resource version as part of index? Then either marking the latest as active (the rest archived) or doing the active query using the max resource version. (using the active/archived sounds the best)

Ladas:
@cben this is ready to re-review

app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
```diff
@@ -37,7 +37,9 @@ def pods_references(pods)
   end
 
   def pod_list
-    target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    targets = target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    # Return only the latest occurrence of each pod
+    targets.sort_by { |x| x.metadata.resourceVersion }.index_by { |x| x.metadata.uid }.values
```
cben: Still relies on `index_by`, and also does O(N*log N) for big N (yeah right, if those were our problems :smile: ).
How about
```
targets.group_by { |x| x.metadata.uid }.transform_values { |xs| xs.max_by { x.metadata.resourceVersion } }.values
```
or
```
latest_by_uid = {}
targets.each do |t|
  uid = x.metadata.uid
  unless latest_by_uid[uid] && latest_by_uid[uid].metadata.resourceVersion > t.metadata.resourceVersion
    latest_by_uid[uid] = t
  end
end
```

Oh, and there is a final snag: I think resourceVersion is a string, any sort/comparison needs .to_i ?

Ladas:
```
       user     system      total        real
using sort_by
  1.410000   0.050000   1.460000 (  1.465202)
using group_by
  1.230000   0.010000   1.240000 (  1.231996)
using each
  0.620000   0.010000   0.630000 (  0.642298)
```

this is on 2M array elements, filtered to 1M unique, based on resource version. I will use the last one, but the time save is not that huge :-) The O(N * log N) is usually leveled by having those methods in C. :-)

cben: OK, forget all I said here.
As discussed [on gitter](https://gitter.im/ManageIQ/manageiq?at=59fb2272e44c43700aab4fb7), and with Mooli... 
 we can't compare/sort by resourceVersion.
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#concurrency-control-and-consistency  insists resourceVersion it's an opaque string.

So drop the sort, just take last.
For Gaprindashvili, it will be fine, since we'll only take DELETED notices.

Longer term, we probably persist last resourceVersion we're "synced" with, and only watch from it.
(and never watch without resourceVersion / with resourceVersion=0)

app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
moolitayer: Sorry for nit picking, but I don't see anywhere that index_by API guarantees to return the last occurrence.
I would be more at ease if that was done explicitly

Ladas: right so @cben was checking the code and it should work like that, nevertheless, for g-release, we will fetch only DELETED watch events, so hopefully just 1 copy per pod. If not, all of them should contain finished_at date

moolitayer: I saw the code to be working, I'm concerned that it's an outcome of the implementation and not
part of the API of `index_by` so i'm asking for an explicit implementation to be on the safe side.

(for master, or G if you choose to backport even though it should have no effect as you explained)

PR #153, opened 2017-11-01
Don't start CollectorWorker if Graph Refresh disabled

agrare:
Vintage :TM: refresh for container providers doesn't reconnect container
images so having duplicate targets queued by the
InventoryCollectorWorker can lead to duplicate records.  For this reason
we want to disable the InventoryCollectorWorker if
`:inventory_object_refresh` is disabled.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker.rb
@@ -1,3 +1,7 @@
 class ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollectorWorker < ManageIQ::Providers::BaseManager::InventoryCollectorWorker
   require_nested :Runner
+
+  def self.has_required_role?
+    !worker_settings[:disabled] && Settings.fetch_path(:ems_refresh, ems_class.ems_type.to_sym, :inventory_object_refresh)
+  end
 end


app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker.rb
```diff
@@ -1,3 +1,7 @@
 class ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollectorWorker < ManageIQ::Providers::BaseManager::InventoryCollectorWorker
   require_nested :Runner
+
+  def self.has_required_role?
+    !worker_settings[:disabled] && Settings.fetch_path(:ems_refresh, ems_class.ems_type.to_sym, :inventory_object_refresh)
```
agrare: TODO I think `Settings.fetch_path(:ems_refresh, ems_class.ems_type.to_sym, :inventory_object_refresh)` needs a helper method somewhere, maybe on the base_manager `::Refresher` ?

cben:
Are images the *only* problem?
Cause Vintage :TM: full refresh can duplicate images too...
I'm all for this in any case.

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_worker.rb
cben: I won't object but I'm equally happy with this as-is, seeing settings full pathes makes it easy to find setting from code and vice versa...
(and hey, average settings path is shorter than our::codebase::average::class::name ;-)

moolitayer:
@agrare G labels please

simaishi:

```
$ git log -1
commit 9853ae91a66cd83a3ea5839e9fc143d7ca0bcc29
Date:   Thu Nov 2 12:31:27 2017 +0200

    Merge pull request #153 from agrare/disable_collector_worker_without_graph_refresh
    
    Don't start CollectorWorker if Graph Refresh disabled
    (cherry picked from commit d27c1b8516548218f7e05c16dd7e63922f0808ac)
```

PR #154, opened 2017-11-01
Add a setting to only send DELETED notices

agrare:
This adds a setting to the InventoryCollectorWorker which allows for only DELETED pod notices to be sent, thus reducing the traffic hitting MiqQueue.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -70,6 +70,7 @@
         :poll: 20.seconds
     :ems_inventory_collector_worker:
       :ems_inventory_collector_worker_kubernetes:
+        :deleted_notices_only: true
         :disabled: false
         :watch_thread_shutdown_timeout: 10.seconds
     :queue_worker_base:
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
@@ -100,6 +100,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollectorMixi
   def watch_thread
     pod_watch_stream.each do |notice|
       break if exit_requested
+      next if notice.type != "DELETED" && worker_settings[:deleted_notices_only]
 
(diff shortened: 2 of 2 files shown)


moolitayer:
@agrare can you please explain why/when `:deleted_notices_only: true` is useful?
(does not seem to give what you originally made the watch for)

config/settings.yml
```diff
@@ -70,6 +70,7 @@
         :poll: 20.seconds
     :ems_inventory_collector_worker:
       :ems_inventory_collector_worker_kubernetes:
+        :deleted_notices_only: false
```
Ladas: I wonder if we should rather expose what types to store? filter_notice_types: ['DELETED', ...] . As it might be ok to do DELETED + MODIFIED for g-release?

agrare:
@oolitayer the content of the delete notice also still has the full spec of the pod so we can create the pod record from just the deletes.  If we only care about not missing a pod then this will satisfy that.
I left it default to false so that we can get better testing of how the queue reacts to a large number of targets then we can switch it back before release if we want.

moolitayer:
> I left it default to false so that we can get better testing of how the queue reacts to a large number of targets then we can switch it back before release if we want.

If we plan to release with true It's seems desired to test if the delete really is enough for the metrics collection. If that should be enough for us maybe we don't need to stress + test the queue any more then that?

config/settings.yml
Ladas: ok

Ladas:
@agrare @moolitayer makes sense to switch on true by default and test mainly that. We can test with false on some big env later.

agrare:
Okay defaulted to true

simaishi:

```
$ git log -1
commit b947c1aa7af2ef31aad6a0a831157dd40629793d
Date:   Thu Nov 2 15:49:17 2017 +0200

    Merge pull request #154 from agrare/setting_to_only_send_deleted_notices
    
    Add a setting to only send DELETED notices
    (cherry picked from commit 41a8bbeb45e6849b418a3f78d0377cb23f45528e)
```

PR #157, opened 2017-11-03
Update master changelog Gaprindashvili B1

JPrause:
[skip ci]


JPrause:
@miq-bot add_label documentation

moolitayer:
@JPrause we had the same PR Gaprindashvili and now for master, Maybe next time we should update only master and then do the normal backport using gaprindashvili/yes?
Is that the plan?

JPrause:
Thanks @moolitayer and I see your point. The two branches will diverge and the gaprindashvili branch should only contain it's future release data (ie, not include e-release data), whereas master is the full running log. I also think what you're seeing is a confluence of multiple releases happening at the same time. We also have constraints as to when we can branch, tag, and build these releases. Almost a "chicken and egg" scenario.

PR #158, opened 2017-11-06
Prometheus unavailable alert

moolitayer:
Contains part of https://github.com/ManageIQ/manageiq/issues/16395

The PR will catch the most common cases where Prometheus is unavailable/returns unexpected response. It won't catch errors where the platform does not start the collector - e.g auth failure. When the collection returns to normal, the alert should resolve. This allows seeing alerts in the monitoring screen when Prometheus is unavailable.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
@@ -1,6 +1,8 @@
 class ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::Stream
   def initialize(ems)
     @ems = ems
+    @current_generation = nil
+    @open_incident = nil
   end
 
   def start
@@ -20,10 +22,10 @@ class ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::Stream
   end
 
   def fetch
-    unless @current_generation
-      @current_generation, @current_index = last_position
-    end
+    check_state
     $cn_monitoring_log.info("Fetching alerts. Generation: [#{@current_generation}/#{@current_index}]")
+    alert_list, error = poll_messages
+    return [error] if error
 
     # {
     #   "generationID":"323e0863-f501-4896-b7dc-353cf863597d",
@@ -38,8 +40,13 @@ class ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::Stream
     #   ...
     #   ]
     # }
-    alert_list = @ems.connect.get(:generation_id => @current_generation, :from_index => @current_index)
     alerts = []
+
+    if collection_incident_ongoing?
+      alerts << resolve_collection_incident
+      $cn_monitoring_log.info("[#{@ems.name}] collection problem resolved")
+    end
+
     @current_generation = alert_list["generationID"]
     return alerts if alert_list['messages'].blank?
     alert_list["messages"].each do |message|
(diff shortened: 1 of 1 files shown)


moolitayer:
@agrare please review - as part of the discussion on event collector unavailable alert. Is this an acceptable partial solution until we have something better?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -81,4 +91,38 @@ def last_position
...
+    )
+    @ems.parent_manager.options[:last_collection_problem] = nil
+    @ems.parent_manager.save!
+    event
+  end
+
+  def status_event_common
+    {
+      "annotations" => {
+        "url"       => "www.example.com",
```
moolitayer: This should be a URL that explains how to resolve the error. I think I will take this from settings

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -81,4 +91,38 @@ def last_position
       last_index ? last_index + 1 : 0
     ]
   end
+
+  def collection_compromised_event
+    # TODO: Using options for this end is a hack, find a better solution
+    @ems.parent_manager.options[:last_collection_problem] ||= Time.zone.now
+    @ems.parent_manager.save!
```
moolitayer: The reason we need to persist the start time is that it is used for resolving of alerts:

https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/e3a711af542f8ba68373e3cd5e2e17198c1b3f51/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb#L91

https://github.com/ManageIQ/manageiq/blob/57b5da5bde3227154d4e1173f28cd324ea266b01/app/models/miq_alert.rb#L251

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
```diff
@@ -81,4 +91,38 @@ def last_position
...
+      "status" => "firing"
+    )
+  end
+
+  def collection_running_event
+    event = status_event_common.merge(
+      "status" => "resolved"
+    )
+    @ems.parent_manager.options[:last_collection_problem] = nil
+    @ems.parent_manager.save!
```
agrare: How often would this save the manager?  On every successful alert?

moolitayer: On every iteration which is currently 20 sec. Good point, I will do it only if changed

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/stream.rb
cben: I was just considering something similar to persist last refresh resourceVersion :-)
The migration-less hack I was thinking of is a CustomAttribute.

moolitayer: :+1: That sounds better then using options.

moolitayer:
@zeari @cben @agrare 

I've pushed a new rebased version and addressed comments.
Honestly I don't like the current solution persisting EmsEvent every queue interval(alerts working).
Also it would have to be almost rewritten after we move to use a streaming API.

I believe raising an EmsEvent might be the solution, I will try to explore that.
The down side is need for policies to be defined. I wonder if we could use the same policy somehow.

moolitayer:
I closing this as it is not a complete solution and would have to be removed once we do have one

PR #160, opened 2017-11-06
Disable inventory collector worker by default

agrare:
Due to the large payload size generated by the watch notices disable the collector worker by default until payloads can be moved to BinaryBlobs table

PR #161, opened 2017-11-06
disable collector worker for now

kbrock:
We want to disable the k8s collector worker for now. It is generating sql statements that are too big for postgres to handle

The data is quite large. 0.5Gb
It is blowing up miq_queue table and pg memory

```
PG::OutOfMemory: ERROR:  out of memory
DETAIL:  Failed on request of size 536,870,912.
: UPDATE "miq_queue" SET "msg_data" = '\x0...'
```


kbrock:
duplicate of #160

PR #164, opened 2017-11-08
Set the target payload for watch inventory collector

agrare:
Set the target payload explicitly and clear it when the refresh is finished.  The post_refresh_ems_cleanup method is called at least from here https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb#L41-L43 so it will be called even if refresh throws an exception.

Related to: https://github.com/ManageIQ/manageiq/pull/16413
                  https://github.com/ManageIQ/manageiq-providers-openshift/pull/69

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
@@ -63,14 +63,13 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollectorMixi
 
     ems_ref = parse_notice_pod_ems_ref(object)
 
-    ManagerRefresh::Target.new(
+    target_opts = {
       :manager     => ems,
       :association => :container_groups,
-      :manager_ref => ems_ref,
-      :options     => {
-        :payload => object.to_json,
-      }
-    )
+      :manager_ref => ems_ref
+    }
+
+    ManagerRefresh::Target.new(target_opts).tap { |target| target.payload = object.to_json }
   end
 
   def stop_watch_thread
--- a/app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
@@ -37,7 +37,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::TargetCollectionMixin
   end
 
   def pod_list
-    target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    target.targets.map { |target| JSON.parse(target.payload, :object_class => OpenStruct) }
   end
 
   private
--- a/app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
@@ -37,7 +37,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::TargetCollectionMixin
   end
 
   def pod_list
-    target.targets.map { |target| JSON.parse(target.payload, :object_class => OpenStruct) }
+    target.targets.map do |target|
+      payload = target.payload
+      next if payload.nil?
+
+      JSON.parse(payload, :object_class => OpenStruct)
+    end.compact
   end
 
   private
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/kubernetes/container_manager/target_collection_mixin.rb
```diff
@@ -37,7 +37,7 @@ def pods_references(pods)
   end
 
   def pod_list
-    target.targets.map { |target| JSON.parse(target.options[:payload], :object_class => OpenStruct) }
+    target.targets.map { |target| JSON.parse(target.payload, :object_class => OpenStruct) }
```
Ladas:
```ruby
target.targets.map { |target| JSON.parse(target.payload, :object_class => OpenStruct) if target.payload }.compact
```

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
```diff
@@ -62,10 +62,7 @@ def queue_target!(watch_data_path)
       :manager     => @ems,
       :association => :container_groups,
       :manager_ref => ems_ref,
-      :options     => {
-        :payload => notice['object'].to_json,
-      },
-    )
+    ).tap { |t| t.payload = notice['object'].to_json }
```
cben: [cosmetic, your call] don't need .tap, can simply `target.payload = ...` on separate line.

agrare: I was keeping this the same as the call in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/164/files#diff-03ceb0538d4c3023378bbb63123f770dR72 but that is returning the result in that line, you're right no need for the tap here

agrare:
> are PR title & description still accurate? I'm not sure where clearing is now done...

Not anymore, at first I was clearing the payload in a post_ems_refresh handler but now we have an association between the queue item and the binary blob so we don't need to explicitly clear the payload here.  I'll update the description

spec/models/manageiq/providers/kubernetes/container_manager/targeted_refresh/targeted_refresh_spec.rb
cben: sorry, that was 100% silly bikeshedding compared to the non-trivial content of these PRs, doesnt matter either way :bike: :house:

agrare: You did say it was cosmetic and I agree this is better so no need to apologize :)
We should have @Ladas update his PR (he did the same thing) https://github.com/ManageIQ/manageiq-providers-openshift/pull/68/files#diff-d6b73355e83b934cb272f17e43c007e3R61 for consistency

Ladas: updated, new PR https://github.com/ManageIQ/manageiq-providers-openshift/pull/69

moolitayer:
@kbrock please review

moolitayer:
@kbrock @cben please take a look

app/models/manageiq/providers/kubernetes/container_manager/inventory_collector_mixin.rb
```diff
@@ -63,14 +63,13 @@ def parse_target_from_notice(notice)
...
       :association => :container_groups,
-      :manager_ref => ems_ref,
-      :options     => {
-        :payload => object.to_json,
-      }
-    )
+      :manager_ref => ems_ref
+    }
+
+    ManagerRefresh::Target.new(target_opts).tap { |target| target.payload = object.to_json }
```
cben: What took me a while to grok about this code is that for 1 conceptual goal of "queue this via DB" is split between 2 *different looking* code channels:
- a BinaryBlob is written to DB immediately in `Target#payload=`.
- the Target, holding among others blob id, is an in-memory ruby object — not even an ActiveRecord — whose content gets written into DB later, inside `EmsRefresh.queue_refresh` via `Target#dump`.

[not opposing and don't have any better suggestion, just hoping re-arch obsoletes this by giving us a more scalable queue...]

cben:
Ah, I see, https://github.com/ManageIQ/manageiq/pull/16413 dependency is still unmerged.

PR #167, opened 2017-11-09
Use Proetheus timeouts form settings

yaacov:
Description

Metrics collection from OCP 3.7 using Prometheus may hit timeout while collecting metrics.

a. the default timeouts are too small
b. users can not set new timeout values to fit their system

This PR uses the settings needed to adjust the timeouts.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1511341

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
@@ -10,6 +10,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Promet
   end
 
   def prometheus_client_new(uri, credentials, options)
+    worker_class = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCollectorWorker
+
     Faraday.new(
       :url     => uri.to_s,
       :proxy   => options[:http_proxy_uri].empty? ? nil : options[:http_proxy_uri],
@@ -18,8 +20,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Promet
         :cert_store => options[:ssl_cert_store]
       },
       :request => {
-        :open_timeout => 2, # opening a connection
-        :timeout      => 5  # waiting for response
+        :open_timeout => worker_class.worker_settings[:prometheus_open_timeout] || 5,
+        :timeout      => worker_class.worker_settings[:prometheus_request_timeout] || 30
       },
       :headers => {
         :Authorization => "Bearer " + credentials[:token]
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -78,5 +78,7 @@
           :ems_metrics_collector_worker_kubernetes:
             :metrics_port: 5000
             :metrics_path: "/hawkular/metrics"
+            :prometheus_open_timeout: 5
+            :prometheus_request_timeout: 30
         :ems_refresh_worker:
           :ems_refresh_worker_kubernetes: {}


yaacov:
@miq-bot add_label bug

@simon3z @moolitayer @ilackarms @joelddiaz please review

cben:
These are new settings right?
Please add them to config/settings.yml, instead of defaults in code.
It's hard to track what is "settable" if one has to infer it from the code...

yaacov:
> Please add them to config/settings.yml, instead of defaults in code.

https://github.com/ManageIQ/manageiq/pull/16429 ?

yaacov:
@cben 👍 

closed ManageIQ/manageiq#16429 , will move the settings to local setting.yml file

joelddiaz:
lgtm

simaishi:
@yaacov  `gaprindashvili/yes` ?

yaacov:
> @yaacov gaprindashvili/yes ?

yes

ilackarms:
@yaacov wondering how this will be configurable from openshift in the manageiq-pod

joelddiaz:
IMO, ideally we want some way for the automation in openshift-ansible to be able to accept configuring these settings

yaacov:
> IMO, ideally we want some way for the automation in openshift-ansible to be able to accept configuring these settings

We will have to ask ( the api guys ? ) if we have an api for setting the settings ? @moolitayer ?

@cben suggested adding that to the new tags/labels @enoodle added for the provider, do we have api for those ?

cben:
I think that's orthogonal.
- If there is use case to configure these per-provider, we can make these
per-provider.
  I don't see the point here, if user needs to increase these timeouts,
there is no harm in increasing for all providers.
- There should be API (and then) ansible for settings.
  And not only Advanced Settings aka settings.yml, ideally all things under
top right menu -> Configuration should have API(s).

yaacov:
> There should be API (and then) ansible for settings.

@cben Do you know what is the api path for that ?

p.s.
I suggested the tags in case their is no api for settings, and their is one for tags ...

cben:
Sorry, I meant such API ought to be added :-)
There exists
https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/settings_controller.rb
but it has a narrow whitelist:
https://github.com/ManageIQ/manageiq-api/blob/220cbbb927/config/api.yml#L2543

- `product` (which only contains deployment wizard)
- `prototype` (which contains MiqQueue/Artemis knobs)

moolitayer:
> We will have to ask ( the api guys ? ) if we have an api for setting the settings ? @moolitayer ?

- we have an API for selected settings, we must actively include what needs to be exposed[1]
- I agree with @cben about the orthogonality of per provider setting - we should add those if they make sense and not because we can API them. I think they do make sense (e.g one provider is behind a slower network) I seem to recall a api PR for that, @enoodle can you confirm we have an API for per provider settings (`options` right?)?

[1] https://github.com/ManageIQ/manageiq-api/commit/cb85286fb70b32ede69488093ecba8d6fb1ed631

moolitayer:
> We have an API for selected settings, we must actively include what needs to be exposed[1]

I see @cben already [commented](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/167#issuecomment-344016081) about that

enoodle:
@cben @yaacov can you take a look at [1] and if it is what you are looking for then the PRs [2][3] it links to see how to configure per provider settings.

[1] https://github.com/ManageIQ/manageiq-ui-classic/pull/1652
[2] ManageIQ/manageiq-providers-kubernetes#45
[3] https://github.com/ManageIQ/manageiq-providers-openshift/pull/32

moolitayer:
@enoodle @joelddiaz is looking for an API

enoodle:
@moolitayer Currently that API for per-provider options that I created is not returning the values, but only the available options per provider. I think it should be changed to also return the values, or another one should be created for that.

moolitayer:
@enoodle what about setting them?

enoodle:
@moolitayer Yes, I will make a PR to add getting/setting the options per provider in another endpoint (something like /api/provider/<ID>/settings )

simaishi:

```
$ git log -1
commit 62f6d243910e2cba0dda09628af25e786cb3fb5f
Date:   Sun Nov 12 15:21:15 2017 +0200

    Merge pull request #167 from yaacov/use-prometheus-timeouts
    
    Use Proetheus timeouts form settings
    (cherry picked from commit 17825b2fdefaffd8b81a1fddf0818c33fb5a4435)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1513116
```

enoodle:
I created https://github.com/ManageIQ/manageiq-api/pull/211 to allow setting the options field per provider. For example:
```
curl --user admin:smartvm -X POST -d '{"action":"edit","options":{"image_inspector_options":{"image_tag":"2.1"}}}' http://localhost:3000/api/providers/1
```

PR #168, opened 2017-11-13
ContainerServicePortConfigs using [:container_service, :name]

Ladas:
ContainerServicePortConfigs using [:container_service, :name]
as manager_ref


cben:
See also https://github.com/ManageIQ/manageiq/pull/16454 for old refresh.  Can be merged independently I believe.

LGTM.   As discussed on https://github.com/ManageIQ/manageiq/pull/16454 `name` might be nil — it's optional IFF there is only 1 port => still unique within a service.

Ladas:
continues in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/212

the name doesn't seem to be uniq

PR #162, opened 2017-11-08
Tag mapping in graph refresh

cben:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/16453 for specs
- [x] https://github.com/ManageIQ/manageiq/pull/16499 for fixing that spec
- [x] https://github.com/ManageIQ/manageiq/pull/16501 to define `Tag.controlled_by_mapping` scope — merged
    (~~OR revive reverted https://github.com/ManageIQ/manageiq/pull/16449~~)

Overview/plan issue: #126

(added test for not removing user-assigned tag has been extracted to #170 and already merged)

@miq-bot add-label enhancement gaprindashvili/yes
https://bugzilla.redhat.com/show_bug.cgi?id=1470021

@Ladas @moolitayer please review

PR #170, opened 2017-11-16
Improve tag mapping specs

cben:
Split from #162

- Use factory for more accurate test, matching UI-created mapping categories.
  Previous test didn't set read_only nor used 'kubernetes:' prefix,
  failed with attempted retagging implementation https://github.com/ManageIQ/manageiq/pull/16449
  and will fail with its upcoming replacement.

- Test that we don't remove user-assigned tags.

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
@@ -24,40 +24,45 @@ shared_examples "kubernetes refresher VCR tests" do |check_tag_mapping: true|
 
   # Smoke test the use of ContainerLabelTagMapping during refresh.
   before :each do
-    @name_category = FactoryGirl.create(:classification, :name => 'name', :description => 'Name')
-    @label_tag_mapping = FactoryGirl.create(
-      :container_label_tag_mapping,
-      :label_name => 'name', :tag => @name_category.tag
-    )
+    mapping = FactoryGirl.create(:tag_mapping_with_category, :label_name => 'name')
+    @name_category = mapping.tag.category
+
+    @user_tag = FactoryGirl.create(:classification_cost_center_with_tags).entries.first.tag
   end
 
-  def full_refresh_test
-    3.times do # Run three times to verify that second & third runs with existing data do not change anything
-      VCR.use_cassette(described_class.name.underscore) do # , :record => :new_episodes) do
-        EmsRefresh.refresh(@ems)
-      end
-      @ems.reload
-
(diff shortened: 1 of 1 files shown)


cben:
@miq-bot add-label test, gaprindashvili/yes

@zeari @moolitayer Please review

moolitayer:
@cben so this change is testing that additional tags that did not come back in refresh are not removed?
Do we always not remove tags that did not come back in refresh? if not, In which cases do we not delete them?

cben:
@moolitayer This tests not removing a user-defined user-assigned tag.
It's assigned after 1st refresh because the object must exist for user to do Edit Tags.

We should of course be removing mapping-generated mapping-assigned tags, when the label disappears.
That will be tested in similar openshift PR I'm sending soon, where we such situation in the cassette.

cben:
Fixed potential double tagging on 3rd refresh.
Rewrote for clearer data flow by arg passing, unrolled the loop, improved comment (I hope).
PTAL.

spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
```diff
@@ -24,40 +24,45 @@
...
+    # Volumes, PVs, and PVCs are tested in _before_deletions VCR.
   end
 
   it "will perform a full refresh on k8s" do
+    # Run three times to verify that second & third runs with existing data do not change anything
     full_refresh_test
+
+    # Now records exist, simulate user assigning tags by Edit Tags, to test later refreshes don't remove them.
+    @replicator.reload.tags |= [@user_tag]
+    @containergroup.reload.tags |= [@user_tag]
```
moolitayer: I might be missing something, but if the refresh does remove the user tags won't this code re set them?

cben: No.  This was a tricky point with the loop but is simpler now it's unrolled:
here is the one and *only* time we set the user tag.
next 2 full refreshes only perform refresh, and assert the tag is still assigned.

Do you see any clearer way to write this?  I'm really happy with the structure I reached currently, but what matters is somebody else being able to understand...

simaishi:

```
$ git log -1
commit d2e40c479ac04aa557d7b714dbab3f6827b3cd52
Date:   Sun Nov 19 23:00:20 2017 +0200

    Merge pull request #170 from cben/tag-mapping-tests
    
    Improve tag mapping specs
    (cherry picked from commit ef6039818308c25baab8abadaf124aa28952e108)
```

PR #171, opened 2017-11-20
Enable getting multiple alert for an incident

moolitayer:
Currently when an incident starts (e.g Container Node High Memory Usage) we get notified ONCE according to our selected notification method. We will be notified again when the incident is resolved.
For our use case (operators monitoring a cluster) one would want to keep getting alerts on an open issue until it is fixed. 

Prometheus keeps sending out messages when an incident is open but we created only one EmsEvent out of them since events with a similar time stamp mask each other and Prometheus startsAt is the same for alerts coming from the same incident.

ems_ref should still consider startsAt since that would still make sure we are only creating one MiqAlertStatus per incident.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
@@ -82,10 +82,9 @@ module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerM
     # TODO(mtayer): remove after https://github.com/ManageIQ/manageiq/pull/16339
     event[:ems_ref] = incident_identifier(labels, annotations, event["startsAt"])
     event[:resolved] = event["status"] == "resolved"
-    timestamp = event[:resolved] ? event["endsAt"] : event["startsAt"]
     {
       :source     => "DATAWAREHOUSE",
-      :timestamp  => timestamp,
+      :timestamp  => Time.zone.now,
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -82,10 +82,9 @@ def extract_event_data(event)
     # TODO(mtayer): remove after https://github.com/ManageIQ/manageiq/pull/16339
     event[:ems_ref] = incident_identifier(labels, annotations, event["startsAt"])
     event[:resolved] = event["status"] == "resolved"
-    timestamp = event[:resolved] ? event["endsAt"] : event["startsAt"]
     {
       :source     => "DATAWAREHOUSE",
-      :timestamp  => timestamp,
+      :timestamp  => Time.zone.now,
```
moolitayer: would be better to do this in `process_alert!` in stream.rb probably

zeari: I think its fine here.

simaishi:

```
$ git log -1
commit add7721a201fd2ab504db82903980f7b3dac701c
Date:   Mon Nov 27 13:19:38 2017 -0500

    Merge pull request #171 from moolitayer/enable_multiple_incident_notifications
    
    Enable getting multiple alert for an incident
    (cherry picked from commit 74bbd13de22e51a62e9a8377d9e1ed057d76802d)
```

PR #173, opened 2017-11-21
Update hawkular gem version

josejulio:
We have released a new version of the `hawkular-client` gem.
It changes the inventory and operations, I don't think this should affect your code in any way.

In a close future we are going to start using prometheus instead of hawkular metrics, so we will remove metrics from the `Hawkular::Client` [1]. 
It will still remain in the code, you will have to instantiate it explicitly (I think you already do that).

[1] https://github.com/hawkular/hawkular-client-ruby/blob/master/lib/hawkular/hawkular_client.rb#L10

### DIFF
--- a/manageiq-providers-kubernetes.gemspec
+++ b/manageiq-providers-kubernetes.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "hawkular-client", "~> 4.1"
+  s.add_runtime_dependency "hawkular-client", "~> 5.0.0.pre1"
   s.add_runtime_dependency "image-inspector-client",  "~>1.0.3"
   s.add_runtime_dependency "kubeclient",      "~>2.4.0"
 


josejulio:
@cben can you please take a look or suggest who could? Thanks!

josejulio:
@miq-bot add_label gaprindashvili/no

josejulio:
@moolitayer please take a look, we are going to need the 5.0.0-pre version for our latest changes and don't want to break anything.

bazulay:
cc @yaacov

yaacov:
@josejulio on my machine the hawkular-client gem stay 4.1 after this change, what am I missing ?

cben:
`>= 4.1` won't pick up any version with letters, only "stable" releases.
See https://github.com/bundler/bundler/issues/1198 and http://guides.rubygems.org/patterns/#prerelease-gems

yaacov:
@josejulio 

a. LGTM for now ( use v4.1 )nothing changed
b. once v5.0 stable is out, we will have to re-check ...

josejulio:
@yaacov @cben sorry for the confusion.
It seems that `bundler` picks up `5.0.0.pre1` version if we also use these changes (from manageiq-providers-hawkular) https://github.com/ManageIQ/manageiq-providers-hawkular/pull/75/files#diff-0a12919b1cdb7af70a52bb3e466613cfL16

```
Using hawkular-client 5.0.0.pre1
```

I'm not really sure how to handle this case, as we need to bump the gem version but we need to do it in sync with you, else `bundler` will complain.

```
Bundler could not find compatible versions for gem "hawkular-client":
  In snapshot (Gemfile.lock):
    hawkular-client (= 5.0.0.pre1)

  In Gemfile:
    manageiq-providers-hawkular was resolved to 0.1.0, which depends on
      hawkular-client (~> 5.0.0.pre1)

    manageiq-providers-kubernetes was resolved to 0.1.0, which depends on
      hawkular-client (~> 4.1)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

I can explicitly specify the gem version to be `5.0.0.pre1` if you think that's better.

yaacov:
@josejulio thanks, I hope to have time to check metrics collection tomorrow 👍

yaacov:
@josejulio @cben I get gemfile conflict ... what am I missing, how do I pass this ?
```
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    manageiq-gems-pending (> 0) was resolved to 0.1.0, which depends on
      actionpack (< 5.2, >= 5.0)

    manageiq-schema was resolved to 0.1.0, which depends on
      rails (~> 5.0.2) was resolved to 5.0.6, which depends on
        actionpack (= 5.0.6)
...
...
```


help :-)

yaacov:
```
bundle show hawkular-client
/home/yzamir/.gem/ruby/2.4.0/gems/hawkular-client-5.0.0.pre1
```

Reading container metrics works for me :-)

LGTM 👍

josejulio:
We are currently under a big change, will reopen this PR once we are ready to merge. Thanks!

PR #176, opened 2017-11-22
Add i18n string placeholders test

mzazrivec:
This is just invocation of standard i18n placeholder test, included from main ManageIQ
repository. It's the same thing that we do in ui-classic to test that the provided translations
honor placeholders in the original (English) strings.

### DIFF
--- a/spec/i18n/placeholders_spec.rb
+++ b/spec/i18n/placeholders_spec.rb
@@ -1,0 +1,3 @@
+describe :placeholders do
+  include_examples :placeholders, ManageIQ::Providers::Kubernetes::Engine.root.join('locale').to_s
+end


spec/i18n/placeholders_spec.rb
```diff
@@ -0,0 +1,3 @@
+describe :placeholders do
+  include_examples :placeholders, ManageIQ::Providers::Kubernetes::Engine.root.join('locale').to_s
```
cben: I see, these come from https://github.com/mzazrivec/manageiq/blob/master/spec/shared/i18n/placeholders.rb

In https://github.com/ManageIQ/manageiq/pull/13202, there was discussion of adding Vmdb::Plugins method to return translations directory — did anything like that happen?
(current `.join(locale)` LGTM too)

mzazrivec: Not that I'm aware of.

moolitayer:
something I don't get, if the specs are coming from the main repository how come we are including examples from the plugin path?

mzazrivec:
@moolitayer `ManageIQ::Providers::Kubernetes::Engine.root.join('locale').to_s` is just
an argument for the shared spec (where to look for the gettext catalogs to test).

cben:
```
[4] pry(main)> ManageIQ.root
=> #<Pathname:/home/bpaskinc/miq/master>
[5] pry(main)> ManageIQ::Providers::Kubernetes::Engine.root
=> #<Pathname:/home/bpaskinc/miq/master/plugins/manageiq-providers-kubernetes>
```

PR #177, opened 2017-11-27
Make sure Container has always the right STI type

Ladas:
Make sure Container has always the right STI type. Before, the
missing image could cause the STI type is Container, which
made metrics capture fail.

Fixes BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1517676

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -1032,6 +1032,7 @@ module ManageIQ::Providers::Kubernetes
 
     def parse_container_spec(container_spec, pod_id)
       new_result = {
+        :type                 => 'ManageIQ::Providers::Kubernetes::ContainerManager::Container',
         :ems_ref              => "#{pod_id}_#{container_spec.name}_#{container_spec.image}",
         :name                 => container_spec.name,
         :image                => container_spec.image,
@@ -1079,7 +1080,6 @@ module ManageIQ::Providers::Kubernetes
       return if container_image.nil?
 
       h = {
-        :type            => 'ManageIQ::Providers::Kubernetes::ContainerManager::Container',
         :restart_count   => container.restartCount,
         :backing_ref     => container.containerID,
         :container_image => container_image


Ladas:
cc @simon3z

bazulay:
@moolitayer  PTAL

cben:
gaprindashvili/?

Ladas:
@cben should be g-release, or even more?

cben:
@miq-bot add-label gaprindashvili/yes
https://github.com/ManageIQ/manageiq-providers-kubernetes/blame/gaprindashvili/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L1082

I think fine/yes euwe/yes darga/yes capablanca/yes?! too (will need manual backport):
https://github.com/ManageIQ/manageiq/blame/fine/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L655
https://github.com/ManageIQ/manageiq/blame/euwe/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L628
https://github.com/ManageIQ/manageiq/blame/darga/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L631
https://github.com/ManageIQ/manageiq/blame/capablanca/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L522
type was set in parse_container (which became parse_container_status), not parse_container_definition (which became parse_container_spec).
@zeari did I read that correctly?
@bazulay how far back do we maintain?

cben:
On second thoughts, I think on fine and older, before ContainerDefinition merge into Container, there is no bug.
Yes, container_status is not always present, but then Container record is just not created at all, and no metrics is collected, so no error message.  IIUC only very new, not yet running, containers have no status, so not collecting metrics is fine.
@zeari please double-check my thinking.

And on gaprindashvili and master, the BZ symptom is not very severe, if it only happens to very new containers (though inventory may be stale, we might already get some metrics by time we collect).
But in any case having wrong STI type can lead to many weird problems, important to fix!

moolitayer:
@Ladas thanks for the fix.
Could you please add a spec that fails before this fix?
(parsing is enough, refresh_parser_spec.rb)

zeari:
> On second thoughts, I think on fine and older, before ContainerDefinition merge into Container, there is no bug.
> Yes, container_status is not always present, but then Container record is just not created at all, and no metrics is collected, so no error message. IIUC only very new, not yet running, containers have no status, so not collecting metrics is fine.
> @zeari please double-check my thinking.
> 

I agree, this is a new issue. either way if it does pop up we have @Ladas manual fix in https://bugzilla.redhat.com/show_bug.cgi?id=1517676

Ladas:
@moolitayer @cben hm, could you guys do that magic with manual editing VCRs? In our VCR data, we do not have Container without image.

cben:
I'd like to avoid manual editing, but I'm right now working on rerecording
a VCR, I'll try to add such container
(eg. a bad image url should reliably result in container without status).

2017-12-04 9:59 GMT+02:00 Ladislav Smola <notifications@github.com>:

moolitayer:
@Ladas refresh_parser_spec.rb does not use VCR and would be enough, thanks

Ladas:
@cben not sure if refresh_parser_spec.rb is good for this? I see you use it mainly for 'it will be parsed good'. While in other providers, this is more of refresher_spec makes sure the right STI type is always there. (also in other providers, we are not setting the type in parser, unless there are sub STI types)

Ladas:
@cben @moolitayer is it ok to just merge this to deal with the blocker BZ? Then @cben would cover this later with his VCR specs? Including this data combination example.

cben:
Ok with me, FWIW

moolitayer:
@Ladas why not add a refresh_parser_spec for the case that went wrong here? we use it as general unit tests related to all things parsing, STI and others. let's reiterate on that if needed

Ladas:
@moolitayer looking at the refresh_parser_spec, you use that mainly for making sure the data are parsed correctly. For this case it's better to have VCR spec, testing the scenarios with missing and present image of a pod, leading to correct data. At least that is what we do in other providers. :-) Since @cben will be adding those VCR specs, I think we can leave it on that?

moolitayer:
@Ladas It would be good to test in VCR specs as an integration test for the complete refresh and in refresh_parser_spec as a unit test - we already check several type attributes in that class.
Since that test is simple enough + fails before the PR and succeeds with it I would feel more confident having the test.

We used to have a requirement at some point to have both types of specs for each new entity.

simaishi:

```
$ git log -1
commit e319f3a7d92ab5f9f5ecc5355c442822a3a692ec
Date:   Mon Dec 11 17:36:01 2017 +0200

    Merge pull request #177 from Ladas/make_sure_container_has_always_the_right_sti_type
    
    Make sure Container has always the right STI type
    (cherry picked from commit cc3618091bc5027b6ac0fee091555efe059cee86)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1524630
```

cben:
@Ladas, does this need a data migration?
I just simulated via `Container.last.update(type: "Container")` and then refresh — it does NOT fix the `type` column.  Neither :batch strategy graph nor classic refresh.
(if I destroy the Container and refresh it re-creates correctly of course)

P.S. found stable way to reproduce missing containerStatuses — a Pending pod — https://bugzilla.redhat.com/show_bug.cgi?id=1524630#c9

Ladas:
@cben yeah. We do not update :type in refresh, for some reason. At this point, we should document a rails c fix, since it's too late for a migration.

PR #159, opened 2017-11-06
Improve Hawkular metrics collection

yaacov:
**Description**

Current metrics collection from Hawkualr produce metrics that differ from metrics produced by OCP consule ui. The metrics can also be inaccurate, for example cpu usage percent may be more then 100 percent.

This PR make use of the same metrics used by OCP.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1455186

Should also fix: https://bugzilla.redhat.com/show_bug.cgi?id=1517064
- warn if no metric endpoint defined
- log failure only once, currently we do a log line for each failed entitly.

PR #184, opened 2017-12-06
Do not run vcr 3 times

yaacov:
**Description**

Because of rspec context switching `@node` is always `nil` in line 28.

To make the `@targets` _reference_ global it need to be defined in a `:all` context.

Here we declare the `@targets` in the `:all` context, and the check for `@target.empty?` will check the global context as expected 🎆 .


yaacov:
While checking a similar issue in @cben 's PR, we found that here we poll the vcr 3 times, the `@node == nil` was done with new context each time, and was always nil.

@simon3z @moolitayer @cben please review.

p.s.
PR #159 will touch the same file ... so this or #159 will have to be rebased after the other is merged ..

yaacov:
Ahhh, checking with @cben, we must put the `@ems = FactoryGirl.create` in the `:all` context for this to work, o/w `@ems` will not be saved between `context` switching and may or may not be available for the tests ...

closing in favour of #135

moolitayer:
can we put the initialization block `before(:all)`?

cben:
yes, see #185

PR #185, opened 2017-12-06
Do not run vcr 3 times - using before(:all)

cben:
Alternative to #184.  Paired with @yaacov 

The `if @node.nil?` trick we had here before didn't work because instance vars are were reset per-test.

New code runs refresh outside the per-test transactions, relies on `@ems.destroy` being correct for cleanup.
@bdunne and @Ladas said on gitter this might be a bad idea.

Overall this test has 3 specs, so it saves 3 - 1 = 2 refreshes, and since it's default graph refresh :smile:, it's small difference — approximately 4.5 sec instead of 5.5 sec.
Perhaps not worth trouble and just let it do 3 refreshes?

(these were attempt to learn towards bigger gain in refresher_spec but not so simple to apply there)

@miq-bot add-label test, performance

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context_spec.rb
@@ -1,43 +1,47 @@
 describe ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::HawkularCaptureContext do
-  @node = nil
-
-  before(:each) do
-    allow(MiqServer).to receive(:my_zone).and_return("default")
+  before(:all) do
     hostname = 'capture.context.com'
     token = 'theToken'
 
-    @ems = FactoryGirl.create(
-      :ems_kubernetes,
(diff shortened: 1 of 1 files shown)


spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context_spec.rb
```diff
@@ -1,43 +1,47 @@
 describe ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::HawkularCaptureContext do
-  @node = nil
-
-  before(:each) do
-    allow(MiqServer).to receive(:my_zone).and_return("default")
+  before(:all) do
```
bdunne: I understand wanting to improve the performance, but I don't think this is worth it for a savings of ~1 second.

cben:
Thanks for the feedback!  I agree, closing this, will just drop the non-working `if @node.nil?` check.

(dropping old refresh — I'd wait until we release Gaprindashvili, but yes, :scissors: :scissors:)

PR #186, opened 2017-12-06
Remove non-working attempt to do refresh only once

cben:
Existing test doesn't work because instance vars were reset for each test, so we do 3 refreshes anyway.
Replaces #184 and #185 that explored ways to make it work but were deemed not worth it.
Overhead of doing 3 refreshes is only 1 sec (5.5s instead of 4.5s).

@miq-bot add-label test, refactoring

@yaacov please review


moolitayer:
@zeari @enoodle we had a couple of attempts, can one of give a fresh look?

moolitayer:
@cben @yaacov Gaprindashvili?

yaacov:
@moolitayer this is spec ... I do not think we need to backport specs ..., but it will do no harm.

PR #187, opened 2017-12-10
Make capture interval a multiple of 30s

yaacov:
**Description**

TL;DR This PR change query rate from  Hawkular/Prometheus to 60s.

Metrics are scraped from k8s/openshift once every 30s into Hawkular/Prometheus.
Currently we query data from Hawkular/Prometheus every 20s, this create a response with missing data.

After discussion we decided to query from Hawkular/Prometheus every 60s, because it is a multiple of base rate (e.g. 30s) and allow for missing heartbeat once in a while.

Removed from #159 

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1524183

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -18,7 +18,7 @@ module ManageIQ::Providers
     require_nested :HawkularCaptureContext
     require_nested :PrometheusCaptureContext
 
-    INTERVAL = 20.seconds
+    INTERVAL = 60.seconds
 
     VIM_STYLE_COUNTERS = {
       "cpu_usage_rate_average"     => {
@@ -51,6 +51,9 @@ module ManageIQ::Providers
     }
 
     def capture_context(ems, target, start_time, end_time)
+      # make start_time align to minutes
+      start_time = start_time.beginning_of_minute
+
       # check for prometheus endpoint, ems must be set
       if ems.connection_configurations.prometheus.try(:endpoint)
         PrometheusCaptureContext.new(target, start_time, end_time, INTERVAL)
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
@@ -60,10 +60,11 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
 
     def process_cpu_counters_rate(counters_rate)
       @metrics |= ['cpu_usage_rate_average'] unless counters_rate.empty?
-      total_cpu_time = @node_cores * CPU_NANOSECONDS * @interval
+      sec_cpu_time = @node_cores * CPU_NANOSECONDS
       counters_rate.each do |x|
+        interval = (x['end'] - x['start']) / 1.in_milliseconds
         timestamp = Time.at(x['start'] / 1.in_milliseconds).utc
-        avg_usage = (x['avg'] * 100.0) / total_cpu_time
+        avg_usage = (x['min'] * 100.0) / (sec_cpu_time * interval)
         @ts_values[timestamp]['cpu_usage_rate_average'] = avg_usage
       end
     end
@@ -72,7 +73,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
       @metrics |= ['mem_usage_absolute_average'] unless gauges_data.empty?
       gauges_data.each do |x|
         timestamp = Time.at(x['start'] / 1.in_milliseconds).utc
-        avg_usage = (x['avg'] / 1.megabytes) * 100.0 / @node_memory
+        avg_usage = (x['min'] / 1.megabytes) * 100.0 / @node_memory
         @ts_values[timestamp]['mem_usage_absolute_average'] = avg_usage
       end
     end
@@ -80,8 +81,9 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
     def process_net_counters_rate(counters_rate)
       @metrics |= ['net_usage_rate_average'] unless counters_rate.empty?
       counters_rate.each do |x|
+        interval = (x['end'] - x['start']) / 1.in_milliseconds
         timestamp = Time.at(x['start'] / 1.in_milliseconds).utc
-        avg_usage_kb = x['avg'] / (1.kilobyte.to_f * @interval)
+        avg_usage_kb = x['min'] / (1.kilobyte.to_f * interval)
         @ts_values[timestamp]['net_usage_rate_average'] = avg_usage_kb
       end
     end
@@ -98,7 +100,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
           {
             'start' => k,
             'end'   => [sum['end'], n['end']].max,
-            'avg'   => sum['avg'] + n['avg']
+            'min'   => sum['min'] + n['min']
           }
         end
       end
@@ -106,11 +108,20 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
 
     def compute_derivative(counters)
       counters.each_cons(2).map do |prv, n|
-        # Add min, median, max, percentile95th, etc. if needed
+        # Add min, median, min, percentile95th, etc. if needed
+        # time window:
+        # 00:00                                        01:00
+        # ^ (sample start time)                        ^ (next sample start time)
+        # ^ (sample min/max/avg value)                 ^ (next sample min/max/avg value)
+        #       ^ (real sample time)         ^ (real sample time)          ^ (real sample time)
+        #      ^ (real sample value)        ^ (real sample value)           ^ (real sample value)
+        # we use:
+        # (T = start of window timestamp, V = min value of window samples)
+        # because the min value is the value of the sample closest to start of window.
         {
-          'start' => n['start'],
-          'end'   => n['end'],
-          'avg'   => n['avg'] - prv['avg']
+          'start' => prv['start'],
+          'end'   => n['start'],
+          'min'   => n['min'] - prv['min']
         }
       end
     end
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context.rb
@@ -195,7 +195,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
       # insert the raw metrics into the @ts_values global object
       raw_metrics['gauge'][full_key].each do |metric|
         timestamp = Time.at(metric['start'] / 1.in_milliseconds).utc
-        @ts_values[timestamp][key] = metric['avg'] unless metric['empty']
+        @ts_values[timestamp][key] = metric['max'] unless metric['empty']
       end
     end
 
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb
@@ -100,7 +100,8 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
 
         {
           "start" => start_sec.to_i.in_milliseconds,
-          "avg"   => x[1].to_f
+          "end"   => (start_sec.to_i + @interval.to_i).in_milliseconds,
+          "min"   => x[1].to_f
         }
       end
     end
--- a/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context_spec.rb
@@ -70,7 +70,8 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Prom
 
         data = context.collect_metrics
 
-        expect(data.count).to be < 18
+        expect(data.count).to be > 10
+        expect(data.count).to be < 13
       end
     end
   end
--- a/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
@@ -47,69 +47,34 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture do
     # TODO: include also sort_and_normalize in the tests
     METRICS_EXERCISES = [
       {
-        :counters => [
+        :counters           => [
           {
             :args => 'cpu/usage',
             :data => [
-              {'start' => 1446500000000, 'end' => 1446500020000, 'avg' => 0},
-              {'start' => 1446500020000, 'end' => 1446500040000, 'avg' => 4000000000}
+              {'start' => 1_446_500_000_000, 'end' => 1_446_500_060_000, 'min' => 0},
+              {'start' => 1_446_500_060_000, 'end' => 1_446_500_120_000, 'min' => 12_000_000_000},
             ]
           },
           {
             :args => 'network/tx',
             :data => [
-              {'start' => 1446500000000, 'end' => 1446500020000, 'avg' => 0},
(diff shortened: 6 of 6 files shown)


yaacov:
@miq-bot add_label bug, metrics

yaacov:
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1524183

bazulay:
@Ladas  @agrare @zeari 
Are we sure this change does not mess up the rollups ?

cben:
SGTM based on verbal explanation of expected effects, but that doesn't mean much :-)

Would love if we had any tests:
- showing metric rows written to DB?
- integtation test showing a user-visible problem fixed?

yaacov:
@cben 👍 about specs:

> Would love if we had any tests:
> showing metric rows written to DB?

rspec is updated for new interval, it does not check actual rows written to DB, but it does check the methods that write data to the DB, this spec will fail if a different interval is used.

> integtation test showing a user-visible problem fixed?

We do not have integration test framework (?) , you can check this is fixed by running commands in rails console:

Before, we will get 40s, 20s intervals, we have missing interval between 1st and 2nd data points:
```
prov = ExtManagementSystem.find(< ID OF CONTAINER PROVIDER>)
container = prov.containers.find(< ID OF A CONTAINER>)
manager = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture.new(container)
manager.perf_collect_metrics('realtime', 5.minutes.ago, 1.minutes.ago)

  { 2017-12-10 14:29:46 UTC=>{"cpu/usage_rate"=>...},
    2017-12-10 14:30:26 UTC=>{"cpu/usage_rate"=>...},
    2017-12-10 14:30:46 UTC=>{"cpu/usage_rate"=>...},
```
After fix, we will get 60s intervals consistently:
```
prov = ExtManagementSystem.find(< ID OF CONTAINER PROVIDER>)
container = prov.containers.find(< ID OF A CONTAINER>)
manager = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture.new(container)
manager.perf_collect_metrics('realtime', 5.minutes.ago, 1.minutes.ago)

  { 2017-12-10 14:45:24 UTC=>{"cpu/usage_rate"=>...},
    2017-12-10 14:46:24 UTC=>{"cpu/usage_rate"=>...},
    2017-12-10 14:47:24 UTC=>{"cpu/usage_rate"=>...},

```

cben:
What I had in mind is precisely these gaps in metrics that previously resulted from some 20s intervals not containing any 30s-spaced samples.

I think the existing test structure is good, doesn't have to be more "integration", just need more data.
Input data, both before and after, is only 2 samples, and expected output is only 1 timestamp — didn't demonstrate the gaps and doesn't (convincingly) demonstrate we now have no gaps.
- New test input data is 60s-spaced.  Is that what we get in reality when we ask hawkular for 60s resolution?

P.S. tests do some stubbing on HawkularLegacyCaptureContext.  Is that what tests actually run, or HawkularCaptureContext?  Should they test both (or maybe all 3 including prometheus)?

yaacov:
> both before and after, is only 2 samples, and expected output is only 1 timestamp — didn't demonstrate the gaps and doesn't (convincingly) demonstrate we now have no gaps.

This is simulated data to show that we calculate correctly delta T of 60s instead of 20s. We do 
```
\dfrac{\delta V}{\delta T}
``` 
We need to check this calculation, because and \delta T equals the interval (even if we have no data), so if we have missing values we have \delta X == 0, but \delta T still equals the interval, witch results in wrong diff value, more simulated data will not change the test of the calculation.

> New test input data is 60s-spaced. Is that what we get in reality when we ask hawkular for 60s resolution?

Yes, you can run the example code from https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/187#issuecomment-350558438 on a Hawkular or Prometheus provider, we also have VCR recording of the `inawkularCaptureContext` rspec [1] that show no gaps with 60s interval.

> tests do some stubbing on HawkularLegacyCaptureContext. Is that what tests actually run, or HawkularCaptureContext? Should they test both (or maybe all 3 including prometheus)?

No, only HawkularLegacyCaptureContext does the calculation, the new Hawkular and Prometheus code get the diff values pre-calculated, so they are immune to the missing data calculation error.

[1] The rspec in the metrics capture sub classes [1] check that we do not have gaps:
Current VCR recording (after #159) use 60s interval, and check that we have ~10 samples on a 10m time span.
https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context_spec.rb#L91

Ladas:
@yaacov can we align the data samples to be always at beginning of the minute? So 12:20:00, 12:21:00, etc. Since there is aligning code that will store them at closest 20s and that might lead to different values, that could theoretically break the rollup.

@agrare @Fryguy I haven't found a code that would require for data to be 20s aligned, have I missed anything? (OpenShift provider has been breaking this condition for a while)  But there is code relying on the alignment, like when we put multiple metrics into one sample (cpu, mem, net) and when we sum samples into 1 value (network inbound+outbound).

So from what I see, alignment to 00s, should serve us the same as 20s alignment.

---------------

Also, using rates, we should not have a problem with accuracy. But using the old way, I think we need to calculate with 'max' sample, instead of 'avg'. Otherwise the values could be again bigger than 100% for e.g. 1CPU core (as a visible bug in data)

yaacov:
> can we align the data samples to be always at beginning of the minute? 

@Ladas sure :-) done 

example:
```
reload!; c = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture.new(ExtManagementSystem.find(89).containers.last)
c.perf_collect_metrics('realtime', 5.minutes.ago, 1.minutes.ago)
```
```
{2017-12-11 08:09:00 UTC=>{"cpu/usage_rate"=>0.0, "memory/usage"=>15003648.0, "cpu_usage_rate_average"=>0.0, "mem_usage_absolute_average"=>0.07717528933696827},
    2017-12-11 08:10:00 UTC=>{"cpu/usage_rate"=>0.0, "memory/usage"=>15003648.0, "cpu_usage_rate_average"=>0.0, "mem_usage_absolute_average"=>0.07717528933696827},
    2017-12-11 08:11:00 UTC=>{"cpu/usage_rate"=>0.0, "memory/usage"=>15003648.0, "cpu_usage_rate_average"=>0.0, "mem_usage_absolute_average"=>0.07717528933696827},
    2017-12-11 08:12:00 UTC=>{"cpu/usage_rate"=>0.0, "memory/usage"=>15003648.0, "cpu_usage_rate_average"=>0.0, "mem_usage_absolute_average"=>0.07717528933696827},
    2017-12-11 08:13:00 UTC=>{"cpu/usage_rate"=>0.0, "memory/usage"=>15003648.0, "cpu_usage_rate_average"=>0.0, "mem_usage_absolute_average"=>0.07717528933696827},
    2017-12-11 08:14:00 UTC=>{"cpu/usage_rate"=>0.0, "memory/usage"=>14995456.0, "cpu_usage_rate_average"=>0.0, "mem_usage_absolute_average"=>0.07713315158685252}
```

yaacov:
> But using the old way, I think we need to calculate with 'max' sample, instead of 'avg'.

This is not realy a problem as the avg == min == max when we have one sample :-)
But you are right, fixing 👍

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -51,6 +51,9 @@ def log_severity
     }
 
     def capture_context(ems, target, start_time, end_time)
+      # make start_time align with INTERVAL
+      start_time = (start_time.to_i / INTERVAL.to_i) * INTERVAL.to_i
```
Ladas: will this result in the same?
```start_time = start_time.beginning_of_minute```

looks more readable

Comment (yaacov) [reply app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb]: @Ladas nice !

will fix specs and then use this 👍

File: app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb [hunk 2]
```diff
@@ -110,7 +110,7 @@ def compute_derivative(counters)
         {
           'start' => n['start'],
           'end'   => n['end'],
-          'avg'   => n['avg'] - prv['avg']
+          'max'   => n['max'] - prv['max']
```
Ladas: is n['start'] the same as prv['end'] ?

yaacov: not if we are missing a window ... but usually yes

Ladas: right, so it it safer to use 

```'start' => prv['end']``` ?

although, we do not use this information in the calculation. Should we rather use that?

```x['max'] / (1.kilobyte.to_f * (n['end'] - prv['end']))``` here, instead of using @interval above?

Comment (yaacov) [reply app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb]: Ladas 👍 nice catch, I saw that a couple of times, didn't fix that immediately, and forgot about that :-)

File: app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb [hunk 3]
```diff
@@ -72,16 +73,17 @@ def process_mem_gauges_data(gauges_data)
...
-        avg_usage = (x['avg'] / 1.megabytes) * 100.0 / @node_memory
+        avg_usage = (x['max'] / 1.megabytes) * 100.0 / @node_memory
         @ts_values[timestamp]['mem_usage_absolute_average'] = avg_usage
       end
     end
 
     def process_net_counters_rate(counters_rate)
       @metrics |= ['net_usage_rate_average'] unless counters_rate.empty?
       counters_rate.each do |x|
+        interval = (x['end'] - x['start']) / 1.in_milliseconds
```
Ladas: right, this is true only if ```n['start']``` the same as ```prv['end']```, right?

otherwise we would need to move the calculation to below, where we have ```n``` and ```prv``` elements available?

yaacov: @Ladas 👍 fixed, now x['end'] is defined is next['start'] on line 115.

yaacov:
@miq-bot add_label compute/containers, gaprindashvili/yes

BZ target 5.9

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -108,9 +110,9 @@ def compute_derivative(counters)
       counters.each_cons(2).map do |prv, n|
         # Add min, median, max, percentile95th, etc. if needed
         {
-          'start' => n['start'],
-          'end'   => n['end'],
-          'avg'   => n['avg'] - prv['avg']
+          'start' => prv['start'],
+          'end'   => n['start'],
+          'max'   => n['max'] - prv['max']
```
Ladas: hm, if we do ```<prv['start'], n['start']>``` shouldn't we use n['min'] - prv['min'], since samples are on the start timestamp?

Also, it's still ```'avg'   => n['max'] - prv['max']```, right? Since we are computing avg between 2 samples?

the data look like this, right?
```
prv_start---------------(prv_end, n_start)---------------n_end
prv_min----------------(prv_max, n_min)---------------n_max
```

or I probably the actual min&max samples are not aligned?

 ```
prv_start---------------(prv_end, n_start)---------------n_end
----prv_min--------prv_max------------n_min--------n_max--
```
Can we actually get a real time of each sample start time? Cause if we are using bucket start/end, that would explain why I saw bad values (like more than 100% cpu on 1core machine)

Comment (yaacov) [reply app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb]: > hm, if we do <prv['start'], n['start']> shouldn't we use n['min'] - prv['min'], since samples are on the start timestamp?

Right 👍 will update to ['min'].

> probably the actual min&max samples are not aligned?

Correct, but we are talking about 30s samples over 60s window:
a. we have 1.. 2 samples per window => `min ~= max ~= avg`
b. sample time error max is < 30s
 
> Can we actually get a real time of each sample start time?

Hawkular:
Not without changing too much of the code, we use Hawkular to sort the samples (and it drops realtime info), and if we ask Hawkular to give as raw data, we will have to add the code to do the sorting ourselves.

Prometheus:
No way to know raw data time by design ( max sample time error is 30s :-) )

P.S
This code is obsolete, It is only a fullback in case we can not get the rate already calculated from Hawkular, openshift from v3.5 and above give us the already calculated rate, and this code is not called any more. see #159 .

**Comment by yaacov** [comment]
@Ladas @moolitayer @cben @bazulay 

Points to note when reviewing:
a. This code is only relevant for openshift v3.4 and below, openshift v3.5 and above do not use this code.
b. We do not support v3.4 any more.
c. This code will be removed soon, when we decide to drop support for old openshifts.

**Comment by yaacov** [comment]
@Ladas thanks 👍 

cc @cben @moolitayer

File: app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb [hunk 5]
```diff
@@ -97,20 +99,29 @@ def compute_summation(data)
...
-        # Add min, median, max, percentile95th, etc. if needed
+        # Add min, median, min, percentile95th, etc. if needed
+        # time window:
+        # 00:00                                        01:00
+        # ^ (sample start time)                        ^ (next sample start time)
+        # ^ (sample min/max/avg value)                 ^ (next sample min/max/avg value)
+        #       ^ (real sample time)         ^ (real sample time)          ^ (real sample time)
+        #      ^ (real sample value)        ^ (real sample value)           ^ (real sample value)
+        # we use:
+        # (T = start of window timestamp, V = min value of winodw samples)
```
cben: s/winodw/window

yaacov: replaced only once :-)

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -97,20 +99,29 @@ def compute_summation(data)
           # Add min, median, max, percentile95th, etc. if needed
           {
             'start' => k,
-            'end'   => [sum['end'], n['end']].max,
-            'avg'   => sum['avg'] + n['avg']
+            'end'   => [sum['end'], n['end']].min,
```
cben: this one I'm not sure about — if `end` is supposed to be the last timestamp of the summation period, this should remain max?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
```diff
@@ -97,20 +99,29 @@ def compute_summation(data)
...
+            'min'   => sum['min'] + n['min']
```
cben: this is weird both before and after.
average over many windows is not sum of sub-window averages (should be divided by count)
minimum over many windows is not sum of sub-window minimums
?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
yaacov: r/max/min/g :-)

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
yaacov: r/max/min/g :-) again ...

> this is weird both before and after.

place sad imoji here :-) 

> average over many windows is not sum of sub-window averages (should be divided by count)
minimum over many windows is not sum of sub-window minimums
?

This function just sum the `tx + rx` values to give `net_total`, **no diff** here, so summation is ok here.

cben: ah, that's what I missed, I thought it's summing along time axis.  :+1:

simaishi:

```
$ git log -1
commit 767ed8106f6808a2d83175b9773f191a86893cbd
Date:   Thu Dec 14 14:00:16 2017 +0200

    Merge pull request #187 from yaacov/metric-capture-interval-60s
    
    Make capture interval a multiple of 30s
    (cherry picked from commit 9be0e668b23e822aeaf3eef960fcb3a0f1fa04cb)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1527094
```

PR #192, opened 2017-12-18
Fix alerts SSL validation trusting custom store 

moolitayer:
https://bugzilla.redhat.com/show_bug.cgi?id=1526877

Introduced in 8d6d7b4ee0f7428d45ecbae74abf006b237955f0


moolitayer:
@agrare please review

moolitayer:
To add on @cben input, I think having fewer possibilities / code path to pass arguments would serve us best.

zgalor:
This was roughly based on https://github.com/prometheus/prometheus_api_client_ruby options.
I think the idea was to allow complete headers (:ssl, :proxy, :request,...). 

I can remove these possibilities, if the usage is obsolete, of confusing.

@yaacov Any special reason to keep them?

agrare:
> I'm baffled by the ssl = options[:options]...

This confused me also, it makes sense if you replace that with `ssl = options` though maybe that was the intent?

cben:
I see.  Similar API among gems is good goal.

So seems both gems support 2 APIs: 
- `options: {verify_ssl: ..., ssl_cert_store: ...}` — is this based on any other lib?
- `ssl: {verify_ssl: ..., cert_store: ...}` — matches Faraday.

Can we settle on just 1?  I think absent a global standard, aligning with a one of the major libraries like Faraday is good tie breaker (also less code :wink: but I care more about reducing API fragmentation)

yaacov:
> options: {verify_ssl: ..., ssl_cert_store: ...} — is this based on any other lib?

yes, it is based on the hawkular client:
https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_client_mixin.rb#L33

simaishi:

```
$ git log -1
commit 165a34709d2db1148ea4ba07be0953e4d14d495d
Date:   Mon Dec 18 08:12:58 2017 -0500

    Merge pull request #192 from moolitayer/fix_alerts_custom_ssl
    
    Fix alerts SSL validation trusting custom store
    (cherry picked from commit d697e6cd74f5a778e8fd2bf183f62458aeb8bc5e)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1527096
```

PR #194, opened 2017-12-27
Tests for parse_pod, specifically for STI type if missing pod status

cben:
Followup to #177.  As discussed there, testing actual `type` set in DB in refresher_spec might be best, but it turned out harder than I thought to capture pods without status in VCR.  A pod with bad image does get a status, with `waiting` state.  (Maybe requesting resources above quota to keep it pending would work?)

Instead, I captured various states a pod goes through from `oc get pods --watch --show-all --all-namespaces -o json` into refresh_parser_spec.
(Tip: `jq 'select((.spec.containers | length) > (.status.containerStatuses | length))'` finds states with missing statuses.)

We already have some tests for parse_container_status/state, but the bug here

Wasn't sure what to test, for now included all states but only tested STI types.

Verified test failed before #177:
```
parse_pod
  sets correct STI types for container_creating_pod
  sets correct STI types for just_created_pod (FAILED - 1)
  sets correct STI types for graceful_deletion_pod
  sets correct STI types for deleted_pod
  sets correct STI types for scheduled_pod (FAILED - 2)
  sets correct STI types for err_pod
```
https://bugzilla.redhat.com/show_bug.cgi?id=1517676
@miq-bot add-labels inventory, test, gaprindashvili/yes

@zeari @enoodle @moolitayer please review
(not cc'ing Ladas because holidays)

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
@@ -210,6 +210,226 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser do
     end
   end
 
+  describe "parse_pod" do
+    # Several stages through which a pod goes, recorded from `oc get pods --watch --show-all -o json`.
+    pod_common = {
+      :apiVersion => "v1",
+      :kind       => "Pod",
+      :metadata   => {
+        :annotations       => {
+          :"openshift.io/scc" => "privileged"
+        },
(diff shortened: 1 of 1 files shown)


simaishi:

```
$ git log -1
commit 71cefb8eef9eac57a705826d3d77148542e2c407
Date:   Mon Jan 8 11:13:17 2018 +0200

    Merge pull request #194 from cben/test_parse_pod
    
    Tests for parse_pod, specifically for STI type if missing pod status
    (cherry picked from commit 2df4967cba252e47c50c5388944f0953b1a6e667)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1524630
```

PR #195, opened 2017-12-27
Use prometheus client instead of faraday

yaacov:
**Description**

We want to leverage the new Prometheus API gem, for metrics collection.

BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1468251 - Use Prometheus Metrics Ruby client

Note: this is the second part of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/132

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
@@ -1,5 +1,5 @@
 module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::PrometheusClientMixin
-  require 'faraday'
+  require 'prometheus/api_client'
 
   def prometheus_client
     @prometheus_uri ||= prometheus_uri
@@ -10,22 +10,10 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Promet
   end
 
   def prometheus_client_new(uri, credentials, options)
-    worker_class = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCollectorWorker
-
-    Faraday.new(
-      :url     => uri.to_s,
-      :proxy   => options[:http_proxy_uri].empty? ? nil : options[:http_proxy_uri],
-      :ssl     => {
-        :verify     => options[:verify_ssl] != OpenSSL::SSL::VERIFY_NONE,
-        :cert_store => options[:ssl_cert_store]
-      },
-      :request => {
-        :open_timeout => worker_class.worker_settings[:prometheus_open_timeout] || 5,
-        :timeout      => worker_class.worker_settings[:prometheus_request_timeout] || 30
-      },
-      :headers => {
-        :Authorization => "Bearer " + credentials[:token]
-      }
+    Prometheus::ApiClient.client(
+      :url         => uri.to_s,
+      :options     => options,
+      :credentials => credentials
     )
   end
 
@@ -37,7 +25,6 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::Promet
     URI::HTTPS.build(
       :host => prometheus_endpoint.hostname,
       :port => prometheus_endpoint.port,
-      :path => "/api/v1/"
     )
   end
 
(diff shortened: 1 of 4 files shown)


yaacov:
@miq-bot add_label bug, metrics, gaprindashvili/yes

BZ traget 5.9

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -10,22 +10,10 @@ def prometheus_client
...
-    Faraday.new(
-      :url     => uri.to_s,
-      :proxy   => options[:http_proxy_uri].empty? ? nil : options[:http_proxy_uri],
-      :ssl     => {
-        :verify     => options[:verify_ssl] != OpenSSL::SSL::VERIFY_NONE,
-        :cert_store => options[:ssl_cert_store]
-      },
-      :request => {
-        :open_timeout => worker_class.worker_settings[:prometheus_open_timeout] || 5,
-        :timeout      => worker_class.worker_settings[:prometheus_request_timeout] || 30
```
enoodle: It seems like those timeouts are not  sent to the prometheus client anymore, are we OK with that?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
```diff
@@ -59,17 +50,11 @@ def labels_to_s(labels, job = "kubernetes-cadvisor")
...
-      response = prometheus_client.get("query", :query => "ALL")
+      data = prometheus_client.query(:query => "ALL")
     rescue StandardError => err
       raise MiqException::MiqUnreachableError, err.message, err.backtrace
     end
 
-    begin
-      data = JSON.parse(response.body)
-    rescue StandardError => err # if auth_proxy fail it returns an html doc
-      raise MiqException::MiqInvalidCredentialsError, err.message, err.backtrace
```
enoodle: Where is the credential error handled now? If it is in the client then line 55 will turn it intu `MiqUnreachableError`

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
enoodle: Ok I see it in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/195/files#diff-8ade5cbaa5f6435425cdbd87ed7d5a96R42

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb
yaacov: 👍 nice catch, now that we have a specialized client we can leverage it's error's to get better differentiations between specific types of errors.

Thanks @enoodle , will remove the code that hide the real error.

moolitayer:
@yaacov FYI while reviewing this PR I noticed that we do not distinguish between different errors from this endpoint. ssl, unauthorized or non existing route all look the same. See for example ssl error against OpenShift and Prometheus:





This was not introduced here, we will want to improve that going forward

cben:
Indeed.  The UI just show all failures the same:
https://github.com/ManageIQ/manageiq-ui-classic/blob/629f2d6/app/controllers/mixins/ems_common_angular.rb#L99-L100
There is some exception classification in AuthenticationMixin:
https://github.com/ManageIQ/manageiq/blob/ed4af9d204/app/models/mixins/authentication_mixin.rb#L348-L361
but it doens't separate out things like "we don't trust the server" from "server doesn't trust us".

What do you think better UX would look like?
And please open new issue if you want us not to forget this.

bazulay:
@moolitayer  please open a github issue based on the 2 previous comments

moolitayer:
https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/200

simaishi:

```
$ git log -1
commit 6fe43de2c8495d3502d019a29fe6d3b9eb256df6
Date:   Wed Dec 27 18:00:01 2017 +0200

    Merge pull request #195 from yaacov/use-prometheus-client
    
    Use prometheus client instead of faraday
    (cherry picked from commit 37b7884b16ea879764bac680f8bd2b60c3c9c3a3)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1530736
```

PR #202, opened 2018-01-04
We need to use existing relation to project so we build valid query

Ladas:
Depends on:

- [x] https://github.com/ManageIQ/manageiq/pull/16741

We need to use existing relation to project so we build valid query,
otherwise the targeted specs fails.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -150,7 +150,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :parent                 => manager,
           :builder_params         => {:ems_id => manager.id},
           :association            => :container_groups,
-          :secondary_refs         => {:by_namespace_and_name => [:namespace, :name]},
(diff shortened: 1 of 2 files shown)


Ladas:
@cben the core dependency brings local db index loaders for secondary indexes. But that fails the current targeted refresh spec, because we are using blacklisted 'namespace'. I will add more validations so the Graph refresh shouts when we add invalid combination of columns for the index.

cben:
Great!  We need to move in this direction.
Are you planning to convert all `:by_namespace_and_name` usages?

I think when done we can stop passing :namespace to inv collections (and blacklisting it):
```diff
h[:container_project] = lazy_find_project(:name => h[:namespace])
h[:container_project] = lazy_find_project(:name => h.delete(:namespace))
```

Ladas:
@cben yes will do, I will first write asserts, that will tell me what to do :-D

Ladas:
@cben ok, these are all occurrences that the new asserts in core shows, sending also OpenShift PR. Could you then remove the actual namespace as followup? I don't want to mess that up. :-)

cben:
@miq-bot add-label gaprindashvili/no
(going over unlabeled PRs...  assuming no as this depends on several unbackported enhancement core PRs with like 40 commits.)

PR #205, opened 2018-01-07
Cast quota values to float in parser, to match float columns

cben:
This works around bug in classic refresh where casting didn't happen,
so we can proceed with container quota archiving independently of
https://github.com/ManageIQ/manageiq/pull/16723
This way backporting of that fix can be decided independently.

https://bugzilla.redhat.com/show_bug.cgi?id=1504560

@miq-bot add-label enhancement, gaprindashvili/yes

@yaacov @zeari @moolitayer Please review.
Didn't add tests, because not sure if I should test anything here,
but will have tests for no unnecessary archiving in the archiving PR.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -934,15 +934,15 @@ module ManageIQ::Providers::Kubernetes
       end
 
       resource_quota.spec.hard.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_desired] = parse_quantity(quota)
+        new_result_h[resource_name][:quota_desired] = parse_quantity(quota).to_f
       end
 
       resource_quota.status.hard.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_enforced] = parse_quantity(quota)
+        new_result_h[resource_name][:quota_enforced] = parse_quantity(quota).to_f
       end
 
       resource_quota.status.used.to_h.each do |resource_name, quota|
-        new_result_h[resource_name][:quota_observed] = parse_quantity(quota)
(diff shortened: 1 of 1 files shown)


zeari:
> This works around bug in classic refresh where casting didn't happen,

Why didnt the casting happen?

zeari:
Do we need the same change for other places that have `parse_quantity`?

cben:
> Why didn't the casting happen?

@zeari see ManageIQ/manageiq#16723, we were using a Rails API wrongly :man_facepalming: 

> Do we need the same change for other places that have parse_quantity?

Not yet, and probably not.
The bug only affects columns used in find_key, until now nobody was crazy enough to identify records with a float column ;-), but it's likely there were some int/string mismatches.
I don't know yet what, if anything, was broken.  If anything was broken, we should backport ManageIQ/manageiq#16723 to fix it — and won't need workarounds like this.  But that fix *may* affect refresh of all columns in all providers, so we should not backport it lightly...

At this stage, I only want to decouple the *new feature* of quota archiving from decision to backport, or not, ManageIQ/manageiq#16723.

(I'm not sure yet if graph refresh had similar bug)

moolitayer:
@cben isn't this a `bug` rather then `enhancement`?

cben:
It's working around a core bug, but in a place where it doesnt affect any
existing container functionality, only needed for the quotes history
enhancement :)
Feel free to relabel.

simaishi:

```
$ git log -1
commit c5f5049810e6b6435dff181c52983569f491b72e
Date:   Tue Jan 9 11:37:20 2018 +0200

    Merge pull request #205 from cben/parse_resource_quota_items-to_f
    
    Cast quota values to float in parser, to match float columns
    (cherry picked from commit bd1eeb05697f422f32a45021686af6a0343f520e)
```

PR #206, opened 2018-01-07
Add a queue metrics capture method to container manager

yaacov:
**Description**
Add a method to queue reading of metrics.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1527877
Target: 5.9.1

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -266,4 +266,17 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     # and these should automatically be translated to alerts.
     true
   end
+
+  def queue_metrics_capture
+    targets = Metric::Targets.capture_container_targets([self], {})
+
+    targets.each do |target|
+      begin
+        target.perf_capture_queue('realtime', :priority => MiqQueue::HIGH_PRIORITY)
+      rescue StandardError => err
+        _log.error("Failed to queue perf_capture for target [#{target.class.name}], [#{target.id}], [#{target.name}]: #{err}")
+        raise
+      end
+    end
+  end
 end


yaacov:
@miq-bot add_label bug, metrics, gaprindashvili/yes

@cben @moolitayer @zeari @Ladas please review

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -266,4 +266,16 @@ def evaluate_alert(_alert_id, _event)
...
   end
+
+  def queue_metrics_capture(interval_name = 'realtime', options = {})
+    targets = Metric::Targets.capture_container_targets([self], {})
+
+    targets.each do |target|
+      begin
+        target.perf_capture_queue(interval_name, options)
+      rescue StandardError => err
+        _log.warn("Failed to queue perf_capture for target [#{target.class.name}], [#{target.id}], [#{target.name}]: #{err}")
```
cben: :+1: [perf_capture_queue](https://github.com/ManageIQ/manageiq/blob/ea927c7205/app/models/metric/ci_mixin/capture.rb#L32) may raise exceptions, especially ArgumentError.
However, I expect in most cases where one target raises, all would raise,
so perhaps this function should just let exception propogate?
[Your call, I dont have opinion here]

Ladas: yeah we should let it propagate and rather do _log.error, there shouldn't be a case where only 1 fails

yaacov: thanks @cben @Ladas 👍

now it's `_log.error` and propagating :cherry_blossom:

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -266,4 +266,17 @@ def evaluate_alert(_alert_id, _event)
     # and these should automatically be translated to alerts.
     true
   end
+
+  def queue_metrics_capture(interval_name = 'realtime', options = {})
```
moolitayer: Since this is meant for queuing in the UI I would prefer if you remove both parameters

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -266,4 +266,17 @@ def evaluate_alert(_alert_id, _event)
...
+    targets = Metric::Targets.capture_container_targets([self], {})
+
+    targets.each do |target|
+      begin
+        target.perf_capture_queue(interval_name, options)
```
moolitayer: See `Metric::CiMixin::Capture.perf_capture_realtime_now`
can you use that here?

moolitayer:
Update: yaacov and myself observed we have some errors in the log post queue_metrics_capture calls.
Seems to be related to subsequent requests where the calculated gap from last refresh is too small.

We also observed some cases where the end_time - start_time for metrics collection is very large (`PrometheusCaptureContext.fetch_counters_data`) but it's not clear if that is a problem OR related to this PR.

yaacov:
> Update: yaacov and myself observed we have some errors in the log post queue_metrics_capture calls.

This errors have no connection to this PR.

> Seems to be related to subsequent requests where the calculated gap from last refresh is too small.

No, the are related to `Erorr/CrashLoop` pods/containers that have no metric data for the time slot requested.

> We also observed some cases where the end_time - start_time for metrics collection is very large

Regular 'historical' requsts ask fot different time slots depending on gaps and missing data, it's ok to have strange :-) time slots, when queue by system worker.

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -266,4 +266,17 @@ def evaluate_alert(_alert_id, _event)
...
     true
   end
+
+  def queue_metrics_capture
+    targets = Metric::Targets.capture_container_targets([self], {})
+
+    targets.each do |target|
+      begin
+        target.perf_capture_queue('realtime', :priority => MiqQueue::HIGH_PRIORITY)
+      rescue StandardError => err
```
cben: [cosmetic, LGTM either way] since you now re-raise, you could move this rescue to top level of the function, saving begin..end.

yaacov: 👍

yaacov: Ahh, I want to have the target that raised the error, If I move the rescue to top level I lose this info :-(

cben: ah, right.  remind me to resist making completely non-important code comments :)

simaishi:
Removed `gaprindashvili/yes` as the BZ was clsoed as not a bug. Do we need to revert this from `master` branch? If so, please create a PR.

PR #212, opened 2018-01-10
Align manager_ref with unique indexes

Ladas:
Align manager_ref with unique indexes PR https://github.com/ManageIQ/manageiq-schema/pull/19

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
@@ -195,8 +195,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :model_class => ContainerEnvVar,
           :parent      => manager,
           :association => :container_env_vars,
-          # TODO: old save matches on all :name, :value, :field_path - does this matter?
-          :manager_ref => [:container, :name],
+          :manager_ref => [:container, :name, :value, :field_path],
         )
       )
     @collections[:security_contexts]      =
@@ -244,7 +243,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
           :model_class => ContainerServicePortConfig,
           :parent      => manager,
           :association => :container_service_port_configs,
-          :manager_ref => [:ems_ref, :protocol] # TODO(lsmola) make protocol part of the ems_ref?
+          :manager_ref => [:container_service, :name]
         )
       )
 
@@ -317,8 +316,9 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::InventoryCollections
         shared_options.merge(
           :model_class    => PersistentVolume,
           :parent         => manager,
-          :builder_params => {:parent => manager},
+          :builder_params => {:parent_id => manager.id, :parent_type => manager.class.base_class},
           :association    => :persistent_volumes,
+          :manager_ref    => [:parent_id, :parent_type, :ems_ref],
         )
       )
     @collections[:persistent_volume_claims] =


app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -244,7 +243,7 @@ def initialize_inventory_collections
           :model_class => ContainerServicePortConfig,
           :parent      => manager,
           :association => :container_service_port_configs,
-          :manager_ref => [:ems_ref, :protocol] # TODO(lsmola) make protocol part of the ems_ref?
+          :manager_ref => [:container_service, :name]
```
Ladas: @cben is this a valid replacement? We do unique index like this

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -317,8 +316,9 @@ def initialize_inventory_collections
         shared_options.merge(
           :model_class    => PersistentVolume,
           :parent         => manager,
-          :builder_params => {:parent => manager},
+          :builder_params => {:parent_id => manager.id, :parent_type => manager.class.base_class},
```
Ladas: @cben active record object that is part of the manager_ref can act weird now

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
```diff
@@ -195,8 +195,7 @@ def initialize_inventory_collections
           :model_class => ContainerEnvVar,
           :parent      => manager,
           :association => :container_env_vars,
-          # TODO: old save matches on all :name, :value, :field_path - does this matter?
-          :manager_ref => [:container, :name],
+          :manager_ref => [:container, :name, :value, :field_path],
```
Ladas: @cben we have unique index like this, should we reduce the list?

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
Ladas: hm, no this is not enough ```DETAIL:  Key (container_service_id, name)=(174, dns) already exists.```

Ladas: wait,the graph refresh should not allow this, I am doing something wrong

Ladas: disregard the other comments, this passed, I had to update old saving code :-)

cben: you mean "name is not unique" [comment](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/168#issuecomment-356729210)?  it is unique within a service?

Ladas: no disregard that, it was too late :-) I was missing the PR for old refresh. Now this passes all specs, ao it should be probably good :-)

app/models/manageiq/providers/kubernetes/container_manager/inventory_collections.rb
cben: hmm.  my instinct is to get rid of value and field_path (which is an [alternative to value](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-container-fields-as-values-for-environment-variables)).
But I need to verify you can't define 2 vars with same name.

Ladas: Ok, let me know then.

PR #217, opened 2018-01-15
Change alert definition meta

moolitayer:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1536087

Match changes in https://github.com/openshift/origin/pull/17608/files

### DIFF
--- a/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
@@ -49,14 +49,14 @@ module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerM
     #
     # {
     #     "annotations": {
-    #         "message": "Node ocp-compute01.10.35.48.236.nip.io is down",
-    #         "severity": "HIGH",
+    #         "description": "Node ocp-compute01.10.35.48.236.nip.io is down",
     #         "source": "ManageIQ",
     #         "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
     #     },
     #     "endsAt": "0001-01-01T00:00:00Z",
     #     "generatorURL": "http://prometheus-4018548653-w3str:9090/graph?g0.expr=container_fs_usage_bytes%7Bcontainer_name%3D%22%22%2Cdevice%3D%22%2Fdev%2Fmapper%2Fvg0-lv_root%22%7D+%3E+4e%2B07&g0.tab=0",
     #     "labels": {
+    #         "severity": "error",
     #         "alertname": "Node down",
     #         "beta_kubernetes_io_arch": "amd64",
     #         "beta_kubernetes_io_os": "linux",
@@ -76,9 +76,10 @@ module ManageIQ::Providers::Kubernetes::MonitoringManager::EventCatcher::RunnerM
     event = event.dup
 
     annotations = event["annotations"]
-    event[:url] = annotations["url"]
-    event[:severity] = parse_severity(annotations["severity"])
     labels = event["labels"]
+
+    event[:url] = annotations["url"]
+    event[:severity] = parse_severity(labels["severity"])
     # TODO(mtayer): remove after https://github.com/ManageIQ/manageiq/pull/16339
     event[:ems_ref] = incident_identifier(labels, annotations, event["startsAt"])
     event[:resolved] = event["status"] == "resolved"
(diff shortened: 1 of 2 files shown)


moolitayer:
@cben @zeari please review

moolitayer:
Once this merges two places need update:
- https://github.com/ManageIQ/manageiq/issues/14238
- https://github.com/container-mgmt/cm-ops-flow/blob/master/README.md

aweiteka:
> Match changes in https://github.com/openshift/origin/pull/17608/files

I will push a change to https://github.com/openshift/origin/pull/17608/ to match labels convention "info|warning|error"

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -77,7 +77,7 @@ def extract_event_data(event)
 
     annotations = event["annotations"]
     event[:url] = annotations["url"]
-    event[:severity] = parse_severity(annotations["severity"])
+    event[:severity] = parse_severity(labels["severity"])
```
cben: does this line work?  it seems labels is only set on next line?

cben:
What happens if you have old rules?
Once you switch to new rules, can you still have old alerts in queue?
Looks like code won't crash, at worst message will be nil & severity will default to "error". :+1:

moolitayer:
Thanks @cben, I added some tests.

I don't think there are backward compatibility issues. I'm aiming to have this in the first released version of Prometheus alerts.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -76,17 +76,18 @@ def extract_event_data(event)
...
     # TODO(mtayer): remove after https://github.com/ManageIQ/manageiq/pull/16339
     event[:ems_ref] = incident_identifier(labels, annotations, event["startsAt"])
     event[:resolved] = event["status"] == "resolved"
     {
       :source     => "DATAWAREHOUSE",
       :timestamp  => Time.zone.now,
       :event_type => "datawarehouse_alert",
-      :message    => annotations["message"],
+      :message    => annotations["description"],
```
aweiteka: We're settling on a convention where 'summary' is the short string (general data) and 'description' is the more verbose, detailed string (with specific data). For example:

```
alert: DockerLatencyHigh
expr: round(max(kubelet_docker_operations_latency_microseconds{quantile="0.9"})
  BY (instance) / 1e+06, 0.1) > 1
for: 5m
labels:
  severity: warning
annotations:
  summary: Docker latency is high
  description: Docker latency is {{ $value }} seconds for 90% of kubelet operations
```

As an operator, the more compact 'summary' is what I would want displayed inline, expanding to display 'description' if I want more details.

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
```diff
@@ -49,14 +49,14 @@ def extract_event_data(event)
...
-    #         "message": "Node ocp-compute01.10.35.48.236.nip.io is down",
-    #         "severity": "HIGH",
+    #         "description": "Node ocp-compute01.10.35.48.236.nip.io is down",
     #         "source": "ManageIQ",
     #         "url": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
     #     },
     #     "endsAt": "0001-01-01T00:00:00Z",
     #     "generatorURL": "http://prometheus-4018548653-w3str:9090/graph?g0.expr=container_fs_usage_bytes%7Bcontainer_name%3D%22%22%2Cdevice%3D%22%2Fdev%2Fmapper%2Fvg0-lv_root%22%7D+%3E+4e%2B07&g0.tab=0",
     #     "labels": {
+    #         "severity": "HIGH",
```
aweiteka: Shouldn't this be one of info|warning|error?

app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb
moolitayer: Right now the screen is designed to have only one type of message.
And although we persist both we pass only the long name to the alert and show in the screen.

Would you like to open a bug for this change?

moolitayer:
> (does this depend on openshift/origin#17608?)

No, it is does not, while this is only about agreeing on the metadata it is important that we release our first version of this based on that agreement. Ready for merge.

moolitayer:
Updated docs:
- https://github.com/container-mgmt/cm-ops-flow/pull/13/files
- https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/217
- https://github.com/ManageIQ/manageiq/issues/14238

simaishi:

```
$ git log -1
commit c53d253a77309b4443d93816a504fb0e6b1f5b5e
Date:   Sun Jan 21 21:38:26 2018 +0200

    Merge pull request #217 from moolitayer/change_alert_meta
    
    Change alert definition meta
    (cherry picked from commit 700ec3fa6ca9dda1f8def99328481d6ae034f45c)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1539067
```

PR #218, opened 2018-01-18
Migrate model display names from locale/en.yml to plugin

mzazrivec:
This is a continuation of the work started in https://github.com/ManageIQ/manageiq/pull/16596 where we want to migrate display model names from locale/en.yml into particular models, where we'll use standard gettext. Main goal here is to make this one aspect of our application pluggable.

Core PR: https://github.com/ManageIQ/manageiq/pull/16836

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_group.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_group.rb
@@ -1,3 +1,7 @@
 class ManageIQ::Providers::Kubernetes::ContainerManager::ContainerGroup < ::ContainerGroup
   alias_attribute :pod_uid, :ems_ref
+
+  def self.display_name(number = 1)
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/kubernetes/container_manager/container_group.rb
```diff
@@ -1,3 +1,7 @@
 class ManageIQ::Providers::Kubernetes::ContainerManager::ContainerGroup < ::ContainerGroup
   alias_attribute :pod_uid, :ems_ref
+
+  def self.display_name(number = 1)
+    n_('Pod (Kubernetes)', 'Pods (Kubernetes)', number)
```
moolitayer: Since we have only kubernetes pods, I think this should keep being just 'Pod' in the screens

PR #221, opened 2018-01-21
[GAPRINDASHVILI] Final translations for gaprindashvili

mzazrivec:
This PR contains final translation fixes / polishing for Gaprindashvili GA

### DIFF
--- a/locale/fr/ManageIQ_Providers_Kubernetes.po
+++ b/locale/fr/ManageIQ_Providers_Kubernetes.po
@@ -1,12 +1,12 @@
 # Laurence Ganseman <laurence.ganseman@stoquart.com>, 2017. #zanata
-# Nathalie Stoquart <nateltrad@gmail.com>, 2017. #zanata
+# croe <croe@redhat.com>, 2017. #zanata
 msgid ""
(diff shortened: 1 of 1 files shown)


cben:
@simaishi sorry, merged this to gaprindashvili without thinking that should check with you, and also forgot we're in blockers only :-|
@mzazrivec you say on https://github.com/ManageIQ/manageiq/pull/16860 it "should be merged with plugin PRs" — is there an order dependence?
@mzazrivec is there a BZ or are translations exempt from blocker requirements?

mzazrivec:
@cben There's no order dependency. There's no bug that I know of, I just got an email from translators saying they polished / corrected some strings for the gaprindashvili release, so here they are.

moolitayer:
:+1:

moolitayer:
Feels like this should have been a master + backport fix (why not have this fix on master as well)

mzazrivec:
In general, master catalogs differ from whatever is in gaprindashvili, so a plain cherry pick / backport wouldn't work. And even if the backport would work, there's still a big chance that the backported catalogs won't be complete.

PR #223, opened 2018-01-23
Rollup specs to show our current weak points

Ladas:
Rollup specs to show our current weak points and that are related
to https://bugzilla.redhat.com/show_bug.cgi?id=1506671

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
@@ -123,7 +123,10 @@ shared_examples "kubernetes rollup tests" do
     Timecop.return
   end
 
-  def add_metrics_for(resource, range, metric_params: {}, step: 20.seconds)
+  # TODO(lsmola) with 60s interval, we need also a spec for the fact that there can be 2 samples. Then if we take
+  # pod1 using 100 cores only in <12:00:00, 12:00:30) and pod2 using 100 cores only in <12:00:30, 12:01:00). The
+  # project avg should be 100 cores, not 200 cores.
+  def add_metrics_for(resource, range, metric_params: {}, step: 60.seconds)
     range.step_value(step).each do |time|
       metric_params[:timestamp]           = time
       metric_params[:resource_id]         = resource.id
--- a/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
@@ -117,10 +117,10 @@ shared_examples "kubernetes rollup tests" do
     end
   end
 
-  def rollup_up_to_project
+  def rollup_up_to_project(rollup_end_time = start_time + 2.hours)
     # Queue hourly rollups for all ContainerGroups
     ContainerGroup.all.each.each do |resource|
-      resource.perf_rollup_to_parents('realtime', start_time, start_time + 2.hours)
+      resource.perf_rollup_to_parents('realtime', start_time, rollup_end_time)
     end
 
     # ContainerGroup hourly rollups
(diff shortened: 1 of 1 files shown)


cben:
Will take me some time to review but just wanted to say these will be awesome to have :heart:

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+  end
+
+  let(:container_node_a) do
+    hardware = FactoryGirl.create(:hardware,
+                                  :cpu_total_cores => 10,
+                                  :memory_mb       => 1024)
+
+    node = FactoryGirl.create(:container_node)
+
+    hardware.update_attributes(:computer_system => node.computer_system)
```
cben: can't you create node first, and pass this `create(:hardware, ...)` args?
[i really don't mind, just curious if there is something tricky...]

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+      have_attributes(
+        :derived_memory_used        => 2048.0,
+        :derived_vm_numvcpus        => 12,
+        :derived_memory_available   => nil,
+        :mem_usage_absolute_average => nil
+      )
+    )
+
+    # TODO(lsmola) Described in https://bugzilla.redhat.com/show_bug.cgi?id=1506671
+    pending("We need to store also actual usage, with only 10.minutes running, the usage should be divided by 6")
```
cben: can you extract pending part to separate test case, so other expectations are checked?

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+      :metric_params => {
+        :cpu_usage_rate_average     => 100.0,
+        :mem_usage_absolute_average => 100.0,
+      }
+    )
+
+    # then the pod a is killed and have another add 10 minutes of 100% cpu_util of 2 cores total and and of memory 2048MB
+    add_metrics_for(
+      container_group_b,
+      start_time + 10.minutes..(start_time + 20.minutes),
```
cben: [nit] parens on both sides please
`(start_time + 10.minutes)..(start_time + 20.minutes),`

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+    )
+
+    # TODO(lsmola) Described in https://bugzilla.redhat.com/show_bug.cgi?id=1506671
+    pending("We need to store also actual usage, with only 10.minutes running, the usage should be divided by 6")
+    expect(project_rollup.cpu_usage_rate_average.round(2)).to eq((expected_cpu_util / 6).round(2))
+    expect(project_rollup.derived_memory_used).to eq(2048 / 6.0)
+  end
+
+  it "checks pods running sequentially are not being multiplicated in project rollup" do
+    # Add 10 minutes of 100% cpu_util of 10 cores total and of memory 1024MB
```
cben: consider renaming container_group_a -> container_group_10core_1GB etc.
then you wouldn't need comments to explain the "mystery guest"...

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
Ladas: yeah, best would be to re arrange the factories

yaacov:
@Ladas  👍 looks good, but what about the broken travis ?

Ladas:
@yaacov hm, I need to check the CI failures. Seems like 2 daily rollups were not computed, while it works locally for me.

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+
+  before do
+    Timecop.travel(end_time)
+  end
+
+  after do
+    Timecop.return
+  end
+
+  def add_metrics_for(resource, range, metric_params: {}, step: 20.seconds)
```
yaacov: we use 60.secondes in our collector [1] may be more real life to use the real interval ?

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb#L21

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+    range.step_value(step).each do |time|
+      metric_params[:timestamp]           = time
+      metric_params[:resource_id]         = resource.id
+      metric_params[:resource_name]       = resource.name
+      metric_params[:parent_ems_id]       = ems.id
+      metric_params[:derived_vm_numvcpus] = resource.container_node.hardware.cpu_total_cores
```
yaacov: side note for future work :-)
we have the "realtime" `vm_muncpus` in hawkular and in prometheus:
in hawkular it's `cpu/node_capacity`
in prometheus it's `machine_cpu_cores`
so we can collect it in the collector ... just need not to overwrite it in the rollups :-)

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
...
+      if metric_params[:mem_usage_absolute_average].to_i > 0
+        metric_params[:derived_memory_used]      = (metric_params[:mem_usage_absolute_average] / 100.0) * resource.container_node.hardware.memory_mb
+        metric_params[:derived_memory_available] = resource.container_node.hardware.memory_mb - metric_params[:derived_memory_used]
```
yaacov: same as the cpu's , we have this data in hawkular/prometheus, we do not need to calculate it from hardware.
see comment https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/223/files#r164412708

yaacov:
nice !! we really need this specs, LGTM pending `connection_configurations` notation and `60sec interval`, see comments above.

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
Ladas: right, I am trying to replicate exactly what we do now

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,396 @@
+# instantiated at the end, for both classical and graph refresh
+shared_examples "kubernetes rollup tests" do
+  let(:ems) do
+    allow(MiqServer).to receive(:my_zone).and_return("default")
+    auth = AuthToken.new(:name => "test", :auth_key => "valid-token")
+    FactoryGirl.create(:ems_kubernetes, :hostname => "10.35.0.169",
```
yaacov: can you use the `connection_configurations` notations [1] , we use it in the other specs ?

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context_spec.rb#L7

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
Ladas: right, there will be another spec for this. It's when there are 2 containers, one have sample <12:00:00, 12:00:30) saying e.g. we are using 100 cores, then next 30s is empty. But another container has <12:00:30, 12:01:00) saying it was using 100 cores.

So with the 1.minute precision, the project rollup (even after a rewrite) can say we used avg 200 cores in that minute. So in rare cases, that could cause a report showing data above the quota.

yaacov: ok, but why use `20.seconds` here ?

Ladas: I can turn it to 60s, it's a habit I guess, since we have 20s everywhere else :-)

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
Ladas: fixed

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
Ladas: hm, right now the pending mostly separates it with: after the pending it should be like this. The specs after this will fail

Ladas:
@cben @yaacov some commented, some fixed. Now we should figure out a strategy how to fix these :-D

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
cben: Cool, I thought it affects whole test but TIL that failed expectations before `pending()` do break the test :+1:

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
```diff
@@ -0,0 +1,424 @@
...
+
+    rollup_up_to_project
+    project_rollup = MetricRollup.where(:resource => container_project, :timestamp => start_time, :capture_interval_name => "hourly").first
+
+    # Check the Project rollup has the wanted elements
+    assert_entities(project_rollup)
+
+    # Right now we take the 10 minutes as the whole hour, so 75% of 2 cores, + 50% of 10 cores is 54.17% of 12 cores
+    expected_cpu_util = (75 * 2 + 50 * 10) / 12.0
+    expect(project_rollup.cpu_usage_rate_average.round(2)).to eq(expected_cpu_util.round(2))
```
cben: In edge cases (literally!), x.round(2) might differ from y.round(2) despite x-y being very small.
No problem if these work but generally you want [`be_within` matcher
](https://relishapp.com/rspec/rspec-expectations/v/3-7/docs/built-in-matchers/be-within-matcher)

cben:
backport labels?  If not hard, I'm in favor of adding such tests to released versions too :clap:

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture/rollup_spec.rb
Ladas: @cben is it? I hope that is not possible :-)

cben: x = 0.2051; y = 0.2049; x.round(2) == 0.21; y.round(2) == 0.2; x - y = 0.00020000000000000573

No matter how you split reals into buckets, there always will be close reals that fall on different sides of a "watershed" line between two buckets... 

P.S. You might be wondering, how come then I was recently proposing rounding as solution for quotas equality (before we made the columns decimal)?!
The reason that could have worked is because we knew all values were in vicinity of 0.001 multiples; if we kept division lines in "illegal region" in middle between legal values, rounding could sort them perfectly into buckets.  (This is also the reason quantized "digital" recordings suffering bounded noise can be read back *perfectly*, while to analog recordings noise is irreversible...)

PR #224, opened 2018-01-25
Add support for alert labels

jhernand:
This patch modifies the Kubernetes provider so that it supports alert
labels. These labels will be extracted from the `full_data` stored in
the event that generated the alert.

Fixes https://bugzilla.redhat.com/1520125

Note that this pull request depends on ManageIQ/manageiq#16866,
which adds the `alert_labels` feature to the core.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/alert_labels.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/alert_labels.rb
@@ -1,0 +1,41 @@
+#
+# This mixin implements the `alert_labels` feature.
+#
+module ManageIQ::Providers::Kubernetes::ContainerManager::AlertLabels
+  extend ActiveSupport::Concern
+
+  #
+  # Any provider that includes this mix-in will support the alert lables, populated from the labels
+  # provided by Prometheus.
+  #
+  included do
+    supports :alert_labels
+  end
+
+  #
+  # Returns the alert labels associated to the given alert status.
+  #
+  # @param alert_status [MiqAlertStatus] The alert status object.
+  # @return [Array<MiqAlertStatusLabel>] The array of labels.
+  #
+  def alert_labels(alert_status)
+    # Find the event that correspond to the given alert status:
+    event = EventStream.find_by(
+      :ems_id  => alert_status.ems_id,
+      :ems_ref => alert_status.event_ems_ref
+    )
+    return [] unless event
+
+    # Extract the hash of labels from the event full data:
+    labels = event.full_data && event.full_data['labels']
+    return [] unless labels
+
+    # Convert the hash to an array of label objects:
+    labels.map do |name, value|
+      label = MiqAlertStatusLabel.new
+      label.name = name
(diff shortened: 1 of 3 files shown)


jhernand:
@moolitayer please review.

moolitayer:
@zeari @yaacov please review

app/models/manageiq/providers/kubernetes/container_manager/alert_labels.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  #
+  # Returns the alert labels associated to the given alert status.
+  #
+  # @param alert_status [MiqAlertStatus] The alert status object.
+  # @return [Array<MiqAlertStatusLabel>] The array of labels.
+  #
+  def alert_labels(alert_status)
+    # Find the event that correspond to the given alert status:
+    events = EventStream.where(
```
yaacov: a little nitpick ... AFAIK `obj.where(...).first` works like `obj.find_by(...)` I think `find_by` is more commonly used in this project, any reason to chose the first here ?

jhernand: No, no reason, I will change to use `find_by`, thanks for the suggestion!

jhernand:
@yaacov the new version of the pull requests uses `find_by` instead of `where`, as you suggested.

app/models/manageiq/providers/kubernetes/container_manager/alert_labels.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Find the event that correspond to the given alert status:
+    event = EventStream.find_by(
+      :ems_id  => alert_status.ems_id,
+      :ems_ref => alert_status.event_ems_ref
+    )
+    return [] unless event
+
+    # Extract the hash of labels from the event full data:
+    data = event.full_data
+    return [] unless data
```
yaacov: sorry for nitpicking again, I do not know how readable the code will be after this change, bu if you want to get codeclimate happy :-) , you can do all the sanity chacks in one line:
```
return [] unless event
data = event.full_data
...
lables.map do ...
...

```
is equivalent to:
```
labels = event.full_data['labels'] rescue {}

lables.map do ...
```

yaacov: please disregard last comment: http://www.rubydoc.info/gems/rubocop/0.52.0/RuboCop/Cop/Style/RescueModifier

zeari:
The changes LGTM but the travis failures look relevant

jhernand:
@yaacov @zeari thans for the comments. In the new version of the pull request I have replaced this:

```ruby
data = event.full_data
return [] unless data
labels = data['labels']
return [] unless labels
```

With this:

```ruby
labels = event.full_data && event.full_data['labels']
return [] unless labels
```

I hope that will make _codeclimate_ happy.

jhernand:
We have decided to stop working on this feature.

PR #227, opened 2018-01-30
Do not raise an error when metrics are missing for one object

yaacov:
**Description**

Currently if a pod has no metrics data (for benign or malignant reasons) we:
a. raise an error and log a trace-log.
b. show users in the UI that metrics collection failed.

We want:
a. do not log missing data as an error ( usually this is benign behaviour of a pod not running )
b. do not show users an error message in the UI sammary page.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1530627

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -1,6 +1,7 @@
 module ManageIQ::Providers
   class Kubernetes::ContainerManager::MetricsCapture < BaseManager::MetricsCapture
     class CollectionFailure < RuntimeError; end
+    class NoMetricsFoundError < RuntimeError; end
 
     class TargetValidationError < RuntimeError
       def log_severity
@@ -95,8 +96,10 @@ module ManageIQ::Providers
       Benchmark.realtime_block(:collect_data) do
         begin
           context.collect_metrics
-        rescue => e
-          _log.error("Hawkular metrics service unavailable: #{e.message}")
+        rescue NoMetricsFoundError => e
+          _log.debug("Metrics unavailable: #{e.message}")
+        rescue StandardError => e
+          _log.error("Metrics unavailable: #{e.message}")
           ems.update_attributes(:last_metrics_error       => :unavailable,
                                 :last_metrics_update_date => Time.now.utc) if ems
           raise
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_capture_context.rb
@@ -132,7 +132,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
     # @param key [String] the metrics key/group_id (e.g. cpu/usage).
     # @return [String] the metrics full key name (e.g. machine/example.com/cpu/usage).
     def get_metrics_key(raw_metrics, type, key)
-      raise CollectionFailure, "no #{type} metrics found for [#{@target.name}]" unless raw_metrics[type]
+      raise NoMetricsFoundError, "no #{type} metrics found for [#{@target.name}]" unless raw_metrics[type]
 
       # each object has only one metrics with some key/group_id ( e.g. each node has only one cpu/usage )
       raw_metrics[type].keys.find { |e| e.ends_with?(key) }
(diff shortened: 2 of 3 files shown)


yaacov:
@miq-bot add_label metrics

@cben @moolitayer @Ladas @gtanzillo please review

yaacov:
@miq-bot add_label wip

I can't reproduce the trace back on my machine, so I can't be sure this really helps ...

@gtanzillo can you check on the RADAR system ?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -8,6 +8,12 @@ def log_severity
       end
     end
 
+    class NoMetricsFoundWarning < RuntimeError
+      def log_severity
+        :warn
```
moolitayer: you meant debug and use below?
since debug will be the only option for now I'm ok with removing log_severity and simply calling debug below for this class, so only change the class name.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -95,8 +101,10 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
       Benchmark.realtime_block(:collect_data) do
         begin
           context.collect_metrics
-        rescue => e
-          _log.error("Hawkular metrics service unavailable: #{e.message}")
+        rescue NoMetricsFoundWarning => e
+          _log.debug("Metrics unavailable: #{e.message}")
```
moolitayer: :+1: for note registering `:last_metrics_error` on the ems in this case

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
yaacov: 👍

yaacov:
@miq-bot remove_label wip

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -95,8 +96,10 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
       Benchmark.realtime_block(:collect_data) do
         begin
           context.collect_metrics
-        rescue => e
-          _log.error("Hawkular metrics service unavailable: #{e.message}")
+        rescue NoMetricsFoundWarning => e
```
moolitayer: @yaacov Could you remove the Warning from the class name?
We used these for things the produce `_log.warn` until now
`NoMetricsFound` is fine

yaacov: 👍

yaacov: no `...Error` postfix, in an Error obj ?

cben: We're using an exception to communicate something we don't consider an error.
I don't see a problem calling a spade a spade :-)
LGTM either way.

Remind me, why are we raising the exception at all (rather than just return empty results)?
Is it so it gets retried?  For the logging here?

yaacov: > We're using an exception to communicate something we don't consider an error.

It is an error pod wise, but a warning provider wise, one day we will tell the user one of his/her pods does not have metrics.

> why are we raising the exception at all (rather than just return empty results)?

yes, we do not know why we have no metrics ... we may have metrics in the next run ...

yaacov:
@moolitayer ^^

moolitayer:
@yaacov if we want this backported for us we will need a different bug since the one included now is closed as not a bug

yaacov:
> @yaacov if we want this backported for us we will need a different bug since the one included now is closed as not a bug

Reading @gtanzillo mail I think this does not solve the problem ... so I think we should wait until we are sure ...

yaacov:
@miq-bot add_label gaprindashvili/yes

yaacov:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1537195
target: 5.9.1

yaacov:
Note:
When backporting must be together with #228

simaishi:

```
$ git log -1
commit 73a8d776f6eb044e95e040ff43533577677a5525
Date:   Tue Jan 30 18:15:23 2018 +0200

    Merge pull request #227 from yaacov/remove-stucktrace-when-pod-has-no-metrics
    
    Do not raise an error when metrics are missing for one object
    (cherry picked from commit c83b42ae4af3a49d3c6a562510aff02210a24ac5)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1552314
```

PR #230, opened 2018-02-01
Properly support "security_protocol" for alerts endpoint

elad661:
Other endpoints support both "verify_ssl" boolean and a
"security_protocol" string (which can be "ssl-without-validation",
"ssl-with-validation-custom-ca" or "ssl-with-validation".

While supporting both the boolean and the string is confusing,
it is important to be consistent across all endpoint types -
otherwise things can get very confusing for API users.

This commit makes sure the alerts endpoint calls verify_ssl_mode
just like other endpoints.

cc @moolitayer @cben

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb
@@ -50,6 +50,31 @@ describe ManageIQ::Providers::Kubernetes::MonitoringManager do
     end
   end
 
+  context "#authentication - no ssl, string" do
+    let(:prometheus_alerts_endpoint) do
+      EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+      FactoryGirl.build(
+        :endpoint,
+        :role              => 'prometheus_alerts',
+        :hostname          => 'alerts-prometheus.example.com',
+        :port              => 443,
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/kubernetes/monitoring_manager_spec.rb
```diff
@@ -97,4 +122,52 @@
...
+
+  context "#authentication - custom ssl, string" do
+    let(:prometheus_alerts_endpoint) do
+      EvmSpecHelper.local_miq_server(:zone => Zone.seed)
+      FactoryGirl.build(
+        :endpoint,
+        :role                  => 'prometheus_alerts',
+        :hostname              => 'alerts-prometheus.example.com',
+        :port                  => 443,
+        :security_protocol     => "ssl-without-validation",
```
cben: does :certificate_authority matter if you set without-validation?  is this what you wanted to test?

elad661: good catch! I specified the wrong security_protocol

cben:
I doubt the tests check actual SSL validation (vcr/webmock probably hook above it) but they do exercise the code.  LGTM.

cben:
@moolitayer suggested unit-testing the return value of the function itself.  I'll try to do some kind of followup (if I don't forget).

PR #233, opened 2018-02-20
abort container ssa if can't fetch metadata

enoodle:
This will prevent issues and report correctly on event where the pod can't be accessed for some reason on that stage. This could happen for example if the pod is deleted by an external command.

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
@@ -101,6 +101,16 @@ class MockImageInspectorClient
   end
 end
 
+class MockCantAccessImageInspectorClient
+  def fetch_metadata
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
```diff
@@ -478,6 +488,19 @@ def create_pod_definition
       end
     end
 
+    context 'fetch_metadata fails to access pod' do
```
cben: Typo fails

cben:
LGTM, need another reviewer.

enoodle:
@zeari @nimrodshn Please review

nimrodshn:
LGTM :+1:

cben:
please add gaprindashvili label if you have an opinion for/against

PR #235, opened 2018-02-20
Test :batch strategy with and without :use_ar_object.

cben:
We hope to only use the fastest {:saver_strategy => :batch, :use_ar_object => false} in production, but testing all helps expose bugs in refresh...


cben:
Did the same for the targeted_refresh spec.

PR #238, opened 2018-02-28
[WIP] disable use-docker-socket feature, remove docker socket from volumes

enoodle:
This will use image inspector docker-less ability to fetch images, that will be added in https://github.com/openshift/image-inspector/pull/95

* requires the SYS_CHROOT capability for the pod


enoodle:
cc @zeari @nimrodshn

PR #234, opened 2018-02-20
bump image-inspector image version to 2.4

enoodle:
Use the new Image-Inspector image.

Main changelog for this version:
Enhancements:
* Can scan running containers
* Can use ClamAV Scanner
* Enable token authentication for web access
* Reports the status of image acquiring
* Support for authentication token to all endpoints

and many bugfixes.

Will solve a couple of BZs:

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1512824
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1484936

This will also allow us to move forward with another BZ https://bugzilla.redhat.com/show_bug.cgi?id=1524616 (which will also require a new version of image-inspector-client to be made..)

PR #239, opened 2018-02-28
[WIP] add authentication to image inspector requests

enoodle:
this PR leverages a new security feature in Image Inspector, which allows a custom authentication check using a shared secret.

The shared secret is initialized by passing the pod environment variable INSPECTOR_AUTH_TOKEN. HTTP requests send to the image inspector pod should be signed with the header "X-Auth-Token" whose value is the shared secret in order to authenticate.

Note that if Image Inspector does not find INSPECTOR_AUTH_TOKEN in the environment, it will skip performing this authentication check on HTTP requests.

- [ ] This depends on #234 to use new image-inspector that includes https://github.com/openshift/image-inspector/pull/38
- [x] This depends on #263 a new version of image-inspector-client released with https://github.com/moolitayer/image-inspector-client/pull/17 => #263 

resubmission of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/104


enoodle:
cc @zeari @nimrodshn

cben:
According to #104 this depends using at least inspector 2.2 => this also depends on #234.

cben:
RFE: https://bugzilla.redhat.com/show_bug.cgi?id=1524616

PR #249, opened 2018-04-16
Add a config option to force legacy hawkular collector

yaacov:
**Description**
Add a config option to force legacy hawkular collector

This PR adds an option to force using old endpoint for metrics capture from Hawkular.

evm => configuration => advanced

**Screenshot** 
force-legacy-hawkular

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1562090

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -51,24 +51,39 @@ module ManageIQ::Providers
       }
     }
 
+    def hawkular_force_legacy?
+      worker_class = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCollectorWorker
+
+      worker_class.worker_settings[:hawkular_force_legacy] || false
+    end
+
+    def prometheus_capture_context(target, start_time, end_time)
+      PrometheusCaptureContext.new(target, start_time, end_time, INTERVAL)
+    end
+
+    def hawkular_capture_context(target, start_time, end_time)
+      # if we have new version of hawkular endpoints (/m endpoint)
+      # use the new collector
+      context = HawkularCaptureContext.new(target, start_time, end_time, INTERVAL)
+      if hawkular_force_legacy? || !context.m_endpoint?
+        _log.info("Using Hawkular legacy metrics collector")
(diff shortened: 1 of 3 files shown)


yaacov:
@miq-bot add_label metrics bug

@cben @moolitayer please review

chrispy1:
@miq-bot add_label blocker

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
```diff
@@ -6,26 +6,98 @@
...
+        @ems_kubernetes_prometheus,
+        @node_prometheus,
+        5.minutes.ago,
+        0.minutes.ago
+      )
+
+      expect(context).to be_a(described_class::PrometheusCaptureContext)
+    end
+
+    it "detect hawkular metrics provider with m metrics endpoint" do
```
cben: Perhaps add another test for the new setting -  with m endpoint but with stub_settings for in legacy.

yaacov: yes, I was in the middle of doing the tests, when I saw the mail about blocker, I think we need this merged ~today ...

If I will add the missing tests to this PR, I will do it in a new commit, so it will be easier to re-review, and we decide to me merge without them, we can always add them in a new PR.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -51,24 +51,39 @@ def log_severity
       }
     }
 
+    def hawkular_force_lagacy?
+      worker_class = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCollectorWorker
+
+      worker_class.worker_settings[:hawkular_force_lagacy] || false
```
Ladas: typo s/lagacy/legacy/ ?

yaacov: Thanks ! fixing

yaacov:
note: leaving un-squashed for now, will squash when ok'ed by reviewers

yaacov:
squashed

PR #255, opened 2018-05-22
mapping.tag is a category Tag, calling .category is ill-defined

cben:
There are category/entry Tags, and category/entry Classifications.
```
                                   .parent
 category Classification "cat"   <----------   entry Classification "ent"
           |^                ^                        |^
       .tag||.classification  \_______________    .tag||.classification
           v|                     .category   \       v|
 category Tag /managed/cat                     entry Tag /managed/cat/ent
```
`.category` is supposed to get the "uncle", the category Classification from an *entry* Tag.

`mapping.tag` is a category Tag (for any-value mappings which are the kind we use).
Currently `.category` on a category Tag ("/managed/cat") happens to work same as `.classification`, but it's ill-defined and will stop working after https://github.com/ManageIQ/manageiq/pull/17418.


cben:
@kbrock @Fryguy please review

cben:
@moolitayer ping this is same as https://github.com/ManageIQ/manageiq-providers-openshift/pull/98

PR #256, opened 2018-05-23
Respect kubevirt token if provided

masayag:
When changing the kubevirt's token under the virtualization tab, the new
token should be used for the validation check instead of the existing token.

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/77

Depends on https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/81

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -81,8 +81,11 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
 
   def verify_kubevirt_credentials
     ensure_infra_manager
-    infra_manager.verify_credentials
-    infra_manager.verify_virt_supported
+    options = {
+      :token => authentication_token(:kubevirt),
+    }
+    infra_manager.verify_credentials(:kubevirt, options)
+    infra_manager.verify_virt_supported(options)
   end
 
   # UI methods for determining availability of fields


masayag:
@cben please review.

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -81,8 +81,11 @@ def verify_prometheus_alerts_credentials
 
   def verify_kubevirt_credentials
     ensure_infra_manager
-    infra_manager.verify_credentials
-    infra_manager.verify_virt_supported
+    options = {
+      :token => authentication_token(:kubevirt),
+    }
+    infra_manager.verify_credentials(:kubevirt, options)
+    infra_manager.verify_virt_supported(options)
```
cben: Does passing options in this line change anything?  AFAICT, `verify_virt_supported` would use `authentication_token(:kubevirt)` anyway?
(I'm ok with passing for consistency, just still trying to understand the change)

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -81,8 +81,11 @@ def verify_prometheus_alerts_credentials
 
   def verify_kubevirt_credentials
     ensure_infra_manager
-    infra_manager.verify_credentials
-    infra_manager.verify_virt_supported
+    options = {
+      :token => authentication_token(:kubevirt),
+    }
+    infra_manager.verify_credentials(:kubevirt, options)
```
cben: Same question here, I'm not sure how this line changes anything.
`verify_credentials` ignores :kubevirt and calls `with_provider_connection(options, &:valid?)`.
IIUC that calls `connect` which would use `authentication_token(:kubevirt)` in any case?

cben:
Am I missing something that happens when UI validates unsaved AR objects?
Does `parent_container_manager.authentication_token(:kubevirt)` then work differently than `child_infra_manager.authentication_token(:kubevirt)` ?
Is the scenario creating new provider / editing existing one / both?
Can you explain in more details what actually happens and how it lead to a bug?

If you can add a test case this fixes, that may also help explain it.

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
masayag: @cben yes, it will be used here:

https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/81/files#diff-ff17b05a7c525eec833f0ddb0134a0d9R121

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
masayag: the first parameter is ignored, but passing _ failed

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
cben: But the value you pass is `authentication_token(:kubevirt)` and when you didn't pass it, the previous code there was defaulting to `authentication_token(:kubevirt)` anyway.
The only difference I see is which manager you call `authentication_token` on.  But child manager is delegating that to parent?

masayag: `authentication_token(:kubevirt)` is being called for different objects (one for the container provider and the other for the infra provider)

cben:
Disscussed offline — yes, it's something that happens when UI validates unsaved AR objects, the unsaved object graph is not fully detached.
I'd like to understand deeper, but this is at worst harmless, LGTM.

Plus, https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/81 already got merged, so this has to be merged too.  nagging @yaacov now to get another eye.

PR #259, opened 2018-06-11
Kubevirt should report its own status

masayag:
The UI expects each provider to report its status.
Since kubevirt uses its own token, it might be valid while the container
provider is invalid.

That causes the default_auth_status in the UI [1] to report a failed
status when editing the kubevirt provider as an infrastructure provider,
since it refers to the status of the parent container manager. As a
result of that failure, the dialog cannot be saved.

For that purpose, the delegated call to authentication_status_ok? should
be implemented on the kubevirt provider side.

[1] https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/mixins/ems_common_angular.rb#L861

Provider's PR is here:
https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/92

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/71

### DIFF
--- a/app/models/manageiq/providers/kubernetes/virtualization_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/virtualization_manager_mixin.rb
@@ -7,7 +7,6 @@ module ManageIQ::Providers::Kubernetes::VirtualizationManagerMixin
     delegate :authentication_check,
              :authentication_for_summary,
              :authentication_status,
-             :authentication_status_ok,
              :authentication_token,
              :authentications,
              :endpoints,


masayag:
@miq-bot assign @cben

app/models/manageiq/providers/kubernetes/virtualization_manager_mixin.rb
```diff
@@ -7,7 +7,6 @@ module ManageIQ::Providers::Kubernetes::VirtualizationManagerMixin
     delegate :authentication_check,
              :authentication_for_summary,
              :authentication_status,
```
cben: how about the other auth-related methods?
- authentication_check — I think this one works but don't remember why :)  Is it always called with correct authtype?
- authentication_for_summary collects all auths, probably ok to delegate.
  For kubevirt, it will include regular and kubevirt auth.
  This is mostly used by UI for display to user, though update_ems_button_save also does `update_ems.authentication_check_types_queue(update_ems.authentication_for_summary.pluck(:authtype), ...)`.
- authentication_status also looks at all auths, returns the "worst" one.
  Does it make sense to reflect parent problem when kubevirt auth is good but parent is bad?

masayag: @cben I'll address the rest of the methods in a different PR, specifically, the only method that shouldn't be delegated is authentication_status, and it also requires to implement the authentication_for_providers on kubevirt infra_manager to return the kubevirt provider only.

That will assure that the status on the quadicon of the provider reflects only the status of kubevirt.

As for the other methods, they seems safe to be delegated since they are being called with the auth_type.

PR #260, opened 2018-06-13
Kubevirt should reports its auth status

masayag:
In addition to aca446f6561f34179519688a7540e93d85d25096 [1], another
authentication related method can be directed to kubevirt instead being
delegated to its parent provider.

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/259

Fixes https://github.com/ManageIQ/manageiq-providers-kubevirt/issues/35


cben:
cc @pkliczewski

Issue #196, opened 2017-12-27
Get rid of old_ems_id


cben:
Archiving used to clear `ems_id` and store it in `old_ems_id`.
`old_ems_id` is unnecessary now that we retain `ems_id` (since https://github.com/ManageIQ/manageiq/pull/15251 and related PRs).

- [x] backfill unset `ems_id` from `old_ems_id` — https://github.com/ManageIQ/manageiq-schema/pull/18 already did that.
- [x] drop any code checking `old_ems_id` / `old_ext_management_system`.
  - [x] #267 Stop checking in plugins/manageiq-providers-kubernetes/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb
  - [x] Stop checking in app/models/metric/ci_mixin/capture.rb #17695
  - [x] remove and stop including app/models/mixins/old_ems_mixin.rb  #17695
  - [x] Stop checking in app/models/chargeback_container_image.rb, app/models/chargeback_container_project.rb #17695
- [x] delete the column.  :fire: :dragon: https://github.com/ManageIQ/manageiq-schema/pull/593

Then there is also `old_container_project_id` in ContainerGroup, not sure if that's obsolete, out of scope here.

cben:
@miq-bot add-label technical debt

cben:
@miq-bot remove-label stale

agrare:
Marking this as `pinned` since we really should finish this out

PR #267, opened 2018-07-04
Stop falling back to old_ems_id

cben:
@miq-bot add-label technical debt
A step towards https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/196.

`ems_id` is always set since we switched to soft delete.  `ems_id` got backfilled from `old_ems_id` for old disconnected records too by https://github.com/ManageIQ/manageiq-schema/pull/18.

@zeari @Ladas please review


cben:
@agrare please review, I shouldn't merge my own PRs.

agrare:
Nice cleanup @cben

PR #272, opened 2018-08-11
Allow proper targeted refresh

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/17846

Allow proper targeted refresh. For passing specs, we must define the correct targeted_arel queries.


Ladas:
@agrare @gtanzillo and we need to build proper targeted_arel queries. Although I think those can be also auto-generated later.

So with this, the targeted persistor should be working (the specs probably do not have 100% coverage, so as we do streaming refresh, we'll be adding more specs)

app/models/manageiq/providers/kubernetes/inventory/persister/definitions/container_collections.rb
```diff
@@ -91,23 +91,38 @@ def initialize_taggings
...
     relation = manager.public_send(association)
     query = ContainerCondition.where(
       :container_entity_type => relation.model.base_class.name,
-      :container_entity_id   => relation, # nested SELECT. TODO: compare to a JOIN.
+      :container_entity_id   => relation, # TODO(lsmola): JOIN will be much better, should be defined as relation
     )
 
+    targeted_arel = lambda do |inventory_collection|
+      # TODO(lsmola): if we use :association ^ instead of :arel, this can be autogenerated
+      p_collection = inventory_collection.parent_inventory_collections.first
```
cben: is this distinct from `parent_collection` defined above?

app/models/manageiq/providers/kubernetes/inventory/persister/definitions/container_collections.rb
```diff
@@ -149,15 +177,27 @@ def add_taggings(parent_name)
...
-      :taggable_id   => relation,
+      :taggable_id   => relation, # TODO(lsmola): JOIN will be much better, should be defined as relation
     ).joins(:tag).merge(Tag.controlled_by_mapping)
 
+    targeted_arel = lambda do |inventory_collection|
+      # TODO(lsmola): if we use :association ^ instead of :arel, this can be autogenerated
+      p_collection = inventory_collection.parent_inventory_collections.first
+      rel = p_collection.db_collection_for_comparison_for(p_collection.targeted_scope.primary_references)
+
+      query = Tagging.where(
```
cben: the `query =` part here does nothing, you just need return value from the lambda, right?

app/models/manageiq/providers/kubernetes/inventory/persister/target_collection.rb
```diff
@@ -1,15 +1,9 @@
 class ManageIQ::Providers::Kubernetes::Inventory::Persister::TargetCollection < ManageIQ::Providers::Kubernetes::Inventory::Persister::ContainerManager
   def targeted?
-    false # TODO(lsmola) get ready for true, which means a proper targeted refresh. That will require more effort.
+    true
   end
 
-  def shared_options
```
cben: why is this deleted?

Ladas: it's defined in the base persister

Ladas:
@cben the targeted_arel is needed because we are not able to build it automatically. Instead of DRYing it, we can rather rewrite the :arel attributes to relations on ems. Then the targeted_arel should be able to generate itself. :-)

cben:
@JPrause the Milestone github offers me here is Sprint 92 ending yesterday — should I use that one?  Is it normal that 93 doesn't exist yet?

cben:
@agrare @gtanzillo I merged assuming you already reviewed this and I'm the one this is blocked on, but seems not (or you didn't say anything)?   
I think it's worth a 2nd review.

JPrause:
@cben yes, please use 92. The milestones are closed Tuesday mornings. If 92 gets closed before you can assign, you can use 93.

JPrause:
@cben I should have said,...every two weeks on the Tuesday morning before sprint review.

PR #273, opened 2018-08-21
bin/setup - create manageiq/log manually

cben:
Similar to https://github.com/ManageIQ/manageiq-ui-classic/pull/4502, https://github.com/ManageIQ/manageiq-providers-nuage/pull/130.

Since https://github.com/ManageIQ/manageiq/pull/17663, there is no manageiq/log directory by default.
This breaks Travis e.g. https://travis-ci.org/ManageIQ/manageiq-providers-kubernetes/jobs/418623069

@miq-bot add-label bug, developer, test

@kbrock @himdel @miha-plesko please review

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -9,7 +9,7 @@ unless gem_root.join("spec/manageiq").exist?
 end
 
 # ensure a log dir always exists
-system "mkdir -p spec/manageiq/log"
+gem_root.join("spec/manageiq/log").mkpath
 
 require gem_root.join("spec/manageiq/lib/manageiq/environment").to_s
 ManageIQ::Environment.manageiq_plugin_setup(gem_root)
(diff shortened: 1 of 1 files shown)


cben:
Switched to native ruby, PTAL
Deviated from https://github.com/ManageIQ/manageiq-providers-nuage/pull/130 — kept it outside the if, as I think if you manually create or symlink a fresh spec/manageiq with log/, `bin/setup` should create it.

cben:
@kbrock @agrare can you merge (dont want to merge my own)

tadeboro:
I think that in case `spec/manageiq` is a symlink, `bin/setup` should not mess with developer managed content. When developer creates this symlink, he took the responsibility of keeping the things up and running.

PR #274, opened 2018-08-22
Move streaming_refresh_enabled method

agrare:
Move the method which checks if streaming_refresh is enabled in config
closer to the supports :streaming_refresh block.

Ref: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/271#discussion_r211266196


cben:
Thanks, LGTM

PR #275, opened 2018-08-22
Restart stale watch threads where they left off

agrare:
If a watch thread fails for any reason (exception, timeout, etc...)
restart it with the last resourceVersion so it picks up where it left
off.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
@@ -38,6 +38,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
       full_refresh
       start_watch_threads
     else
+      ensure_watch_threads
       targeted_refresh
     end
   end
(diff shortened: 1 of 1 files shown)


agrare:
@cben going from your linked comment https://github.com/kubernetes/website/pull/6540#discussion_r160745657 I'm setting the collection resourceVersion to the latest notice.object's resourceVersion.  Does this look like a valid approach?

app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
```diff
@@ -103,14 +114,26 @@ def stop_watch_threads
...
 
-    until finish.value
-      watch_stream.each { |notice| queue.push(notice) }
+    until finished?
+      watch_stream.each do |notice|
+        # Update the collection resourceVersion to be the most recent
+        # object's resourceVersion so that if this watch has to be restarted
+        # it will pick up where it left off.
+        resource_version = notice.object.metadata.resourceVersion
+        resource_versions[entity_type] = resource_version
```
cben: this updates a shared hash (at distinct keys) from multiple threads, is that fully safe in ruby?

cben:
Merging.  I'll be on vacation all next week so will try to unblock as much stuff as possible today.

app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
agrare: Yeah I'll use a `Concurrent::Hash` to be safe

PR #276, opened 2018-08-23
Refresh service catalog entities

Ladas:
Refresh service catalog entities

Depends on:
- [x] https://github.com/ManageIQ/manageiq-schema/pull/260
- [x] https://github.com/ManageIQ/manageiq/pull/17895

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -68,6 +68,15 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     def kubernetes_version
       'v1'
     end
+
+    def kubernetes_service_catalog_connect(hostname, port, options)
+      options = {:path => '/apis/servicecatalog.k8s.io', :version => service_catalog_api_version}.merge(options)
+      kubernetes_connect(hostname, port, options)
+    end
+
+    def service_catalog_api_version
+      'v1beta1'
+    end
   end
 
   PERF_ROLLUP_CHILDREN = :container_nodes
--- a/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
@@ -1,22 +1,22 @@
 class ManageIQ::Providers::Kubernetes::Inventory::Collector::ContainerManager < ManageIQ::Providers::Kubernetes::Inventory::Collector
   def namespaces
-    entities_iterator(connection, :namespaces)
+    @namespaces ||= entities_iterator(connection, :namespaces)
   end
 
   def pods
-    entities_iterator(connection, :pods)
+    @pods ||= entities_iterator(connection, :pods)
   end
 
   def cluster_service_classes
-    entities_iterator(service_catalog_connection, :cluster_service_classes)
+    @cluster_service_classes ||= entities_iterator(service_catalog_connection, :cluster_service_classes)
   end
 
   def cluster_service_plans
-    entities_iterator(service_catalog_connection, :cluster_service_plans)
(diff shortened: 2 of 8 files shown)


Ladas:
cc @gtanzillo

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
```diff
@@ -1,15 +1,42 @@
 class ManageIQ::Providers::Kubernetes::Inventory::Collector::ContainerManager < ManageIQ::Providers::Kubernetes::Inventory::Collector
   def namespaces
-    @namespaces ||= connection.get_namespaces
+    entities_iterator(connection, :namespaces)
   end
 
   def pods
-    @pods ||= connection.get_pods
```
Ladas: @agrare hm, you are memoizing this because you then fetch the collection.,resourceVersion from it, right?

Would it make sense to just memoize this value? Because I would like this to be an iterator doing paginated query later

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -68,6 +68,15 @@ def kubernetes_auth_options(options)
     def kubernetes_version
       'v1'
     end
+
+    def kubernetes_service_catalog_connect(hostname, port, options)
+      options = {:path => '/apis/servicecatalog.k8s.io', :version => service_catalog_api_version}.merge(options)
```
cben: we have too many ways to connect, what with `ems.connect(:service => ...)`, also I think supporting `ems.connect(:path => ...)`, and code somewhere I can't find now deriving path & version dynamically.

Nothing wrong with this method however :+1:.  Just wish I had time to clean up some of the other ways...

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
```diff
@@ -1,15 +1,42 @@
...
   end
+
+  def service_catalog_connection
+    @service_catalog_connection ||= manager.connect(:service => "kubernetes_service_catalog")
+  end
+
+  def entities_iterator(client, entity)
+    # TODO(lsmola) change to iterator, that will fetch paginated response from the server, never fetching everything
+    # at once
+    client.send("get_#{entity}")
```
cben: nit: it's important here that `entity` is plural, e.g. "pods" and not "pod", but this is unobvious looking at an arg named `entity`.

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
```diff
@@ -1,15 +1,42 @@
 class ManageIQ::Providers::Kubernetes::Inventory::Collector::ContainerManager < ManageIQ::Providers::Kubernetes::Inventory::Collector
   def namespaces
-    @namespaces ||= connection.get_namespaces
+    @namespaces ||= entities_iterator(connection, :namespaces)
   end
 
   def pods
-    @pods ||= connection.get_pods
+    @pods ||= entities_iterator(connection, :pods)
```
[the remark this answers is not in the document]
cben: interesting.
A slightly open question in kubeclient iterator interface was how it should expose whole-collection resource_version (and other metadata).  Should iterator yield it with every item, or return it at the end?
With streaming parse (https://github.com/abonas/kubeclient/pull/254), this depends on order of json.  k8s puts `items` last, so version is known by time we start yielding; this is unlikely to change but I feel weird hardcoding this assumption...
For chunking (https://github.com/abonas/kubeclient/issues/283), we'll have metadata in each chunk, so either API works.  Chunking also brings risk of getting `410 Gone` if we wait too long, not sure how to handle that.

Ladas:
@agrare ugh, I've miss clicked, wanted to merge your AWS PR :-/

agrare:
@Ladas no worries, LGTM was going to merge it when I saw it was green anyway

PR #278, opened 2018-08-23
Finish the watch stream connection when exiting

agrare:
Prevent thread joins always timing out due to the watch stream blocking
for far longer than the join timeout.

Ref: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/271#discussion_r212213786

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
@@ -1,7 +1,7 @@
 module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
   include Vmdb::Logging
 
-  attr_accessor :ems, :finish, :initial, :resource_versions, :watch_threads, :queue
+  attr_accessor :ems, :initial, :resource_versions, :watch_streams, :watch_threads, :queue
 
   def after_initialize
     super
@@ -26,11 +26,11 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
   private
 
   def setup_streaming_refresh
-    self.finish            = Concurrent::AtomicBoolean.new(false)
     self.initial           = true
     self.queue             = Queue.new
     self.resource_versions = {}
-    self.watch_threads     = {}
+    self.watch_streams     = Concurrent::Map.new
+    self.watch_threads     = Concurrent::Map.new
   end
 
   def do_work_streaming_refresh
@@ -108,7 +108,12 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
   def stop_watch_threads
     safe_log("#{log_header} Stopping watch threads...")
 
-    finish.value = true
+    # First call WatchStream#finish to forcibly terminate the loop, this
+    # closes the HTTP connection and will cause the #each method to raise an
+    # exception (until https://github.com/abonas/kubeclient/pull/315 is applied).
+    watch_streams.each_value(&:finish)
+
+    # Next loop through each thread and join them cleanly
     watch_threads.each_value { |thread| thread.join(10) }
 
     safe_log("#{log_header} Stopping watch threads...Complete")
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
```diff
@@ -108,7 +108,13 @@ def ensure_watch_threads
   def stop_watch_threads
     safe_log("#{log_header} Stopping watch threads...")
 
-    finish.value = true
+    # First call WatchStream#finish to forcibly terminate the loop, this
+    # closes the HTTP connection and will cause the #each method to raise an
```
cben: to be precise, the #each method crashes and rescues *internally*.
it's awkward implementation, but should be invisible.
if you see any exception leaking outside it, it's kubeclient bug (we already escalated the rescue couple times, https://github.com/abonas/kubeclient/pull/280 and https://github.com/abonas/kubeclient/pull/315)

oh, I see, I think manageiq is still using old kubeclient 2.5.2 where some are uncatched :-(
includes 280 but not 315.
bumping kubeclient to 4.x is still blocked on several things that are basically ready, I just need to test image scanning.
I could also release 2.5.3 with 315 backported.

ok to catch HTTP::ConnectionError for now but let's have a more precise comment and do nag me to bump or backport.

app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
```diff
@@ -122,18 +128,24 @@ def watch_thread(entity_type)
...
-        # Update the collection resourceVersion to be the most recent
-        # object's resourceVersion so that if this watch has to be restarted
-        # it will pick up where it left off.
-        resource_version = notice.object.metadata.resourceVersion
-        resource_versions[entity_type] = resource_version
+    watch_stream = start_watch(entity_type, resource_version)
+    watch_streams[entity_type] = watch_stream
+
+    begin
+      loop do
```
cben: is the loop for restarting if it disconnects (https://github.com/abonas/kubeclient/pull/275)?  have you actually observed that, and does this work?
I think you need a fresh `start_watch()` to reconnect (with fresh resource_version of course)

agrare: I haven't seen it yet but yes that's exactly why I had this loop, if this won't work i'll drop the loop and just let the thread restart.

cben: :+1: thread restarting looks good, less code paths is good

app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
agrare: :+1:

agrare: Okay added that PR to the comment re: the exception

app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
```diff
@@ -26,11 +26,11 @@ def before_exit(_message, _exit_code)
   private
 
   def setup_streaming_refresh
-    self.finish            = Concurrent::AtomicBoolean.new(false)
     self.initial           = true
     self.queue             = Queue.new
     self.resource_versions = {}
-    self.watch_threads     = {}
+    self.watch_streams     = Concurrent::Map.new
+    self.watch_threads     = Concurrent::Map.new
```
cben: reminder, not necessarily in this PR: you also planned resource_versions to be Concurrent::Map.

is this https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Hash.html ?
> locks against the object itself for every method call, ensuring only one thread can be reading or writing at a time

will lock contention be a problem?
I'd guess this is minor compared to overhead of reading notices from network.
anyway I'm cool with erring on side of safety and profiling later.

PR #281, opened 2018-09-04
Handle missing service catalog endpoint

agrare:
Running against minishift I get a KubeException 404 error collecting
service catalog classes and plans:
An error has occurred during work processing: HTTP status code 404, 404
Not Found


cben:
404 makes sense.  Service catalog adds its APIs it adds at run time, they won't work without it running or simply not installed at all (it [uses api agregatation](https://svc-cat.io/docs/install/), CRDs would be similar).

Ideally I'd like to ignore only the API not existing.  If it exists but different error, should at least log it.
There's probably a way to check that using kubeclient `.discover`, not sure if public API is sufficient.  Opened https://github.com/abonas/kubeclient/issues/349 to figure out and document.

PR #283, opened 2018-09-11
Fix watches when using service catalog endpoint

agrare:
Using watches with the service catalog endpoint is failing when it goes to parse a watch update.  Eventually we should use watches for the services but not up to that yet


cben:
LGTM

PR #284, opened 2018-09-11
Use discover to catch service catalog

agrare:
Rather than waiting to call Kubeclient::Client#get_entity_type for
the 404 exception we can call Kubeclient::Client#discover and catch the
exception when connecting.

- [x] Depends on: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/283


cben:
LGTM.  Neat use of `.tap` :+1:

agrare:
@cben there's probably a reason for why it is the way it is but what would you think about kubeclient doing `.discover` when connecting?  That way if the service isn't setup or is down for some reason the connect would actually fail (which at least to me seems more correct) rather than on the first method call?

cben:
answered on https://github.com/abonas/kubeclient/issues/349

PR #287, opened 2018-09-13
Streaming refresh for service catalog entities

agrare:
This adds the service_catalog entity types to the list of watched entities.

1. I had to change the collector methods from cluster_service_offerings and cluster_service_parameters_sets to the native entity type so that i could get the resourceVersion with the entity type from the collection.
2. I moved from using `ems.connect` to saving the connect options and using raw_connect so that each thread didn't require an ActiveRecord connection.  With 5 threads I was exhausting the number of connections in the connection pool.

Dependent: https://github.com/ManageIQ/manageiq-providers-openshift/pull/113

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
@@ -74,7 +74,13 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
 
   def save_resource_versions(inventory)
     entity_types.each do |entity_type|
-      resource_versions[entity_type] = inventory.collector.send(entity_type).resourceVersion
+      collection = inventory.collector.send(entity_type)
+
+      # TODO: this is if we can't get service catalog entities and just return an
+      # empty array.
+      # When we move to getting the full collector for an entity in the same thread
+      # that we do watches this won't be an issue.
+      resource_versions[entity_type] = collection.resourceVersion if collection.respond_to?(:resourceVersion)
     end
   end
 
@@ -188,6 +194,10 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
   end
 
   def entity_types
+    @entity_types ||= all_entity_types.reject { |entity| connection_for_entity(entity).nil? }
+  end
+
+  def all_entity_types
     kubernetes_entity_types + service_catalog_entity_types
   end
 
--- a/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
@@ -30,7 +30,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
   def setup_streaming_refresh
     self.initial           = true
     self.queue             = Queue.new
-    self.resource_versions = {}
+    self.resource_versions = Concurrent::Map.new
     self.watch_streams     = Concurrent::Map.new
     self.watch_threads     = Concurrent::Map.new
   end
--- a/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
@@ -155,12 +155,22 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::StreamingRefreshMixin
     connection_for_entity(entity_type).send(watch_method, :resource_version => resource_version)
   end
 
-  def connection_for_entity(_entity_type)
-    kubernetes_connection
+  def connection_for_entity(entity_type)
+    if kubernetes_entity_types.include?(entity_type)
+      kubernetes_connection
+    elsif service_catalog_entity_types.include?(entity_type)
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -53,6 +53,12 @@ def kubernetes_connect(hostname, port, options)
           :read => Settings.ems.ems_kubernetes.read_timeout.to_f_with_method
         }
       )
+
+      # Test the API endpoint at connect time to prevent exception being raised
+      # on first method call
+      conn&.discover
```
cben: Does the rest of the PR depend on this?
It's a far-reaching change. Probably safe, and a good idea, but I wonder if it's worth splitting to its own PR in case it breaks something and we'll want to revert.

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
```diff
@@ -28,19 +28,16 @@ def cluster_service_parameters_sets
   private
 
   def connection
-    @connection ||= manager.connect(:service => "kubernetes")
+    @connection ||= connect("kubernetes")
```
cben: if `connect` fails, catches KubeException and returns nil, this will re-attempt it every time.
is that desired or should we cache the negative result?
it's possible that catalog was absent/down initially and is installed/started later.
I tend to say at least for full refreshes should re-attempt everything.

anyway, if it's deliverate I think it's worth a comment, otherwise it looks accidental.

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
Ladas: it looks like this was just moved? So it should be safe, since we still call it for every connect

agrare: Used to only call discover for the service catalog endpoint, you're right I'll pull this out into a separate PR

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
agrare: > if connect fails, catches KubeException and returns nil, this will re-attempt it every time.

Yes I wanted the ability for the endpoint to be "fixed" without having to restart appliances to pick it up.

Ladas: you're right, but again that was the current state. So we should refactor that in followup, if needed

app/models/manageiq/providers/kubernetes/container_manager/streaming_refresh_mixin.rb
```diff
@@ -170,8 +186,27 @@ def kubernetes_entity_types
...
+  end
+
+  def all_entity_types
+    kubernetes_entity_types + service_catalog_entity_types
+  end
+
+  def connect(service = "kubernetes")
+    opts = connect_options.merge(:service => service)
+
+    ems.class.raw_connect(opts[:hostname], opts[:port], opts).tap(&:discover)
```
cben: So this is all fallout from not sharing Kubeclient::Client objects between threads, right?
This led to calling `connect` in each thread, which required reading DB in each thread, which exhausted AR connections?  (BTW, can't we just increase AR connections?  is it sensible to limit AR connections < num threads?)

I know I recently said :+1: to not sharing Client objects, but I thought it'd be a strict simplification for MIQ...
Anyway I'm OK with this (and I'll take this mild complication any day over debugging thread bugs).

But with my Kubeclient hat on, the simplest way to represent connect options ought to be a `Client` instance, and I'm sad if it fails that role.
Was there any actual problem sharing a Kubeclient instance between multiple threads?

agrare: Thats a good point however since the provider doesn't control the number of threads we can't just increase the pool size.  We would have to (in MIQ core) make the pool size the largest of all possible thread counts in all the plugins.

Even though we're caching connections between threads I think avoiding a DB connection is still a good idea here.  We already have three endpoints that we are connecting to and if a different thread connects to each one first plus the main thread we're getting close to that connection pool limit already.  If we add another endpoint or two in this process then we're blown again.

Thoughts?

Ladas: we'll solve the DB pool limit by the collector/persistor, which we are doing next, right? Since collector will not connect to the DB anymore.

agrare: Well until we get the collectors out of the rails engine to force no db connections it will still be possible for threads to run into this.

cben: how does running out of AR connections manifest?  do we get a clear error, or is it just stuck and needs a wizard to understand that's the problem?

hmm, collectors without rails will be nice but would need to get configuration where to connect somehow.  so will look kinda like this (compute options separately) anyway I guess...

agrare: Oh there's a clear error, active record raises an exception and the thread exits and gets restarted (usually just to die again when it tries to connect).

And yeah we need to have the connection options passed in, here I use the main thread's AR connection to save this before starting the threads but when this is running in a container we will most likely get these options passed in as env vars.

PR #288, opened 2018-09-17
Discover API endpoint in kubernetes_connection

agrare:
Call discover when connecting to a kubernetes endpoint to prevent exceptions on future method calls.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -42,7 +42,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
     def kubernetes_connect(hostname, port, options)
       require 'kubeclient'
 
-      Kubeclient::Client.new(
+      conn = Kubeclient::Client.new(
         raw_api_endpoint(hostname, port, options[:path]),
         options[:version] || kubernetes_version,
         :ssl_options    => Kubeclient::Client::DEFAULT_SSL_OPTIONS.merge(options[:ssl_options] || {}),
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -53,6 +53,12 @@ def kubernetes_connect(hostname, port, options)
           :read => Settings.ems.ems_kubernetes.read_timeout.to_f_with_method
         }
       )
+
+      # Test the API endpoint at connect time to prevent exception being raised
+      # on first method call
+      conn&.discover
```
cben: I think `&.` could be `.` ?  `conn` is never nil.

agrare: :+1: done

cben: hmm, Travis fails `undefined method `discover' for nil:NilClass`
guess I was wrong, I just don't see how :confused:

cben: https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/spec/models/manageiq/providers/kubernetes/container_manager_spec.rb#L168 (and L180)
`expect(Kubeclient::Client).to receive(:new)`
Do you feel like improving stubbing to return something having `.discover` method, or give up?

agrare: It might just be specific to the spec tests stubbing the return of Kubeclient::Connect

agrare: I'm changing those tests to actually return an instance of Kubeclient::Client, will push in a minute

agrare: There we go that did the trick

agrare:
Thanks @cben !

PR #290, opened 2018-09-17
Refresh Service Catalog service instances

Ladas:
Refresh Service Catalog service instances

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/service_instance.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/service_instance.rb
@@ -1,0 +1,2 @@
+class ManageIQ::Providers::Kubernetes::ContainerManager::ServiceInstance < ::ServiceInstance
+end
--- a/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
@@ -25,6 +25,10 @@ class ManageIQ::Providers::Kubernetes::Inventory::Collector::ContainerManager < 
     @cluster_service_plans ||= service_catalog_connection&.get_cluster_service_plans || []
   end
 
+  def service_instances
+    @service_instances ||= service_catalog_connection.get_service_instances
+  end
+
   private
 
   def connection
--- a/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
@@ -26,7 +26,7 @@ class ManageIQ::Providers::Kubernetes::Inventory::Collector::ContainerManager < 
   end
 
   def service_instances
-    @service_instances ||= service_catalog_connection.get_service_instances
+    @service_instances ||= service_catalog_connection&.get_service_instances || []
   end
 
   private
--- a/app/models/manageiq/providers/kubernetes/inventory/collector/watches.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/collector/watches.rb
@@ -20,4 +20,8 @@ class ManageIQ::Providers::Kubernetes::Inventory::Collector::Watches < ManageIQ:
   def cluster_service_plans
     @cluster_service_plans ||= notices['ClusterServicePlan']&.map { |notice| notice.object } || []
   end
+
+  def service_instances
+    @service_instances ||= notices['ServiceInstance']&.map { |notice| notice.object } || []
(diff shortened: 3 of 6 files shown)


app/models/manageiq/providers/kubernetes/inventory/collector/watches.rb
```diff
@@ -20,4 +20,8 @@ def cluster_service_classes
   def cluster_service_plans
     @cluster_service_plans ||= notices['ClusterServicePlan']&.map { |notice| notice.object } || []
   end
+
+  def service_instances
+    @service_instances ||= notices['ServiceInstance']&.map { |notice| notice.object } || []
```
Ladas: @agrare not sure if this will just work?

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
```diff
@@ -25,6 +25,10 @@ def cluster_service_plans
     @cluster_service_plans ||= service_catalog_connection&.get_cluster_service_plans || []
   end
 
+  def service_instances
+    @service_instances ||= service_catalog_connection.get_service_instances
```
cben: what if `service_catalog_connection` fails?  does this need `&.` and `|| []` like above?

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -57,6 +58,26 @@ def parse_service_class(service_class)
...
+  def parse_service_instance(service_instance)
+    persister.service_instances.build(
+      :name                   => service_instance.metadata.name,
+      :ems_ref                => service_instance.spec.externalID,
+      :service_offering       => persister.service_offerings.lazy_find(service_instance.spec.clusterServiceClassRef.name),
+      :service_parameters_set => persister.service_parameters_sets.lazy_find(service_instance.spec.clusterServicePlanRef.name),
+      :extra                  => {
+        :metadata => service_instance.metadata,
+        :spec     => service_instance.spec,
+        :status   => service_instance.status
```
cben: Do we need this copy of all data in inventory just in case?
Was the idea of `extra` to actively store things just in case we'll need them, or as place we *could later* store new things without migration, in case we need to?

Ladas: right now, we are not sure what everything will we needed by the ServiceCatalog team

app/models/manageiq/providers/kubernetes/container_manager/service_instance.rb
```diff
@@ -0,0 +1,2 @@
+class ManageIQ::Providers::Kubernetes::ContainerManager::ServiceInstance < ::ServiceInstance
```
cben: is it used automagically or do you need `:type => 'ManageIQ::Providers::Kubernetes::ContainerManager::ServiceInstance'` in parser?

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
Ladas: well, it's probably not a too big of an issue to keep it here

we'll probably move away from having everything as a column, unless it's really needed

app/models/manageiq/providers/kubernetes/container_manager/service_instance.rb
Ladas: it will be generated automatically :-)

app/models/manageiq/providers/kubernetes/inventory/collector/watches.rb
agrare: @Ladas as long as the entity type name is actually `'ServiceInstance'` should be fine :)

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -57,6 +58,26 @@ def parse_service_class(service_class)
...
+    service_instances.each do |service_instance|
+      parse_service_instance(service_instance)
+    end
+  end
+
+  def parse_service_instance(service_instance)
+    persister.service_instances.build(
+      :name                   => service_instance.metadata.name,
+      :ems_ref                => service_instance.spec.externalID,
+      :service_offering       => persister.service_offerings.lazy_find(service_instance.spec.clusterServiceClassRef.name),
```
agrare: Looks like `service_instance.spec.clusterServiceClassRef` can be nil

PR #291, opened 2018-09-18
Handle deleted entities

agrare:
Process delete notices using the targeted_scope


app/models/manageiq/providers/kubernetes/inventory/parser/watches.rb
```diff
@@ -13,44 +13,36 @@ def parse
...
-      next if notice.type == "DELETED"
-
-      parse_service_plan(service_plan)
+      service_plan_inv_obj = parse_service_plan(service_plan)
+      assign_deleted_on(service_plan_inv_obj, service_plan) if notice.type == "DELETED"
     end
   end
+
+  def assign_deleted_on(inv_obj, object)
+    inv_obj.data[:deleted_on] = object.metadata.deletionTimestamp || Time.now.utc
```
Ladas: ```object&.metadata&.deletionTimestamp``` to be safe?

PR #297, opened 2018-10-23
Fix spec assuming default zone

agrare:
This container scan test was assuming the EMS was put in the default
zone when checking the zone name.


agrare:
cc @Ladas

PR #300, opened 2018-11-01
[GAPRINDASHVILI] Fix MetricsCapture and Refresher specs failing due to a missing zone

cben:
Partial backport of #289 (minus file that didn't exist on gaprindashvili).

Stubbing MiqServer.my_zone to return a dummy string was affecting the zone with which events were added to MiqQueue, which broke now that MiqQueue requires a real zone (https://github.com/ManageIQ/manageiq/pull/18068#issuecomment-435087412).

@agrare @Ladas @simaishi please review

Will also want to backport #297 but I think this will be green without it (we'll see in a minute), if yes I'll just make #297 gaprindashvili/yes separately...

### DIFF
--- a/spec/factories/ext_management_system.rb
+++ b/spec/factories/ext_management_system.rb
@@ -1,0 +1,8 @@
+FactoryGirl.define do
+  factory :ems_kubernetes_with_zone, :parent => :ems_kubernetes do
+    zone do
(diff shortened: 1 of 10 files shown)


agrare:
Thanks @cben !

PR #303, opened 2018-11-12
Raise event containerimage_scan_complete only when scan succeeds.

lfu:
Raise event containerimage_scan_complete when smart analysis finishes without errors.
Otherwise raise event containerimage_scan_abort.

Includes https://github.com/ManageIQ/manageiq/pull/18189.

https://bugzilla.redhat.com/show_bug.cgi?id=1499161

@miq-bot add_label bug, hammer/yes

cc @gmcculloug


cben:
LGTM.
naming: I found "_abort" for failure confusing but it follows pattern of `vm_scan_complete` / `vm_scan_abort` so :+1: for consistency.

The core PR has to be merged first, right?

app/models/manageiq/providers/kubernetes/container_manager/scanning/job.rb
```diff
@@ -264,6 +260,11 @@ def cleanup(*args)
     queue_signal(:done, message, status)
   end
 
+  def raise_image_scan_event(image)
+    suffix = state == 'finishing' ? 'complete' : 'abort'
```
gmcculloug: Would ultimately like to see some of these text fields, like `finishing`, turned into constants.  But that's better in a follow refactoring PR.

PR #304, opened 2018-11-16
Different way to check for scan job's complete state.

lfu:
State finishing does not exist in hammer branch.

Followup of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/303.

https://bugzilla.redhat.com/show_bug.cgi?id=1499161

@miq-bot add_label bug, hammer/yes

cc @gmcculloug

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
@@ -216,17 +216,20 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
     context "completes successfully" do
       before(:each) do
         allow_any_instance_of(described_class).to receive_messages(:collect_compliance_data) unless OpenscapResult.openscap_available?
-
         expect(@job.state).to eq 'waiting_to_start'
-        @job.signal(:start)
       end
 
       it 'should report success' do
+        expect(MiqEvent).to receive(:raise_evm_job_event).with(@image, :type => "scan", :suffix => "complete")
+
+        @job.signal(:start)
+
         expect(@job.state).to eq 'finished'
         expect(@job.status).to eq 'ok'
       end
 
       it 'should persist openscap data' do
+        @job.signal(:start)
         skip unless OpenscapResult.openscap_available?
 
         expect(@image.openscap_result).to be
@@ -424,6 +427,7 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
       end
 
       it 'should report the error' do
+        expect(MiqEvent).to receive(:raise_evm_job_event).with(@image, :type => "scan", :suffix => "abort")
         @job.signal(:start)
         expect(@job.state).to eq 'finished'
         expect(@job.status).to eq 'error'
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
```diff
@@ -495,6 +503,7 @@ def create_pod_definition
         allow_any_instance_of(described_class).to receive_messages(
           :image_inspector_client => MockFailedImageInspectorClient.new("Error", OSCAP_ERROR_MSG, "", IMAGE_ID)
         )
+        expect(MiqEvent).to receive(:raise_evm_job_event).with(@image, :type => "scan", :suffix => "complete")
```
lfu: @cben Is this event correct for this test case?

cben: Hmm, good question!  Raised it on https://bugzilla.redhat.com/show_bug.cgi?id=1499161#c26

This is not a total scanning failure — that's `MockCantAccessImageInspectorClient`.
Here test simulates image-inspector that did run but gave OSCAP_ERROR_MSG="Unable to run OpenSCAP: Unable to get RHEL dist number", which is normal behavior for scanning images not based on RHEL.
We don't get security results, but it is a definite answer — this is not some transient error and there's no point re-scanning.
And we do get some non-security SSA info — CFME gets access to image filesystem and collects some info like package versions (I think only for RPM-based images?).

So, is such image compliant or not?
I guess not, and should not do positive openshift annotation, but not sure if should add negative "deny execution" annotation?
Can we give user control of how to treat this case by editing the policy?  I don't see an easy way,  we'd need a third event — neither _abort nor _complete, probably not worth it...  Well I suppose most users that want the "deny execution" action at all, also want the draconian interpretation "not known good => bad".
So IMHO this case should raise _abort event.  But I want to see if anyone on the BZ has an opinion.
Non-blocker for this PR.

cben:
LGTM.
I'm also open to backporting #299 which added "finishing".
But this is harmless, at least the tests here :heart: are worth having anyway.

Let me just double-check this passes on hammer too, with and without @299, and I'll merge...

cben:
Yep, they pass both way.

Meanwhile #299 got hammer/backported.
However reverting this to old logic `suffix = state == 'finishing' ? 'complete' : 'abort'` on hammer doe fails 4 tests with ``undefined method `scan_result_message' for nil:NilClass`` and "complete" becoming "abort": https://gist.github.com/cben/e7fef6c712e5400063df3ee33500c760  Shrug, I'm too lazy to dig why.

So merging as-is with new working logic.

simaishi:
Hammer backport details:

```
$ git log -1
commit 3974721e329bb77147f303c687c6adda13abcbf6
Date:   Tue Nov 20 14:05:51 2018 +0200

    Merge pull request #304 from lfu/openscap_1499161_2
    
    Different way to check for scan job's complete state.
    
    (cherry picked from commit 2454b3d7ccdb6712cf2c3eb3ffe2987df88de94b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1499161
```

PR #306, opened 2018-11-20
[GAPRINDASHVILI] Use working-set instead of usage for memory

yaacov:
Ref: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1654463


yaacov:
> @yaacov we need to be able to handle both working_set and usage on gaprindashvili because we support openshift < 3.7 right?

If the collector detect old Hawkular it will fall back to legacy code that use `usage`, only if it detect `/apu/m/..` endpoint, it will use the new version using `working_set`

a. the code for that check for old version:

https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb#L68

b. legace code using `usage`:

https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_legacy_capture_context.rb

yaacov:
@agrare hi, sorry I gave wrong links above:

code that check for old version:

https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/gaprindashvili/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb#L68

legacy code that use usage`:

https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/gaprindashvili/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_legacy_capture_context.rb

agrare:
Oh okay and the existing definition of "new hawkular" already includes the versions that have working_set ?  In that case I'm :+1: on this

yaacov:
> already includes the versions that have working_set ? 

yes, it's 3.7 for `m` endpoint also.

PR #308, opened 2018-11-26
Honor the event catcher filtered_events blacklist

agrare:
The base_manager event_catcher maintains a list of filtered events from
the ems event type blacklist.  This list should be checked when checking
if an event is to be filtered or not.

Based on https://github.com/ManageIQ/manageiq/pull/14632

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1653871

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
@@ -75,10 +75,7 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EventCatcherMixin
     end
 
     supported_reasons = ENABLED_EVENTS[event_data[:kind]] || []
-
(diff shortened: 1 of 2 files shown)


cben:
LGTM and I like #309 even better.

But I'm confused — what happened to https://github.com/ManageIQ/manageiq/pull/14632, did it get lost when we moved that code to this repo?

agrare:
> and I like #309 even better.

I agree, started out just doing the simplest fix for the bz but I liked keeping all of the "filtered" logic together a lot better.

agrare:
> what happened to ManageIQ/manageiq#14632, did it get lost when we moved that code to this repo?

Yeah it was prior to extracting the code out of core, I grabbed the tests as they were but thought I could simplify the other changes so didn't just copy the whole PR.

agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/309

PR #312, opened 2018-12-04
API Chunking Support

agrare:
Pass a limit to each get_ call to prevent massive numbers of objects being returned and have them returned in multiple pages instead.

This reduces memory usage of the collector and also reduces load on the kubernetes API server.

https://bugzilla.redhat.com/show_bug.cgi?id=1722808


cben:
@agrare How much work do you think it would take to revive this?  cc @oourfali 
Your Kubeclient changes have been merged & released, and the code here looks good :+1:
It should also automatically work with older api that ignores `limit`.
My guess is the bigger effort would be testing for regressions / unforeseen stability problems down the line.

#### Motivation:
In https://bugzilla.redhat.com/1722808, users were seeing timeouts on openshift side for manageiq's large refresh requests.  I suspect in that particular setup it could be relieved on openshift side (https://bugzilla.redhat.com/1729461), and I'm not aware of any concrete bug in manageiq contributing to that, aside from the fact manageiq makes huge requests.
But chunked requests are known to be better for all layers in k8s (apiserver, etcd) so would improve scaling if we can spare the effort.

manageiq-providers-kubernetes.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency("hawkular-client",                 "~> 4.1")
   s.add_runtime_dependency("image-inspector-client",          "~> 2.0")
-  s.add_runtime_dependency("kubeclient",                      "~> 4.0")
+  s.add_runtime_dependency("kubeclient",                      "~> 4.1")
```
cben: Better bump to 4.1.1, it fixed a regression https://github.com/abonas/kubeclient/issues/376 found in 4.1.0.
(at a glance `KUBERNETES_ENTITIES`, `OPENSHIFT_ENTITIES` that we request are unaffected but anyway)

agrare:
Hey @cben the biggest issue for me was not being able to re-record the VCRs for specs 
If the original systems are still around that would be helpful otherwise we'll basically need to rewrite them.

manageiq-providers-kubernetes.gemspec
agrare: :+1:

spec/models/manageiq/providers/kubernetes/container_manager/refresh_mixin_spec.rb
```diff
@@ -2,7 +2,8 @@
 
 describe ManageIQ::Providers::Kubernetes::ContainerManager::RefresherMixin do
   let(:client)  { double("client") }
-  let(:dummy) { (Class.new { include ManageIQ::Providers::Kubernetes::ContainerManager::RefresherMixin }).new }
```
agrare: This was failing because there was no refresher_options defined on `Class`

config/settings.yml
```diff
@@ -59,6 +59,7 @@
   :kubernetes:
     :refresh_interval: 15.minutes
     :streaming_refresh: false
+    :chunk_size: 1_000
```
cben: :+1: nice

cben:
VCR recording is quite automated with https://github.com/ManageIQ/guides/blob/master/providers/openshift.md#automated-script-to-record-new-vcr

You need a clean empty openshift running, easiest is with minishift as documented in that doc.
Then you run that script from the providers-openshift repo.

We never bothered with separate recording of a kubernetes cluster; instead we just copy the updated VCRs over here to providers-kubernetes repo.
```
cp -v ../manageiq-providers-openshift/spec/vcr_cassettes/manageiq/providers/openshift/container_manager/refresher*deletions*  spec/vcr_cassettes/manageiq/providers/kubernetes/container_manager/
```
The assertions are split between the tests in the 2 repos, here we test the upstream k8s objects.

- I don't know how the 2 `refresher_targeted_refresh_*.yml` files got recorded, I think Ladislav added them.

- The `refresher.yml` file apparently dates back to single manageiq repo and was manually edited since then :-(
  IIRC the plan was to eventually entirely replace it with the scripted `refresher*deletions*` files, not sure if there are any valuable tests left that use it.
  Maybe manually edit it to match the limit param, or config vcr to ignore limit param when matching; the responses would remain as a test for "server doesn't understand chunking"?

agrare:
Thanks @cben !  I got the kubernetes and openshift specs passing.  I had the original `refresher.yml` specs match only on the method and path ignoring the query so it is the "no-limit-supported" test also

manageiq-providers-kubernetes.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency("hawkular-client",                 "~> 4.1")
   s.add_runtime_dependency("image-inspector-client",          "~> 2.0")
-  s.add_runtime_dependency("kubeclient",                      "~> 4.0")
+  s.add_runtime_dependency("kubeclient",                      "~> 4.1.1")
```
Fryguy: Minor, but if you want to keep the auto-upgrade on minor version, perhaps... 

```suggestion
  s.add_runtime_dependency("kubeclient",                      "~> 4.1", ">= 4.1.1")
```

cben: Good catch.
I'm not aware of incompatible changes in any 4.y.
There is a second dependency through `fog-kubevirt`, currently on `kubeclient (~> 4.1.2)`, so it's better to not constrain without need — say later kubevirt discover they need something in 4.3, they won't need to bump two places.

spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
```diff
@@ -28,7 +28,7 @@
   end
 
   def full_refresh_test(expected_extra_tags: [])
-    VCR.use_cassette(described_class.name.underscore) do # , :record => :new_episodes) do
+    VCR.use_cassette(described_class.name.underscore, :match_requests_on => [:method, :host, :path]) do # , :record => :new_episodes) do
```
cben: Could you add a comment here, to explain why we're doing this?

agrare: :+1: sure thing

agrare: Done

simaishi:
Hammer backport details:

```
$ git log -1
commit b3577b80897095a285e26adee9aa8c18dd48fa9f
Date:   Fri Sep 20 16:32:29 2019 +0300

    Merge pull request #312 from agrare/use_api_chunking
    
    API Chunking Support
    
    (cherry picked from commit bea27dd9fc49d1e488b8d17ec9c92bdc62046d22)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1754071
```

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 5066f371cfd15ddbe3032fb82604e59629a11ea3
Date:   Fri Sep 20 16:32:29 2019 +0300

    Merge pull request #312 from agrare/use_api_chunking
    
    API Chunking Support
    
    (cherry picked from commit bea27dd9fc49d1e488b8d17ec9c92bdc62046d22)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1767834
```

PR #318, opened 2018-12-31
[GAPRINDASHVILI] use memory working_set in legacy collector

yaacov:
**Description**
Add an option to use `memory/working_set` metrics with legacy metrics collector.

In https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/306 [1] we only replaced `memory/usage` with `memory/working_set` for the new collector, leaving the legacy collector unchanged.

The `memory/working-set` was intruduced in heapster v1.1 [2] used in origin metrics v1.3 [3], and can be used safely with legacy openshift/hawkular (pre 3.7).

This PR adds the option to use the legacy collector using `memory/working_set`.
It adds an option to enable/disable this option to make is possible for users that do not have the new `memory/working_set` or want to leave the current metric unchanged to chose witch metrics to sue.

Options is set to `false` not to force the change on old systems.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1654463

[1] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/306
[2] https://github.com/kubernetes-retired/heapster/commit/3ca8b4fb0b7d469e5334772e6b4a736834bce03c
[3] https://github.com/openshift/origin-metrics/commit/e4837dfb85f98f17c908aae8511f2527a3be7813

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_legacy_capture_context.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/hawkular_legacy_capture_context.rb
@@ -3,11 +3,25 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
     include ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::HawkularClientMixin
     include CaptureContextMixin
 
+    def hawkular_legacy_use_working_set?
+      worker_class = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCollectorWorker
+
+      worker_class.worker_settings[:legacy_use_working_set] || false
+    end
+
+    def memory_category
+      if hawkular_legacy_use_working_set?
+        "/memory/working_set"
+      else
+        "/memory/usage"
+      end
+    end
+
     def collect_node_metrics
       cpu_resid = "machine/#{@target.name}/cpu/usage"
       process_cpu_counters_rate(fetch_counters_rate(cpu_resid))
 
-      mem_resid = "machine/#{@target.name}/memory/usage"
+      mem_resid = "machine/#{@target.name}#{memory_category}"
       process_mem_gauges_data(fetch_gauges_data(mem_resid))
 
       net_resid = "machine/#{@target.name}/network"
@@ -23,7 +37,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
       cpu_resid = "#{@target.name}/#{group_id}/cpu/usage"
       process_cpu_counters_rate(fetch_counters_rate(cpu_resid))
 
-      mem_resid = "#{@target.name}/#{group_id}/memory/usage"
+      mem_resid = "#{@target.name}/#{group_id}#{memory_category}"
       process_mem_gauges_data(fetch_gauges_data(mem_resid))
     end
 
@@ -36,7 +50,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture
       process_cpu_counters_rate(compute_summation(cpu_counters))
 
       mem_gauges = @target.containers.collect do |c|
-        fetch_gauges_data("#{c.name}/#{group_id}/memory/usage")
+        fetch_gauges_data("#{c.name}/#{group_id}#{memory_category}")
       end
       process_mem_gauges_data(compute_summation(mem_gauges))
 
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -81,5 +81,6 @@
             :prometheus_open_timeout: 5
             :prometheus_request_timeout: 30
             :hawkular_force_legacy: true
+            :legacy_use_working_set: false
         :ems_refresh_worker:
           :ems_refresh_worker_kubernetes: {}


yaacov:
Waiting for Ryan answer.

cc:// @oourfali @cben

cben:
Why gaprindashvili, shouldn't this go to master first?

Code LGTM, waiting for decision.

yaacov:
> Why gaprindashvili, shouldn't this go to master first?

In master we already replaced `memory/usage` with `memory/working_set` in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305, when we back-ported to gaprindashvili we didn't want users of old openshifts to fail getting metrics because they do not have the `memory/working_set` so we left the code for legacy openshifts unchanged.

Unfortunately the BZ reporter is using gaprindashvili and an openshift/heapster that fall-back to legacy collector because it is missing the `/m` endpoint, so in the case of the BZ reporter, the partial fix we back-ported, did not help.

So to help the BZ reporter we need to backport the change in the legacy collectror also, we make this change optional because some users that fall-back to the legacy collector may not have the `memory/working_set` metrics, and we do not want to break their metrics without an option to collect the existing `memory/usage` in case they do not have `memory/working_set`. 

We are currently waiting for the reporter to check that this fix actually fix the problem in the affected system.

agrare:
> when we back-ported to gaprindashvili we didn't want users of old openshifts to fail getting metrics because they do not have the memory/working_set so we left the code for legacy openshifts unchanged.

Yeah I'm not sure why we wouldn't do this in master first either.  When those users with old openshifts upgrade to that version of MIQ wouldn't they have the same issue that this is trying to fix though?

The extra config option complicates things for the user, is it possible to try to collect `working_set` first and fall back to `usage` if that comes back nil?  If we're worried about the performance of that we could do a cache_with_timeout by EMS to store which supports working_set

yaacov:
> Yeah I'm not sure why we wouldn't do this in master first either. When those users with old openshifts upgrade to that version of MIQ wouldn't they have the same issue that this is trying to fix though?

In https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305#issuecomment-439954326 and https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305#issuecomment-439997721 we decided that in master we should not support pre v3.7 openshifts (acutally I now know that this should be safe for ver >= v1.3).

> The extra config option complicates things for the user, is it possible to try to collect working_set first and fall back to usage if that comes back nil? If we're worried about the performance of that we could do a cache_with_timeout by EMS to store which supports working_set

Yes, we have a discussion on that, the main reason that made us chose a flag, over automatic check, is that we want to know what metrics is used in a definite way.

For example, in the case if the BZ reporter the code automatically looked for the '/m' endpoint and than fall-back to use 'mamory/usage' instead of 'mamory/working_set' without having any indication for the BZ reporter, who just saw "bad" numbers :-(, If we set it in option flag, it either collect the metrics user is asking for, or fail loudly and make the user look for the reason (being for example not having the metric in hawkular) .

agrare:
I don't think there is anything wrong with looking at the version of the API that is being connected to and changing the behavior based on that.  That isn't abusing a single column that is just working with what the provider is giving us.

Switching on existence of `/m` does seem odd but if we looked at the openshift/heapster version and changed our behavior based on that would be fine.

I would support switching everyone over to `memory/working_set` or looking at the API version and checking if `memory/working_set` was present in that version.

I really don't like just adding another config option, is there any reason why someone with a current enough version _wouldn't_ want to use working_set?

yaacov:
> I really don't like just adding another config option, is there any reason why someone with a current enough version wouldn't want to use working_set?

We already changed it to working_set for all on master, the only thing preventing us from doing it here is that the reporter of the BZ can not promise us we will not need to collect metrics from pre v1.3 openshifts.

An option here will be to set the flag to `true` by default instead of `false`, WDYT ?

> I would support switching everyone over to memory/working_set or looking at the API version and checking if memory/working_set was present in that version.

We looked into this solution but we already failed matching the right openshift version with the available metrics once.

IMHO, If we go with dynamic checking - checking if memory/working_set is present is preferable than checking for specific version.

IMHO, User flag is preferable, because it makes the metrics predictable, users will either get the metrics the want or fail, they will never get another metrics they did not want, even if the system does not support the wanted metrics.

My vote is for keeping the the flag, but making it true by default, @agrare is this an option you can agree on ?

agrare:
> We looked into this solution but we already failed matching the right openshift version with the available metrics once.

The same version of openshift can have different versions of metrics/heapster or there was just an issue with the version matching?

> the only thing preventing us from doing it here is that the reporter of the BZ can not promise us we will not need to collect metrics from pre v1.3 openshifts.

Do we actually support pre v1.3 versions of origin with gaprindashvili?  It looks like OCP 3.5 is the oldest version supported by gaprindashvili (https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.6/pdf/support_matrix/Red_Hat_CloudForms-4.6-Support_Matrix-en-US.pdf)  which is based on origin v1.5 (https://docs.openshift.com/container-platform/3.5/release_notes/ocp_3_5_release_notes.html).

> My vote is for keeping the the flag, but making it true by default, @agrare is this an option you can agree on ?

It looks like v1.3 is sufficiently old that defaulting to using working_set would be preferable to the opposite.

yaacov:
> The same version of openshift can have different versions of metrics/heapster or there was just an issue with the version matching?

It was an issue of missing documentation, we looked at dates of PRs that introduce specific feature and miss calculted the version each feature was introduced. I now think working_set was introduced in v1.3 but I may be wrong ( again :-) )
    
> Do we actually support pre v1.3 versions of origin with gaprindashvili?

If we do not, we can -

a - revert, https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/306
b - take https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305

@agrare @cben WDYT ?

agrare:
@yaacov got confirmation from @Fryguy that we do not need to support OCP prior to 3.5 on gaprindashvili so I would recommend dropping the config option and just taking https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305 as is.

yaacov:
Thanks !

@adam, can you take care of reverting #306 and taking #305 ?

agrare:
@yaacov yes, https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/320

yaacov:
closing in favour of #320

PR #319, opened 2019-01-03
MonitoringManager spec: update to explicitly start from unvalidated auth

cben:
https://github.com/ManageIQ/manageiq/pull/18266 changed factories by default to create "Valid" auth, requiring explicit opt-out for test scenarios like here unvalidated->`check_types`->expected status.

Fixes red travis: https://travis-ci.org/ManageIQ/manageiq-providers-kubernetes/jobs/474807718#L2098
(tracking issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/4921#issuecomment-451129620)

@kbrock @bdunne please review


kbrock:
Thanks @cben - I may have messed up a few of these :(

PR #320, opened 2019-01-03
[GAPRINDASHVILI] Revert #306 in favor of clean backport of #305

agrare:
Legacy hawkular supports memory/working_set for all versions of openshift supported by gaprindashvili so instead of using the old memory/usage counter    switch to the new memory/working_set value.

This reverts https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/306 so that we can backport https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/305 as-is instead.


yaacov:
Thanks :+1:

PR #335, opened 2019-09-05
Call .create_job from the Scanning class

agrare:
Instead of calling Job.create_job and passing the scanning job class
name we can call create_job from the final class not the base class.
This opens up future refactorings for the base create_job class.


agrare:
This has a cyclic dep on https://github.com/ManageIQ/manageiq/pull/19265

PR #340, opened 2019-10-07
Only support graph refresh

agrare:
Related to: https://github.com/ManageIQ/manageiq-providers-openshift/pull/148


agrare:
@Ladas okay these two can be merged together now

PR #341, opened 2019-10-07
Add params_for_create and verify_credentials

agrare:
Add methods to get the required parameters to create a provider and a
method to verify the credentials.

https://github.com/ManageIQ/manageiq/issues/18818

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -61,11 +61,12 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
           :validate   => [{:type => "required-validator"}]
         },
         {
-          :component  => "text-field",
-          :name       => "endpoints.default.port",
-          :type       => "number",
-          :isRequired => true,
-          :validate   => [
+          :component    => "text-field",
+          :name         => "endpoints.default.port",
+          :type         => "number",
+          :isRequired   => true,
+          :initialValue => DEFAULT_PORT,
+          :validate     => [
             {
               :type => "required-validator"
             },
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -50,71 +50,87 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
   end
 
   def self.params_for_create
-    @params_for_create ||= {
-      :title  => "Configure #{description}",
-      :fields => [
-        {
-          :component  => "text-field",
-          :name       => "endpoints.default.hostname",
-          :label      => "Hostname",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
-        },
-        {
-          :component    => "text-field",
-          :name         => "endpoints.default.port",
-          :type         => "number",
-          :isRequired   => true,
-          :initialValue => DEFAULT_PORT,
-          :validate     => [
-            {
-              :type => "required-validator"
-            },
-            {
-              :type             => "validatorTypes.MIN_NUMBER_VALUE",
-              :includeThreshold => true,
-              :value            => 1
-            },
-            {
-              :type             => "validatorTypes.MAX_NUMBER_VALUE",
-              :includeThreshold => true,
-              :value            => 65_535
-            }
-          ]
-        },
-        {
-          :component    => "text-field",
-          :name         => "endpoints.default.path",
-          :label        => "API Path",
-          :type         => "hidden",
-          :initialValue => "/api"
-        },
-        {
-          :component  => "text-field",
-          :name       => "endpoints.default.bearer",
-          :label      => "Token",
-          :type       => "password",
-          :isRequired => true,
-          :validate   => [{:type => "required-validator"}]
-        },
-        {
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -49,6 +49,89 @@ def self.description
...
+          :name       => "endpoints.default.hostname",
+          :label      => "Hostname",
+          :isRequired => true,
+          :validate   => [{:type => "required-validator"}]
+        },
+        {
+          :component  => "text-field",
+          :name       => "endpoints.default.port",
+          :type       => "number",
+          :isRequired => true,
```
cben: Does UI set a default for this?
Typically 6443 for k8s, 8443 for openahift, but both may use regular 443 if accessing from outside via a route.

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -49,6 +49,89 @@ def self.description
...
+              :includeThreshold => true,
+              :value            => 65_535
+            }
+          ]
+        },
+        {
+          :component    => "text-field",
+          :name         => "endpoints.default.path",
+          :label        => "API Path",
+          :type         => "hidden",
```
cben: Ack, I don't think we ever exposed this to users in UI.  Is it even settable in API?  I suspect it's hard-coded to `/api` in code (or nothing, thats default and kubeclient let's you omit it).  I'm pretty sure openahift's sister `/oapi` is hard-coded.

There are use cases to change it, notably Rancher mounts the API at a deeper sub-path, though kubeclient currently doesn't deal with that (https://github.com/abonas/kubeclient/issues/318)

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -49,6 +49,89 @@ def self.description
...
+          :label     => "HTTP Proxy"
+        },
+        {
+          :component => "checkbox",
+          :name      => "endpoints.default.verify_ssl",
+          :label     => "Verify SSL"
+        },
+        {
+          :component => "text-field",
+          :name      => "endpoints.default.ca_file",
```
cben: Is it a file?  UI/API is passing the CA as ascii string not a file name?

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -49,6 +49,89 @@ def self.description
...
+        },
+        {
+          :component => "text-field",
+          :name      => "endpoints.default.http_proxy",
+          :label     => "HTTP Proxy"
+        },
+        {
+          :component => "checkbox",
+          :name      => "endpoints.default.verify_ssl",
+          :label     => "Verify SSL"
```
cben: This should have initialValue true 💂

app/models/manageiq/providers/kubernetes/container_manager.rb
agrare: Yeah the UI defaults to 6443 for k8s and 8443 for openshift, I'll update this to add an initialValue :+1:

app/models/manageiq/providers/kubernetes/container_manager.rb
agrare: Yeah I almost didn't add this at all, but I wanted to allow e.g. the API to override any of the options passed to `kubernetes_connect` even if the UI doesn't show it

app/models/manageiq/providers/kubernetes/container_manager.rb
agrare: The ssl_options on kubeclient call this `ca_file` [link](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/common.rb#L13)

app/models/manageiq/providers/kubernetes/container_manager.rb
agrare: :+1:

agrare:
hey @cben !

> Should this include all options in UI or just minimum to check auth

It will include all of the options on the UI, I have this in WIP until I get the alerts, metrics, virt endpoints added

> I think it's worth preparing the openshift PR as well to see which parts change and how it can override them.

Absolutely, I'm currently doing that to see if most of this can be moved to the container_manager_mixin and overridden, or better to just keep them separate.  Should have that PR today.

> What about downstream productization? I think it hides some options not supported by Red Hat, would that affect this?

That is interesting I hadn't considered that, is this, I assume this is handled today with the `config/permissions.yml` blacklist.  If the entire provider is disabled (e.g. ems-type:kubernetes) then I think the UI just shouldn't show this at all.  If one of the sub-types were disabled though this would be tricky.  We could check if a child-manager type is enabled when building the DDF to return.

I'll investigate this but it looks like currently the only instance of a base-type being supported and a single child-manager not being supported is the S3 child manager of AWS which doesn't require different endpoint credentials.

> P.S. I'm off to holidays/shutdown till ~Sep 23. I'll try to respond to questions but don't block for my review :)

Enjoy!

app/models/manageiq/providers/kubernetes/container_manager.rb
agrare: Done

agrare:
I punted on the other endpoint types so that we can get 100% coverage of all the provider default endpoints (kubernetes and openshift are the last two)

PR #352, opened 2019-12-12
Fix parsing capacity fields of the form 50M or 1G

agrare:
If the storage capacity attribute is of the form 50M or 1G the
parse_capacity_field method which only expects IEC 60027-2 style values
will fail:

```
MIQ(ManageIQ::Providers::Openshift::ContainerManager::RefreshParser#parse_capacity_field) Capacity attribute - storage was in bad format - 1G
```

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1782902

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb
@@ -875,7 +875,7 @@ module ManageIQ::Providers::Kubernetes
     def parse_capacity_field(key, val)
       return nil unless val
       begin
-        val.iec_60027_2_to_i
+        parse_quantity(val)
       rescue ArgumentError
         _log.warn("Capacity attribute - #{key} was in bad format - #{val}")
         nil
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_parser_spec.rb
@@ -1502,5 +1502,53 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshParser do
         )
       end
     end
+
+    describe "parse_capacity_field" do
+      let(:node_spec) do
+        array_recursive_ostruct(
+          :metadata => {
+            :name              => "10.35.0.169",
+            :uid               => "6de77025-35f0-11e5-8917-001a4a5f4a00",
+            :resourceVersion   => "5302",
+            :creationTimestamp => "2015-07-29T12:50:45Z",
+            :labels            => {
+              :"kubernetes.io/hostname" => "10.35.0.169"
+            }
+          },
+          :spec     => {
+            :externalID => "10.35.0.169"
+          },
+          :status   => {
+            :capacity => {
+              :cpu    => "2",
+              :memory => capacity_memory,
+              :pods   => "40"
+            }
+          }
+        )
+      end
+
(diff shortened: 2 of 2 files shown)


agrare:
@cben do you know if there was a specific reason that parse_capacity_field didn't fall back to the standard decimal check that `parse_quantity` does?

cben:
Don't remember, I'll try to confirm with docs.
Can you add a test case will actual data that's failing?

agrare:
@cben :+1: tests added for both styles

cben:
Could any of these places contain fractions expressed as e.g. `300m` to represent 0.3?
`.decimal_si_to_f` does support this, will return Float instead of Integer.

The only value I know to actually take fractional milli values is cpu.

- [x] The places that now call `parse_capacity_field()` are all memory/storage, should be integers.

- [ ] We do have a place where we take `node.status.try(:capacity).try(:cpu)` without parsing!
https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/89de5b82c12e47bf632017a59d5fd79dd7e18c4e/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L709-L711
  For comparison, on one node I have `status.capacity.cpu: "4"` vs nearby `status.allocatable.cpu: "3500m"`.  Is `status.capacity.cpu` formally guaranteed to be whole?  No, this field is not individually documented.  Internal k8s representation is standard `ResourceList` which can contain milli fractions.  So theoretically this could explode one day.
  The `hardwares.cpu_total_cores` column which is `t.integer`, wouldn't take a float nor a string suffixed with "m".  So I think we'd get a crash bug report, and not trivial to fix proactively now.
  And certainly out of scope for this PR.

cben:
@agrare please add labels for backporting as desired.
The concrete example from BZ seems to be of PersistentVolumn storage_capacity becoming `nil`, and this line:
https://github.com/ManageIQ/manageiq-providers-kubernetes/blame/89de5b82c12e47bf632017a59d5fd79dd7e18c4e/app/models/manageiq/providers/kubernetes/container_manager/refresh_parser.rb#L849
goes back (with variations) to [2016](https://github.com/ManageIQ/manageiq-providers-kubernetes/commit/b90b99c41032bab4c92a44b22bdab8aad861d87a#diff-0324981fdb3019ce6d98f9c86d97f2bbL327), so it's a question of how far back we _want_ to patch...

agrare:
Thanks so much for digging into this @cben !

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 4522a459c8092ae43bed0797bfe6b8de5a96751b
Date:   Sun Dec 15 12:22:11 2019 +0200

    Merge pull request #352 from agrare/fix_capacity_field_parsing_of_decimal_values

    Fix parsing capacity fields of the form 50M or 1G

    (cherry picked from commit ecea23d8452803b1fcaeb91e7ced7ce100668442)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1785361
```

Hammer backport details:

```
$ git log -1
commit 50fd580553647c5f56123905f604b869372a977d
Date:   Sun Dec 15 12:22:11 2019 +0200

    Merge pull request #352 from agrare/fix_capacity_field_parsing_of_decimal_values

    Fix parsing capacity fields of the form 50M or 1G

    (cherry picked from commit ecea23d8452803b1fcaeb91e7ced7ce100668442)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1785358
```

PR #353, opened 2020-01-03
Subclass ContainerManager metrics capture

kbrock:
kubernetes is a container metrics capture class. subclassing the proper parent

part of ManageIQ/manageiq#19684


cben:
LGTM :+1:

Have you tested core with the provider change(s) but without https://github.com/ManageIQ/manageiq/pull/19684?  Sounds like it should be safe because currently both generic and specific `capture_ems_targets` both lead to `capture_container_targets`, but ruby const lookup always worries me :)

cben:
Do you want me to merge now, or waiting for more reviews / coordinated merge?

agrare:
@cben this looks safe to merge by itself, all of the functionality changes are in the core PR

PR #354, opened 2020-02-20
Lazy load kubeclient/image inspector library with autoload

jrafanie:
Note, there doesn't seem to be one single entrypoint for these classes and
mixins so when possible, we now autoload any constants being used in these
files.

Note, it's best to require or use ruby's autoload on libraries like this.
Normal rails models and other files that need to be reloaded by rails in dev
mode need to be loaded by using the constant directly to force const_missing
or use require_dependency.

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresh_mixin_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresh_mixin_spec.rb
@@ -1,5 +1,3 @@
-require 'kubeclient'
-
 describe ManageIQ::Providers::Kubernetes::ContainerManager::RefresherMixin do
   let(:client)  { double("client") }
   let(:ems)     { FactoryBot.create(:ems_kubernetes) }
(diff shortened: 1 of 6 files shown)


spec/models/manageiq/providers/kubernetes/container_manager/refresh_mixin_spec.rb
```diff
@@ -1,5 +1,3 @@
-require 'kubeclient'
```
jrafanie: this is done by the loading the RefreshMixin above.

cben:
Looks sane.

I'm out of touch.  Is there any central guide to how manageiq does loading?

jrafanie:
> I'm out of touch. Is there any central guide to how manageiq does loading?

@cben Not really.  As I'm going through providers and other repos and finding these library requires, I'm trying to find patterns and the only real pattern is "operations" (talking to a provider) vs. "domain objects" (the manageiq metadata about a provider, etc.).  We often mix these roles and therefore loading provider/managers often loads an "operations" type library like fog-kubevirt/rest-client/kubeclient/etc.  The more we can separate these concerns, the better.

We can't actually split these concerns now but delay loading "operations" libraries, can give us some of the benefits for now.

PR #355, opened 2020-03-04
Fix warning and strict partial failure

djberg96:
Refactoring PR that eliminates a constant redefinition warning. It also implements previously undefined methods either by including the `Vmdb::Logging` module (for the `_log` method) or by defining it directly (the `log_prefix` method). Without these, the specs fail with strict partial validation turned on.

There are other constants that should probably be replaced with `let` but this PR just focuses on fixes for now.

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
@@ -133,6 +133,8 @@ class MockFailedImageInspectorClient < MockImageInspectorClient
 end
 
 describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
+  let(:img_acq_err) { "can't find image" }
+
   context "SmartState Analysis Methods" do
     before(:each) do
       EvmSpecHelper.create_guid_miq_server_zone
@@ -521,7 +523,6 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
 
     context 'fetch_metadata fails to access pod' do
       it 'reports failure when fetch_metadata fails to access pod' do
-        IMG_ACQ_ERR = "can't find image".freeze
         allow_any_instance_of(described_class).to receive_messages(
           :image_inspector_client => MockCantAccessImageInspectorClient.new
         )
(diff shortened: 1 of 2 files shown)


djberg96:
@miq-bot add label test, refactoring

djberg96:
@miq-bot add_label test, refactoring

spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
```diff
@@ -133,6 +133,8 @@ def fetch_oscap_arf
 end
 
 describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job do
+  let(:img_acq_err) { "can't find image" }
```
agrare: Does this even need to be a let variable?  Could you just `expect(@job.message).to eq "image acquiring error: can't find image"`?

spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
```diff
@@ -535,15 +536,14 @@ def create_pod_definition
 
     context 'Image Acquiring Status' do
       it 'Detects when image acquiring failed and reports the error' do
-        IMG_ACQ_ERR = "can't find image".freeze
         allow_any_instance_of(described_class).to receive_messages(
-          :image_inspector_client => MockFailedImageInspectorClient.new("Sucess", "", IMG_ACQ_ERR, IMAGE_ID)
+          :image_inspector_client => MockFailedImageInspectorClient.new("Sucess", "", img_acq_err, IMAGE_ID)
```
cben:
```suggestion
          :image_inspector_client => MockFailedImageInspectorClient.new("Success", "", img_acq_err, IMAGE_ID)
```

cben: LGTM other than this.

spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
djberg96: @agrare ok, updated.

spec/models/manageiq/providers/kubernetes/container_manager/scanning/job_spec.rb
djberg96: @cben OK, updated.

PR #360, opened 2020-03-23
Cleanup watches refresh from topology

agrare:
Depends on: https://github.com/ManageIQ/manageiq-providers-openshift/pull/159


simaishi:

PR #362, opened 2020-03-25
Refactor graph refresh to match current standards

agrare:
Kubernetes graph refresh was one of the first added and didn't follow the standards from the rest of the providers (inventory/{collector,parser,persister}).

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/363

Cross repo tests: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/96


app/models/manageiq/providers/kubernetes/inventory/collector/container_manager.rb
```diff
@@ -1,49 +1,97 @@
...
   end
 
-  def service_catalog_connection
-    @service_catalog_connection ||= connect("kubernetes_service_catalog")
+  def kubernetes_entities
+    [
+      {:name => 'pods'}, {:name => 'services'}, {:name => 'replication_controllers'}, {:name => 'nodes'},
+      {:name => 'endpoints'}, {:name => 'namespaces'}, {:name => 'resource_quotas'}, {:name => 'limit_ranges'},
+      {:name => 'persistent_volumes'}, {:name => 'persistent_volume_claims'}
+    ].freeze
```
Fryguy: hard to tell if this was copied from somewhere but can this be one-per-line?

agrare: Yes it was copied and sure it can be a one-liner.  Am trying to keep the code as similar as possible during the "move"

agrare: Done

simaishi:

PR #363, opened 2020-03-26
Cleanup unused target_collection

agrare:
This is going to make refactoring the graph refresh code here easier and when we do watches in the future it likely won't be this way anyway.

Depends on: https://github.com/ManageIQ/manageiq-providers-openshift/pull/160


simaishi:

PR #364, opened 2020-03-26
Assorted graph refresh cleanups

agrare:


simaishi:

PR #366, opened 2020-04-09
Use watches for targeted refresh

agrare:
Use kubernetes watches for targeted refresh of ContainerManager entities

TODO:
- [x] Allow streaming refresh able to be disabled by settings
- [x] Restart watches if a user initiates a full refresh
- [x] Handle 410 Gone status (moving full refresh to be per entity would simplify watch restart)
- [x] Add specs covering all ADDED/MODIFIED/DELETED for all entity types
  - [x] pods
  - [x] services
  - [ ] ~~endpoints (These are weird because endpoints are used by services but not saved directly)~~ These are going to take more work, going to do a follow-up for this.  For now these will be updated every full refresh
  - [x] replication_controllers
  - [x] nodes
  - [x] namespaces
  - [x] resource_quotas
  - [x] limit_ranges
  - [x] persistent_volumes
  - [x] persistent_volume_claims
- [x] Targeted refresh parsing of a pod's container_image

Depends: 
- [x] https://github.com/ManageIQ/manageiq/pull/20098

https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/369

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/runner.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/runner.rb
@@ -19,22 +19,35 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::Runner <
   end
 
   def do_work
-    ensure_threads if streaming_refresh?
+    streaming_refresh? ? ensure_threads : stop_threads
 
     super
   end
 
-  def before_exit(_message, _exit_code)
-    return unless streaming_refresh?
+  def deliver_queue_message(msg)
+    # If we are using streaming refresh a user initiated full refresh should
+    # do a full refresh and restart the collector threads
+    if streaming_refresh?
+      super { restart_inventory_collector if full_refresh_queued?(msg) }
+    else
+      super
+    end
+  end
 
-    finish.make_true
-    stop_threads
+  def before_exit(_message, _exit_code)
+    stop_threads if streaming_refresh?
   end
 
   private
 
   attr_accessor :collector_threads, :refresher_thread
   attr_reader   :ems, :finish, :queue, :refresh_notice_threshold, :resource_version_by_entity
+
+  def restart_inventory_collector
+    stop_collector_threads
+    full_refresh
+    ensure_collector_threads
+  end
 
   def kubernetes_entity_types
     %w[pods replication_controllers nodes namespaces resource_quotas limit_ranges persistent_volumes persistent_volume_claims].freeze
@@ -87,6 +100,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::Runner <
   end
 
   def stop_threads
+    finish.make_true
     stop_collector_threads
     stop_refresher_thread
   end
@@ -100,11 +114,15 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::Runner <
   end
 
   def stop_refresher_thread
+    return unless refresher_thread&.alive?
+
     queue.push(nil) # Push a nil to unblock the refresher thread
     refresher_thread.join(10)
   end
 
   def refresher
+    _log.debug("Starting refresher thread")
+
     loop do
       notices = []
 
@@ -119,6 +137,8 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::Runner <
 
       partial_refresh(notices)
     end
+
+    _log.debug("Exiting refresher thread")
   end
 
   def ensure_collector_threads
@@ -135,6 +155,14 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::Runner <
 
   def stop_collector_threads
     collector_threads.each_value(&:stop!)
+  end
+
+  def refresh_queued?(msg)
+    msg.class_name == "EmsRefresh" && msg.method_name == "refresh"
+  end
+
+  def full_refresh_queued?(msg)
+    refresh_queued?(msg) && msg.data.any? { |klass, _id| klass == ems.class.name }
   end
 
   def streaming_refresh?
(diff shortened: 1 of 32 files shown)


app/models/manageiq/providers/kubernetes/inventory/parser/watch_notice.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def endpoints
+    # TODO: endpoints are only used for services
+  end
+
+  def replication_controllers
+    collector.replication_controllers.each do |notice|
+      replication_controller = notice.object
+
+      persister.container_repliation_controllers.targeted_scope << replication_controller.metadata.uid
```
chessbyte, line 38: typo: `container_repliation_controllers`

app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/runner.rb
```diff
@@ -1,2 +1,188 @@
...
+
+    @refresher_thread           = nil
+    @refresh_notice_threshold   = 100
+    @collector_threads          = {}
+    @resource_version_by_entity = {}
+  end
+
+  def do_before_work_loop
+    # Prime the entities' resourceVersions by performing an initial full refresh
+    # if we're using streaming refresh, otherwise queue an inital full the standard way
```
chessbyte, line 17: typo: `inital`

agrare: :+1:

app/models/manageiq/providers/kubernetes/inventory/collector/watch_notice.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def parse_notices!(all_notices)
+    notices_by_kind = all_notices.group_by { |notice| notice.object&.kind }.except(nil)
+    notices_by_kind.each do |kind, notices|
+      # The notices returned by the Kubernetes API contain always the complete representation of the object, so it isn't
+      # necessary to process all of them, only the last one for each object.
+      notices.reverse!.uniq! { |n| n.object&.metadata&.uid }
+
+      # Only add ADDED/MODIFIED to the collectors so deleted obejcts will be removed
```
chessbyte, line 35: typo: `obejcts`

agrare: :( I'm useless without spell-check

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -408,40 +385,59 @@ def node_computer_systems(parent, hash)
...
-    labels = parse_labels(service)
     new_result.merge!(
+      :container_project => lazy_find_project(:name => new_result[:namespace]),
       # TODO: We might want to change portal_ip to clusterIP
-      :portal_ip        => service.spec.clusterIP,
-      :session_affinity => service.spec.sessionAffinity,
-      :service_type     => service.spec.type,
-      :labels           => labels,
-      :tags             => map_labels('ContainerService', labels),
-      :selector_parts   => parse_selector_parts(service),
```
chessbyte, line 436: why are `:labels`, `:tags`, and `: selector_parts` not merged in the new code?

app/models/manageiq/providers/kubernetes/inventory/parser/watch_notice.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def persistent_volume_claims
+    collector.persistent_volume_claims.each do |notice|
+      pvc = notice.object
+
+      persister.persistent_volume_claims.targeted_scope << pvc.metadata.uid
+      next if notice.type == "DELETED"
+
+      parse_persistent_volume_claim(pvc)
+    end
```
chessbyte, line 114: The methods in this file look like prime candidates for a refactoring, as they are all mostly copy/paste with minor changes in each.  Perhaps, we can call a helper method with enough parameters so it makes the calls dynamically.

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
agrare: They were added to the `new_result` hash so that they could be saved by legacy refresh, or pulled out and saved by graph refresh.  There isn't any reason to add them to the hash just to extract them later now that we only use graph refresh.

chessbyte: Thanks for the explanation.

app/models/manageiq/providers/kubernetes/inventory/parser/watch_notice.rb
agrare: A lot of these could be, these are all of the "easy" collections though where all of the manager_refs happen to be `metadata.uid` and no other handling has to be done (like will be necessary for endpoints/services here and projects/namespaces/images on the openshift side).

I'd like to follow up with refactoring after I get a clearer picture of how the openshift parser will inherit from this and handle some of those issues unless you'd like that done for this PR.

chessbyte: future refactoring is fine

agrare: Actually it should be better now, I'm parsing all of the notices together and using the base `Parser::ContainerManager` instead of overriding each entity parser

agrare:
Depends on backport of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/370 and https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/371

agrare:
@simaishi https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/374 backports this and dependent PRs as well as disabling the streaming refresh setting by default for the jansa branch 
 while leaving it on for kasparov

simaishi:

PR #370, opened 2020-04-21
Refactor Parser::ContainerManager

agrare:
The ContainerManager Parser was written to allow for the core parsing to
be shared with graph and legacy refresh which is why a number of items
are added to the hash then deleted and handled differently.

This can all be cleaned up now that we only use graph refresh.


simaishi:

PR #371, opened 2020-04-21
Move no-longer-shared specs into main context

agrare:
The main body of refresh specs were shared between classic and graph
refresh contexts.

Now that there is no classic refresh being tested we no longer need the
shared context.


cben:
LGTM

simaishi:

PR #372, opened 2020-04-30
Add targeted refresh for services and endpoints

agrare:
Services and endpoints are different k8s collections but are combined into one model in the MIQ refresh parser.  This means we cannot parse these as standalone notices like the other collections.

If we get a notice about either an endpoint or a service, perform a get_ for the other to allow the parser to work normally.

Followup to https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/366

### DIFF
--- a/app/models/manageiq/providers/kubernetes/inventory/collector.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/collector.rb
@@ -1,4 +1,14 @@
 class ManageIQ::Providers::Kubernetes::Inventory::Collector < ManageIQ::Providers::Inventory::Collector
   require_nested :ContainerManager
   require_nested :WatchNotice
+
+  def connect(service)
+    manager.connect(:service => service)
+  rescue KubeException
+    nil
+  end
+
+  def kubernetes_connection
+    @kubernetes_connection ||= connect("kubernetes")
+  end
 end
--- a/app/models/manageiq/providers/kubernetes/inventory/collector/watch_notice.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/collector/watch_notice.rb
@@ -1,5 +1,5 @@
 class ManageIQ::Providers::Kubernetes::Inventory::Collector::WatchNotice < ManageIQ::Providers::Kubernetes::Inventory::Collector
-  attr_reader :additional_attributes, :pods, :services, :replication_controllers,
+  attr_reader :additional_attributes, :pods, :replication_controllers,
               :namespaces, :nodes, :notices, :resource_quotas, :limit_ranges,
               :persistent_volumes, :persistent_volume_claims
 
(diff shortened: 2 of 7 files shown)


app/models/manageiq/providers/kubernetes/inventory/collector.rb
```diff
@@ -1,4 +1,14 @@
...
+
+  def connect(service)
+    manager.connect(:service => service)
+  rescue KubeException
+    nil
+  end
+
+  def kubernetes_connection
+    @kubernetes_connection ||= connect("kubernetes")
+  end
```
agrare, line 13: Moved this up to the Collector base class so it can be shared by `Collector::ContainerManager` and `Collector::WatchNotice`

app/models/manageiq/providers/kubernetes/inventory/collector/watch_notice.rb
```diff
@@ -12,13 +12,46 @@ def initialize(manager, notices)
...
+  # the same name and namespace, and vice versa
+  def endpoints
+    return @endpoints if @endpoints.any?
+
+    @endpoints = @services.each_with_object([]) do |service, results|
+      name      = service.metadata.name
+      namespace = service.metadata.namespace
+
+      results << kubernetes_connection.get_endpoint(name, namespace)
+    rescue Kubeclient::ResourceNotFoundError
```
agrare, line 30: Not all services have endpoints so we have to handle 404

app/models/manageiq/providers/kubernetes/inventory/collector/watch_notice.rb
```diff
@@ -12,13 +12,46 @@ def initialize(manager, notices)
...
 
+  # Endpoints and services come from two different watches but are
+  # merged into one model in manageiq.  This means we have to watch
+  # for updates to both entity kinds and if we receive an update
+  # we have to get the current state of the other one.
+  #
+  # If we get an endpoint notice we have to get the service with
+  # the same name and namespace, and vice versa
+  def endpoints
+    return @endpoints if @endpoints.any?
```
agrare, line 23: Using `return if any?` instead of `@endpoints ||=` because if there aren't any endpoint or service notices the parser is expecting `[]`

app/models/manageiq/providers/kubernetes/inventory/collector/watch_notice.rb
```diff
@@ -12,13 +12,46 @@ def initialize(manager, notices)
...
+
+    @endpoints = @services.each_with_object([]) do |service, results|
```
agrare, line 25: Need to use `@services` here not `services` because we want to look only at notices in this targeted refresh not go out and fetch any services

app/models/manageiq/providers/kubernetes/inventory/parser/watch_notice.rb
```diff
@@ -18,12 +18,11 @@ def parse_notices
       object = notice.object
       kind   = object.kind
 
+      collection_name = resource_by_entity(kind.underscore)
+      next if collection_name.nil?
+
```
agrare, line 23: Endpoints don't have an associated resource so we need to skip them here.  Any deletes will be handled by the Service entity notice.

app/models/manageiq/providers/kubernetes/inventory/parser/watch_notice.rb
```diff
@@ -18,12 +18,11 @@ def parse_notices
...
       inventory_collection = persister.send(resource_by_entity(kind.underscore).tableize)
       inventory_collection.targeted_scope << object.metadata.uid
     end
   end
-
-  def cgs_by_namespace_and_name
-    nil
-  end
```
agrare, line 28: Now that we're properly collecting endpoints we don't need to disable container_groups for services in targeted refresh

simaishi:
Jansa backport details:

```
$ git log -1
commit 7b81bcee15950b9c7e88c02dc52ea282586641cc
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Mon May 4 09:41:07 2020 -0400

    Merge pull request #372 from agrare/collect_endpoints_when_targeting_services

    Add targeted refresh for services and endpoints

    (cherry picked from commit 777a8de1599492a08af0d9176a96a41ceec02d21)
```

PR #385, opened 2020-06-29
Adjusted DDF schema and credentials validation

skateman:
In order to make this form work, I had to do some tradeoffs:
* The `default_value_for` in the `ContainerManagerMixin` prevents the usage of `create_from_params` so I deleted it.
* We are (kinda) using the same authentication for each endpoint (except kubevirt) by copying the same authentication with different authtypes. I solved this via redefining the `.create_from_params` with the replication of the same data among endpoints. However, I'd rather see the provider using just the single authentication instead.
* As we never send down passwords to the client and the token authentication has no any public value that we could send down, we end up with not submitting anything in the `authentications` if there's no token change. This can cause all authentications to be deleted upon an edit without a token change. In order to avoid this, I redefined the `#edit_with_params` to make sure that all `authentications` exist for all the `endpoints`. Again, this would be much easier if we could just use a single auth instead of copying the existing one N times.
* The endpoint for the kubevirt provider is the same as the default one, but it uses a different token. On the old forms this was sorted out by copying the data on the form level, which is not possible via DDF. For now the user has to fill these fields manually, but I hope that the `Detect` button will be able to copy these values on-demand.
* The `Detect` button doesn't work yet, there's [no API endpoint for it](https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/391) ~~and there's a [styling issue](https://github.com/ManageIQ/manageiq-ui-classic/issues/7167) that I'm trying to fix~~
* ~~Validation is only working for the default endpoint and partially for kubevirt. The other endpoints need to have [context-free methods](://github.com/ManageIQ/manageiq-providers-kubernetes/issues/387) than can work with the params I am sending. Kubevirt is using the same validation as the default endpoint, but it doesn't actually detect if the given token allows access to the kubevirt features.~~

Parent issue: https://github.com/ManageIQ/manageiq/issues/18818

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -653,7 +653,9 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
       case endpoint_name
       when 'default', 'kubevirt' # this also (partially) validates kubevirt
         !!raw_connect(hostname, port, options)
-      when 'prometheus', 'prometheus_alerts', 'hawkular'
+      when 'hawkular'
+        verify_hawkular_credentials(hostname, port, options)
+      when 'prometheus', 'prometheus_alerts'
         # TODO: implement validation calls for these endpoint types
         return true
       else
@@ -722,6 +724,27 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
 
     def service_catalog_api_version
       'v1beta1'
+    end
+
+    def hawkular_connect(hostname, port, options)
+      require "hawkular/hawkular_client"
+
+      uri = raw_api_endpoint(hostname, port, options[:path] || "/hawkular/metrics")
+      credentials = {:token => options[:bearer]}
+      ssl_options = options[:ssl_options] || {:verify_ssl => OpenSSL::SSL::VERIFY_NONE}
+
+      hawkular_options = {
+        :http_proxy_uri => options[:http_proxy] || VMDB::Util.http_proxy_uri,
+        :tenant         => options[:tenant] || '_system',
+        :verify_ssl     => ssl_options[:verify_ssl],
+        :ssl_cert_store => ssl_options[:ca_file],
+      }
+
+      Hawkular::Metrics::Client.new(uri, credentials, hawkular_options)
+    end
+
+    def verify_hawkular_credentials(hostname, port, options)
+      !!hawkular_connect(hostname, port, options)&.avail&.get_data('all', :limit => 1)&.kind_of?(Array)
     end
   end
 
--- a/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
@@ -8,13 +8,6 @@ module ManageIQ::Providers::Kubernetes::ContainerManagerMixin
   METRICS_ROLES = %w(prometheus hawkular).freeze
 
   included do
-    default_value_for :port do |provider|
-      # port is not a column on this table, it's delegated to endpoint.
-      # This may confuse `default_value_for` to apply when we do have a port;
-      # checking `provider.port` first prevents this from overriding it.
-      provider.port || provider.class::DEFAULT_PORT
-    end
-
     supports :streaming_refresh do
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -38,77 +31,582 @@ module ClassMethods
...
     def verify_credentials(args)
-      default_endpoint = args.dig("endpoints", "default")
-      hostname, port = default_endpoint&.values_at("hostname", "port")
+      endpoint_name = args.dig("endpoints").keys.first
+      endpoint = args.dig("endpoints", endpoint_name)
 
-      options = default_endpoint&.slice("path", "http_proxy")&.symbolize_keys || {}
+      token = args.dig("authentications", "bearer", "auth_key") || args.dig("authentications", "kubevirt", "auth_key")
+      token = MiqPassword.try_decrypt(token)
+      token ||= find(args["id"]).auth_user_token(endpoint_name) if args["id"]
```
skateman, line 575: @agrare this line should kick in when we edit a provider as the password was not sent down to the client. The problem with this method is that it [doesn't work](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/authentication_mixin.rb#L115) for container providers and I suspect it never worked because the `userid` is `nil` in any existing database I'm playing with. Also the return value doesn't really correspond with the method name. Anyway, I need a way to extract the token from the DB somehow ...

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -163,6 +661,27 @@ def service_catalog_api_version
...
+    # super() below. In order to prevent this, the authentications are set to a dummy value if the related
+    # endpoint exists among the submitted data.
+    endpoints.each do |endpoint|
+      case endpoints['role']
+      when 'default' # The default endpoint is paired with the bearer authentication
+        bearer ? nil : authentications << {'authtype' => 'bearer'}
+      when 'kubevirt' # Kubevirt has its own authentication, no need for replication
+        kubevirt ? nil : authentications << {'authtype' => 'kubevirt'}
+      else # Replicate the bearer authentication for any other endpoints
+        authentications << {'authtype' => endpoint['role']}.reverse_merge(bearer || {})
```
skateman, line 678: The `{}` is probably wrong at the end as it won't replicate the token properly if we're editing a provider to add a new endpoint. I could possibly retrieve the token from the DB, but I don't have *The Right Way&trade;* to do it as described above.

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -38,77 +31,651 @@ module ClassMethods
...
+      when 'default', 'kubevirt' # this also (partially) validates kubevirt
+        !!raw_connect(hostname, port, options)
+      when 'hawkular'
+        verify_hawkular_credentials(hostname, port, options)
+      when 'prometheus'
+        verify_prometheus_credentials(hostname, port, options)
+      when 'prometheus_alerts'
+        verify_prometheus_alerts_credentials(hostname, port, options)
+      else
+        # TODO: maybe we need an error message here
```
skateman, line 663: Do we need a special error message if something unexpected comes here? @agrare

app/models/manageiq/providers/kubernetes/container_manager_mixin.rb
```diff
@@ -38,77 +31,651 @@ module ClassMethods
...
+        :bearer      => token,
+        :ssl_options => {
+          :verify_ssl => endpoint&.dig("security_protocol") == 'ssl-without-validation' ? OpenSSL::SSL::VERIFY_NONE : OpenSSL::SSL::VERIFY_PEER,
+          :ca_file    => endpoint&.dig("certificate_authority")
+        }
       }
 
-      !!raw_connect(hostname, port, options)
+      case endpoint_name
+      when 'default', 'kubevirt' # this also (partially) validates kubevirt
```
skateman, line 654: Is this enough for kubevirt? It doesn't really verifies if the token allows access to it.

agrare: :+1: my bad I thought you said this was validating kubevirt sufficiently (I should have checked), added specific kubevirt validation

skateman:
I guess this is good to go then :+1:

agrare:
@gtanzillo can you review since I have some commits in here now?

PR #388, opened 2020-07-02
Fix "private method 'watch' called for WatchThread"

agrare:
Starting the watch threads assigns the watch thread to a self.thread instance variable, this is throwing a private method exception on older rubies (I don't see this on 2.7)

```
ERROR -- : [NoMethodError]: private method 'watch' called for #<ManageIQ::Providers::Openshift::ContainerManager::RefreshWorker::WatchThread:0x0000558b852f8af8>
ERROR -- : app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb:43:in 'collector_thread'
app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb:23:in 'block in start!'
```

This isn't the default on jansa so it isn't critical to be backported, but a user could change the settings and enable it.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
@@ -31,10 +31,13 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThr
     thread&.join(join_limit)
   end
 
+  protected
+
+  attr_accessor :resource_version, :thread, :watch
+
   private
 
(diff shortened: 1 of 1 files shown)


Fryguy:
Why are we making them all public, as opposed to just the one?

agrare:
Just out of an abundance of caution since I wasn't sure what was wrong, but I think the issue is these need to be protected instead of private

Fryguy:
I guess what I'm wondering is why all of them as opposed to just the :thread accessor.  Or have we hit this with multiple methods?

agrare:
Because the same thing would happen for any of the attr_accessors, I'll update to just make those protected and leave the attr_readers private.

PR #389, opened 2020-07-15
PERF_ROLLUP_CHILDREN should be an Array

bdunne:
The only callers then call `#to_miq_a`

https://github.com/ManageIQ/manageiq/blob/33d994ae207105a508d9e214d39d9ec66ac066dd/app/models/metric/rollup.rb#L189
https://github.com/ManageIQ/manageiq/blob/33d994ae207105a508d9e214d39d9ec66ac066dd/app/models/metric/rollup.rb#L204


bdunne:
@miq-bot assign Fryguy

cben:
Nice cleanup, LGTM

PR #395, opened 2020-09-03
Bump hawkular-client to 5.0.0

agrare:
Hawkular-client 5.0.0 opens up the rest-client version requirement allowing for newer rest-clients to be used.


manageiq-providers-kubernetes.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency("hawkular-client",                 "~> 4.1")
+  s.add_runtime_dependency("hawkular-client",                 "= 5.0.0.pre2")
```
agrare, line 16: This is just temporary until 5.0.0 is released

PR #396, opened 2020-09-09
Fix bug where settings changes are not reflected in the list

Fryguy:
Since we were memoizing a dynamic value, subsequent changes were not reflected.

Depends on https://github.com/ManageIQ/manageiq/pull/20533


Fryguy:
At @skateman's recommendation, I changes this to just no longer memoize (diff is much easier to read with [whitespace disabled](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/396/files?w=1)).  I still need DDF.find_field for tests.

agrare (approved):

agrare:
We do not use DDF forms for provider creation on jansa so this isn't necessary

PR #398, opened 2020-09-11
Plugin Generator Updates

Fryguy:
Plugin Generator Updates

### DIFF
--- a/manageiq-providers-kubernetes.gemspec
+++ b/manageiq-providers-kubernetes.gemspec
@@ -1,26 +1,30 @@
-$:.push File.expand_path("../lib", __FILE__)
+# coding: utf-8
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'manageiq/providers/kubernetes/version'
 
-require "manageiq/providers/kubernetes/version"
+Gem::Specification.new do |spec|
+  spec.name          = "manageiq-providers-kubernetes"
+  spec.version       = ManageIQ::Providers::Kubernetes::VERSION
+  spec.authors       = ["ManageIQ Authors"]
 
-Gem::Specification.new do |s|
-  s.name        = "manageiq-providers-kubernetes"
-  s.version     = ManageIQ::Providers::Kubernetes::VERSION
-  s.authors     = ["ManageIQ Developers"]
-  s.homepage    = "https://github.com/ManageIQ/manageiq-providers-kubernetes"
-  s.summary     = "Kubernetes Provider for ManageIQ"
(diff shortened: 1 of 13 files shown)


manageiq-providers-kubernetes.gemspec
```diff
@@ -1,26 +1,30 @@
...
+  spec.add_dependency "image-inspector-client",          "~> 2.0"
+  spec.add_dependency "kubeclient",                      "~> 4.6"
+  spec.add_dependency "more_core_extensions",            ">= 3.6", "< 5"
+  spec.add_dependency "prometheus-alert-buffer-client",  "~> 0.3.0"
+  spec.add_dependency "prometheus-api-client",           "~> 0.6"
 
-  s.add_development_dependency("codeclimate-test-reporter", "~> 1.0.0")
-  s.add_development_dependency("recursive-open-struct",     "~> 1.0.0")
-  s.add_development_dependency("simplecov")
+  spec.add_development_dependency "recursive-open-struct", "~> 1.0.0"
```
Fryguy, line 28: I find it really weird that recursive-open-struct is a dev dependency.  Can we just remove this?  kubeclient already [brings in this dependency](https://github.com/abonas/kubeclient/blob/e3b170f9e3ec79920f5fa93b721f1a3e62c6d782/kubeclient.gemspec#L36), and presumably we only use it in specs because kubeclient is expecting it?

agrare: Looks like this came from https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/16 

>  Since we use this heavily in tests here, we should state our reliance on this gem.

^ is why it is a dev dependency

However we only use this heavily in specs because that is what kubeclient uses under the hood, so if kubeclient changed that in the future our specs would have to change as well so calling this out separately from kubeclient isn't that helpful IMO

:+1: for removing it

agrare (approved):

manageiq-providers-kubernetes.gemspec
cben: To clarify, these are not "inputs kubeclient is expecting".  We `require 'recursive-open-struct'` in tests of provider logic where we feed input "like what kubeclient would output" without actually using kubeclient.
Personally I like the rule of thumb "depend on everything we `require` directly" just because it's a simple way to decide :grin:.

However Adam is not wrong.  Hypothetically, if kubeclient switched a different implementation with compatible contract, the direct dependency would allow these tests to remain unmodified, but probably we _want_ them to break so we'll update :wink:
Also, kubeclient will never change its default. And anyway we don't want another similar `.foo.bar` format, if anything we better opt-in to `as: :parsed_symbolized` mode and rewrite access to `[:foo][:bar]`...

Bottom line: I'm ok either way.

PR #399, opened 2020-09-17
Use the tag_mapper from the core Persister

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/20561


cben:
Nice refactoring :clap:

Issue #412, opened 2020-11-16
Consistency Issue?  "NO Proxy" vs "No Proxy"


chessbyte:
Are these the same for i18n purposes?  Should they be changed to be the same string?

[No Proxy](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager_mixin.rb#L646) vs [NO Proxy](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/options.rb#L30)

cben:
yes, definitely same.

PR #414, opened 2020-11-23
Consistent usage of "No Proxy" for internationalization

agrare:
Fixes https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/412

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/options.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/options.rb
@@ -27,8 +27,8 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::Options
               :help_text => N_('HTTPS Proxy to connect image inspector pods to the internet. example: https://user:password@my_https_proxy'),
             },
             :no_proxy    => {
-              :label     => N_('NO Proxy'),
-              :help_text => N_('NO Proxy lists urls that should\'nt be sent to any proxy. example: my_file_server.org'),
+              :label     => N_('No Proxy'),
+              :help_text => N_('No Proxy lists urls that should\'nt be sent to any proxy. example: my_file_server.org'),
             },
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/container_manager/options.rb
```diff
@@ -27,8 +27,8 @@ def advanced_settings
               :help_text => N_('HTTPS Proxy to connect image inspector pods to the internet. example: https://user:password@my_https_proxy'),
             },
             :no_proxy    => {
-              :label     => N_('NO Proxy'),
-              :help_text => N_('NO Proxy lists urls that should\'nt be sent to any proxy. example: my_file_server.org'),
+              :label     => N_('No Proxy'),
+              :help_text => N_('No Proxy lists urls that should\'nt be sent to any proxy. example: my_file_server.org'),
```
chessbyte, line 31: should the help text be simplified to: `URLs that should not be sent to any proxy.  example: my_file_server.org`?  The double negative (No Proxy ... shouldn't) are not clear, especially for a help text.  If we need the "No Proxy" wording in the help text, then maybe something like "The No Proxy option/setting lists URLs ..."

cben:
At least s/should/will/ to clarify its what the field means, not instruction to user.
Maybe "will bypass proxies" or "will ignore proxy settings"?

simaishi:
Kasparov backport details:

```
$ git log -1
commit 1c6fd425472800b8d549ede4a7f56c1bf7b3ac98
Date:   Mon Nov 23 12:20:13 2020 -0500

    Merge pull request #415 from agrare/cve_location_consistency

    Consistent usage of CVE Location for i18n

    (cherry picked from commit 66e4bf08fffce6d1b893947e335a8295890c6632)
```

PR #416, opened 2020-12-15
Mark unused variable, remove redundant safe navigation

djberg96:
The rubocop linter picked up a couple things. Specifically, there's no need to do safe navigation on a `kind_of?` check since `NilClass` implements it.

I also marked the `pod_id` variable in `parse_container_status` as unused. I can remove it, but wasn't sure if the second argument was there to comply with an interface.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
@@ -840,7 +840,7 @@ class ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager < Man
     end
   end
 
-  def parse_container_status(container, pod_id)
+  def parse_container_status(container, _pod_id)
     container_image = parse_container_image(container.image, container.imageID)
     return if container_image.nil?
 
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -840,7 +840,7 @@ def parse_quantity_decimal(value)
     end
   end
 
-  def parse_container_status(container, pod_id)
+  def parse_container_status(container, _pod_id)
```
agrare, line 843: @djberg96 shouldn't we just remove the unused variable?  I believe this is only called in one place in this class

djberg96: I'll try it, but I'm gonna do some cross-repo tests as well, just to make sure.

djberg96:
@miq-bot cross_repo_tests manageiq, manageiq-api, manageiq-ui-classic

agrare:
@djberg96 `manageiq, manageiq-api, manageiq-ui-classic` doesn't make a whole lot of sense for a cross repo test of this change.  This is a method in kubernetes refresh so if anything I think a cross-repo test with `manageiq-providers-openshift` would be the one to do.

djberg96:
@miq-bot cross_repo_tests manageiq-providers-openshift

agrare (approved):

PR #418, opened 2021-01-15
Move logger initialization and configuration into plugins

Fryguy:
Move logger initialization and configuration into plugins

### DIFF
--- a/lib/manageiq/providers/kubernetes/engine.rb
+++ b/lib/manageiq/providers/kubernetes/engine.rb
@@ -13,16 +13,6 @@ module ManageIQ
         def self.plugin_name
           _('Kubernetes Provider')
         end
-
-        def self.init_loggers
-          $kube_log ||= Vmdb::Loggers.create_logger("kube.log")
-          $cn_monitoring_log ||= Vmdb::Loggers.create_logger("cn_monitoring.log")
-        end
-
-        def self.apply_logger_config(config)
-          Vmdb::Loggers.apply_config_value(config, $kube_log, :level_kube)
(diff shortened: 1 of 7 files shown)


lib/manageiq/providers/kubernetes/engine.rb
```diff
@@ -13,6 +13,16 @@ def self.vmdb_plugin?
         def self.plugin_name
           _('Kubernetes Provider')
         end
+
+        def self.init_loggers
+          $kube_log ||= Vmdb::Loggers.create_logger("kube.log")
```
agrare, line 18: `$kube_log` is only used by https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/kubernetes_event_monitor.rb#L36
That to me looks like it should be a `_log`

lib/manageiq/providers/kubernetes/engine.rb
```diff
@@ -13,6 +13,16 @@ def self.vmdb_plugin?
...
+          $cn_monitoring_log ||= Vmdb::Loggers.create_logger("cn_monitoring.log")
```
agrare, line 19: `$cn_monitoring_log` is also only used in the MonitoringManager::EventCatcher https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/monitoring_manager/event_catcher/runner_mixin.rb#L19-L21
Maybe we don't need this either?

agrare (approved):

PR #422, opened 2021-02-03
Fix instance connect options for username/password

agrare:
The connect options for the class-level take `options[:username]` and `options[:password]` but the instance connect passes `options[:user]` and `options[:pass]`

These [`connect_options`](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/fc33713d32bcaa3a0b345aaa5f19dab1865c929a/app/models/manageiq/providers/kubernetes/container_manager.rb#L998-L1011) are passed directly to [`.raw_connect`](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/fc33713d32bcaa3a0b345aaa5f19dab1865c929a/app/models/manageiq/providers/kubernetes/container_manager.rb#L995) which is just [`kubernetes_connect`](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/fc33713d32bcaa3a0b345aaa5f19dab1865c929a/app/models/manageiq/providers/kubernetes/container_manager.rb#L58-L60) and none of the [openshift methods](https://github.com/ManageIQ/manageiq-providers-openshift/blob/master/app/models/manageiq/providers/openshift/container_manager.rb#L48-L67) deal with user/password.  Kubernetes connect call [`kubernetes_auth_options`](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/fc33713d32bcaa3a0b345aaa5f19dab1865c929a/app/models/manageiq/providers/kubernetes/container_manager.rb#L811)  which checks for and sets `:username` and `:password` not `:user` or `:password` [[ref]](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/fc33713d32bcaa3a0b345aaa5f19dab1865c929a/app/models/manageiq/providers/kubernetes/container_manager.rb#L828-L831)

Required for: https://github.com/ManageIQ/manageiq-providers-amazon/pull/679

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -999,8 +999,8 @@ Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
     options.merge(
       :hostname    => options[:hostname] || address,
       :port        => options[:port] || port,
-      :user        => options[:user] || authentication_userid(options[:auth_type]),
-      :pass        => options[:pass] || authentication_password(options[:auth_type]),
+      :username    => options[:username] || authentication_userid(options[:auth_type]),
+      :password    => options[:password] || authentication_password(options[:auth_type]),
       :bearer      => options[:bearer] || authentication_token(options[:auth_type] || 'bearer'),
       :http_proxy  => self.options ? self.options.fetch_path(:proxy_settings, :http_proxy) : nil,
       :ssl_options => options[:ssl_options] || {


agrare:
I still need to make sure this isn't used anywhere, I think we use bearer tokens so universally this might have been broken for a long time though

agrare:
Okay, so I _think_ that this has been broken since https://github.com/ManageIQ/manageiq/pull/3056 because it replaced `kubernetes_connect(hostname, port, username, password)` with `kubernetes_connect(hostname, port, options)` [[ref]](https://github.com/ManageIQ/manageiq/pull/3056/files#diff-5c0bd1b3f52404d88b82ffef818b33999d8ee73effff42dcd11215f4c9d635d9L20) and it uses `options[:username]` and `options[:password]` [[ref]](https://github.com/ManageIQ/manageiq/pull/3056/files#diff-5c0bd1b3f52404d88b82ffef818b33999d8ee73effff42dcd11215f4c9d635d9R26) but it sets `options[:user]` and `options[:pass]` [[ref]](https://github.com/ManageIQ/manageiq/pull/3056/files#diff-5c0bd1b3f52404d88b82ffef818b33999d8ee73effff42dcd11215f4c9d635d9R48-R49)

chessbyte (approved):

PR #424, opened 2021-02-04
Skip refresh if no notices left after filtering

agrare:
Add a filter_notices method that can be subclassed by other providers and skip the refresh if there are no filtered notices left.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
@@ -55,6 +55,8 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThr
           break
         end
 
+        next if noop?(notice)
+
         queue.push(notice)
       end
 
@@ -68,6 +70,11 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThr
   rescue => err
     _log.error("Watch thread for #{entity_type} failed: #{err}")
     _log.log_backtrace(err)
+  end
+
+  def noop?(_notice)
+    # Allow sub-classes to filter notices
+    false
   end
 
   def connection(_entity_type = nil)


Fryguy:
Should this somehow be combined with the existing filter_notices code

 https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/23faaebed5f197c09592a8fff7094f844c09d12a/app/models/manageiq/providers/kubernetes/inventory/collector/container_manager/watch_notice.rb#L56-L66

app/models/manageiq/providers/kubernetes/inventory/collector/container_manager/watch_notice.rb
```diff
@@ -4,7 +4,7 @@ class ManageIQ::Providers::Kubernetes::Inventory::Collector::ContainerManager::W
               :persistent_volumes, :persistent_volume_claims
 
   def initialize(manager, notices)
-    @notices = filter_notices(notices)
+    @notices = notices
```
Fryguy, line 7: My only concern is here is that there are other callers to WatchNotice that might pass in unfiltered notices.  Are all callers covered?  We might want a comment here explainging that filtering needs to happen by the caller (or perhaps that's overkill?)

Aside, what is the purpose of this class?  The name represents a concrete thing, a WatchNotice (i.e. a concrete noun), but it derives from Collector (a proper noun more or less), which is strange.

Fryguy: As discussed, we may want to rename WatchNotice to Streaming/StreamingCollector/Streamer, but not for this PR.

Additionally, we should think about keeping the filtering logic in this class, but exposing it as a class method...something like a "WatchNotice.preprocess_notices".

agrare: > Aside, what is the purpose of this class? The name represents a concrete thing, a WatchNotice

The collector/parser all act on a concrete thing.  The `Parser::ContainerManager` parses a container manager and the `Parser::WatchNotice` parses watch notices.  The Collector exposes the notices in a consistent way to the Parser, it is initialized with notices from the runner.

I'm okay renaming this but would rather do that in a different PR set since it isn't really related to this change and will require PRs to all subclasses

agrare: Came up with a less invasive approach

PR #432, opened 2021-04-23
Use pluggable provider secrets

agrare:
https://github.com/ManageIQ/manageiq/issues/21157


spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
```diff
@@ -6,8 +6,7 @@
   let!(:ems) do
     FactoryBot.create(
       :ems_kubernetes_with_zone,
-      :hostname        => "10.35.0.169",
```
agrare, line 9: So this wasn't really "secret" because it is just from a local minishift env but figured it'd be nicer to not have "hard coded" addresses that have to be changed later anyway

PR #434, opened 2021-05-06
Handle a null byte in Node systemUUID

agrare:
If the Node systemUUID contains a null byte the PostgreSQL query throws an exception

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
@@ -724,6 +724,10 @@ describe ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager do
                                  :ext_management_system => @ems,
                                  :name                  => "instance_id")
       end
+
+      it "fails when there is a NULL byte in the providerID" do
+        @node[:identity_system] = "UUID\u0000"
+      end
     end
 
     context "succesful attempts" do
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
```diff
@@ -724,6 +724,10 @@
                                  :ext_management_system => @ems,
                                  :name                  => "instance_id")
       end
+
+      it "fails when there is a NULL byte in the providerID" do
+        @node[:identity_system] = "UUID\u0000"
```
Fryguy, line 729: aside, this looks so strange - like the test isn't testing anything...testing in the after block feels like a smell.

Fryguy:
Does this have to go back to lasker?

spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
agrare: Yeah I agree I think it is their idea of "dry tests" but it'd be more obvious if the testes were just in a method that was called from each `it`

agrare:
This was found on Kasparov on an openshift cluster on ppc64le

Fryguy:
Backported to `lasker` in commit 9d4e9f16f4fca3b52d3dc13e65d80c43eff010f2.

```text
commit 9d4e9f16f4fca3b52d3dc13e65d80c43eff010f2
Date:   Thu May 6 18:04:39 2021 -0400

    Merge pull request #434 from agrare/fix_node_cross_link_null_byte
    
    Handle a null byte in Node systemUUID
    
    (cherry picked from commit 9a68d4567cb1e0e003bef30674ec858039cfa051)
```

PR #435, opened 2021-05-18
Add supports :metrics and check for metrics endpoint

agrare:
https://github.com/ManageIQ/manageiq/pull/21224

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -33,6 +33,10 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
     )
   end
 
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -33,6 +33,10 @@ def monitoring_manager_needed?
     )
   end
 
+  supports :metrics do
+    unsupported_reason_add(:metrics, _("No metrics endpoint has been added")) unless supports_metrics?
+  end
+
   def supports_metrics?
```
agrare, line 40: Not sure if this was supposed to override the old supports_feature_mixin `supports_#{feature}?` method, need to see where this is used and if we can use `supports?(:metrics)` instead

agrare:
cc @Fryguy

PR #436, opened 2021-05-20
Strip any null bytes from systemUUID in node hash

agrare:
When K8s is installed on power it has been seen that the nodes' systemUUID property contains null bytes.  This was causing the cross_link_node query to fail but once we get past that the save_inventory also fails.

Strip null bytes before saving the node hash to fix:
```
[----] I, [2021-05-20T15:17:49.836289 #456047:2ab88a5f7964]  INFO -- : EMS: [OCPpower], id: [16] Saving EMS Inventory...
[----] E, [2021-05-20T15:17:49.893094 #456047:2ab88a5f7964] ERROR -- : Error when saving InventoryCollection:<ContainerNode> with strategy: , saver_strategy: batch, targeted: false. Message: string contains null byte
[----] E, [2021-05-20T15:17:49.893391 #456047:2ab88a5f7964] ERROR -- : MIQ(ManageIQ::Providers::Openshift::ContainerManager::RefreshWorker::Runner#refresh_block) EMS [OCPpower], id: [16] Refresh failed: string contains null byte
[----] E, [2021-05-20T15:17:49.893593 #456047:2ab88a5f7964] ERROR -- : [ArgumentError]: string contains null byte  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2021-05-20T15:17:49.893791 #456047:2ab88a5f7964] ERROR -- : /opt/IBM/infrastructure-management-gemset/gems/activerecord-5.2.4.4/lib/active_record/connection_adapters/postgresql/quoting.rb:21:in `escape_string'
```

### DIFF
--- a/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
@@ -293,7 +293,7 @@ class ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager < Man
   def cross_link_node(new_result)
     # Establish a relationship between this node and the vm it is on (if it is in the system)
     provider_id = new_result[:identity_infra]
-    bios_uuid   = new_result[:identity_system]&.gsub("\u0000", "")&.downcase
+    bios_uuid   = new_result[:identity_system]&.downcase
 
     host_instance   = find_host_by_provider_id(provider_id) if provider_id.present?
     host_instance ||= find_host_by_bios_uuid(bios_uuid) if bios_uuid.present?
@@ -330,7 +330,7 @@ class ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager < Man
     if node_info
       new_result.merge!(
         :identity_machine           => node_info.machineID,
-        :identity_system            => node_info.systemUUID,
+        :identity_system            => node_info.systemUUID&.gsub("\u0000", ""),
         :container_runtime_version  => node_info.containerRuntimeVersion,
         :kubernetes_proxy_version   => node_info.kubeProxyVersion,
         :kubernetes_kubelet_version => node_info.kubeletVersion
--- a/spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
@@ -703,30 +703,22 @@ describe ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager do
 
   describe "cross_link_node" do
     context "expected failures" do
-      before :each do
-        @node = OpenStruct.new(
-          :identity_system => "f0c1fe7e-9c09-11e5-bb22-28d2447dcefe",
-        )
-      end
-
-      after :each do
-        parser.send(:cross_link_node, @node)
-        expect(@node[:lives_on_id]).to eq(nil)
-        expect(@node[:lives_on_type]).to eq(nil)
-      end
+      let(:systemUUID) { "f0c1fe7e-9c09-11e5-bb22-28d2447dcefe" }
+      let(:node) { OpenStruct.new(:identity_system => systemUUID) }
 
       it "fails when provider type is wrong" do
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -293,7 +293,7 @@ def uuid_provider_types
   def cross_link_node(new_result)
     # Establish a relationship between this node and the vm it is on (if it is in the system)
     provider_id = new_result[:identity_infra]
-    bios_uuid   = new_result[:identity_system]&.gsub("\u0000", "")&.downcase
+    bios_uuid   = new_result[:identity_system]&.downcase
```
agrare, line 296: Simpler to do this substitution at the source :point_down: [Line 333](https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/436/files#diff-e9ac444f824d15fdf4a70d8a7ff9d178772401ba8c0916459f0c24b78b963298L333)

spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
```diff
@@ -703,30 +703,22 @@
 
   describe "cross_link_node" do
     context "expected failures" do
-      before :each do
-        @node = OpenStruct.new(
-          :identity_system => "f0c1fe7e-9c09-11e5-bb22-28d2447dcefe",
-        )
-      end
```
agrare, line 710: No reason for this to be a `before :each` can just be a `let()`

spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
```diff
@@ -703,30 +703,22 @@
...
-
-      after :each do
-        parser.send(:cross_link_node, @node)
-        expect(@node[:lives_on_id]).to eq(nil)
-        expect(@node[:lives_on_type]).to eq(nil)
-      end
```
agrare, line 716: Pull these out into each `it` so it doesn't look like the tests aren't doing anything

spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
```diff
@@ -703,30 +703,22 @@
...
-                                 :ext_management_system => @ems,
-                                 :name                  => "instance_id")
-      end
+                                :ext_management_system => @ems,
+                                :name                  => "instance_id")
+
+        parser.send(:cross_link_node, node)
 
-      it "fails when there is a NULL byte in the providerID" do
-        @node[:identity_system] = "UUID\u0000"
```
agrare, line 729: Now that we're stripping this out in the parse_node method a null byte won't even make it to `cross_link_node` so we don't need to check this anymore

spec/models/manageiq/providers/kubernetes/inventory/parser/container_manager_spec.rb
```diff
@@ -905,6 +897,43 @@
...
+            :resourceVersion   => '369104',
+            :creationTimestamp => '2015-12-06T11:10:21Z'
+          },
+          :spec     => {
+            :providerID => 'aws:///zone/aws-id'
+          },
+          :status   => {
+            :nodeInfo => {
+              :machineID  => 'id',
+              :systemUUID => "VENDOR,0123AB45C\u0000"
```
agrare, line 917: This is the format seen on a live power ocp cluster

agrare:
Should be backported together with https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/434

Fryguy:
Backported to `lasker` in commit ec92d5f4eb2379884b231efc776a7fd5364c201c.

```text
commit ec92d5f4eb2379884b231efc776a7fd5364c201c
Date:   Fri May 21 12:41:45 2021 -0400

    Merge pull request #436 from agrare/strip_null_bytes_before_saving
    
    Strip any null bytes from systemUUID in node hash
    
    (cherry picked from commit 9b5b107dfcf702901683f02ac19263e2730d217a)
```

PR #437, opened 2021-06-24
Restart the refresh worker when credentials change

agrare:
If the credentials or hostname change we should restart the refresh worker to break the blocking open connections that use the old credentials/endpoint

- [x] Depends https://github.com/ManageIQ/manageiq/pull/21296

Fixes https://github.com/ManageIQ/manageiq/issues/21253


Fryguy:
Partial backport to `morphy` via https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/459

PR #440, opened 2021-07-20
Allow duplicate hostnames with different ports

agrare:
Override the base `hostname_uniqueness_valid?` method for container managers to allow for multiple EMSs with the same hostname but different ports.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -98,6 +98,21 @@ class ManageIQ::Providers::Kubernetes::ContainerManager < ManageIQ::Providers::C
     ems_type
   end
 
+  def hostname_uniqueness_valid?
+    return unless hostname_required?
+    return unless hostname.present? # Presence is checked elsewhere
+    # check uniqueness per provider type
+
+    existing_providers = self.class.all - [self]
+    existing_endpoints = existing_providers.map do |ems|
+      next if ems.hostname.nil?
+
+      "#{ems.hostname.downcase}:#{ems.port}"
+    end
+
+    errors.add(:hostname, N_("has to be unique per provider type")) if existing_endpoints.include?("#{hostname.downcase}:#{port}")
+  end
+
   def self.params_for_create
     {
       :fields => [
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -98,6 +98,21 @@ def self.label_mapping_prefix
...
+  def hostname_uniqueness_valid?
+    return unless hostname_required?
+    return unless hostname.present? # Presence is checked elsewhere
+    # check uniqueness per provider type
+
+    existing_providers = self.class.all - [self]
+    existing_endpoints = existing_providers.map do |ems|
+      next if ems.hostname.nil?
+
+      "#{ems.hostname.downcase}:#{ems.port}"
```
agrare, line 110: If this were core I'd want to handle port being `nil` here but our k8s provider has always taken a port (6443 for k8s or 8443 for ocp) so pretty safe that we have a port here.

agrare:
As previously mentioned I think we could drop this uniqueness from core but that is more involved to ensure the entire codebase doesn't depend on unique hostnames.  Enabling this for just the container providers is lower risk

app/models/manageiq/providers/kubernetes/container_manager.rb
Fryguy: I merged, but what about amazon EKS? Does that also take a port?

agrare: @Fryguy yes a port is required for EKS https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/container_manager.rb#L186-L195

Sorry should have called that out in my comment on k8s and OCP, will update

Fryguy:
Backported to `lasker` in commit 86bca2fa15dd98eec959f113286bdc5358584ffb.

```text
commit 86bca2fa15dd98eec959f113286bdc5358584ffb
Date:   Mon Jul 26 11:19:05 2021 -0400

    Merge pull request #440 from agrare/allow_multiple_providers_same_hostname_different_port
    
    Allow duplicate hostnames with different ports
    
    (cherry picked from commit 73331d9a5328920b9fb265df2872e0b65b9ba078)
```

PR #444, opened 2021-08-25
Fix watch_thread noop argument

agrare:
Broken by: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/443


Fryguy:
is there a way to add a spec?

PR #445, opened 2021-08-25
Support oracle cloud node cross-linking

agrare:
OracleCloud nodes don't have a schema in their provider_id, this was causing a nil NoMethodError when trying to parse the scheme://instance_id

### DIFF
--- a/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
@@ -262,8 +262,14 @@ class ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager < Man
 
   def find_host_by_provider_id(provider_id)
     scheme, instance_uri = provider_id.split("://", 2)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -262,8 +262,14 @@ def map_labels(model_name, labels)
 
   def find_host_by_provider_id(provider_id)
     scheme, instance_uri = provider_id.split("://", 2)
-    prov, name_field = scheme_to_provider_mapping[scheme]
-    instance_id = instance_uri.split('/').last
+    if instance_uri.present?
+      prov, name_field = scheme_to_provider_mapping[scheme]
+      instance_id = instance_uri.split('/').last
+    elsif provider_id.start_with?("ocid1.instance")
```
chessbyte, line 268: Shouldn't this elsif bit be implemented in the Oracle Provider repo?  Maybe this method can be split in 2 and subclassed on the Oracle side.

PR #448, opened 2021-09-01
[WIP] Allow subclasses to override parts of connect_options

agrare:
Make it simpler for provider subclasses to override only parts of the kubernetes connect_options.

Cloud-hosted k8s clusters like OKE and IBM Cloud Kubernetes only need to set the bearer token


agrare:
Figured a better way of doing this, closing

PR #453, opened 2021-11-10
Delete hawkular metrics support

agrare:
Openshift 3.11 is the last version to support hawkular metrics, also supports prometheus metrics, and is going EOL in June 2022


manageiq-providers-kubernetes.gemspec
```diff
@@ -18,7 +18,6 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "hawkular-client",                 "~> 5.0"
```
Fryguy, line 21: 🎉

agrare: Yeah, very happy about this particular line being deleted :laughing:

Fryguy: For some reason dropping a dependency is my favorite part of any deletion; if the PR is designed to specifically delete a dependency, I'm like WAY happier about it.

cc @jrafanie 😉

agrare: Yup hawkular-client was preventing us from updating something a while ago (rest-client I think?) and took forever to get the maintainers to cut a release for us.

jrafanie: 🎉 One less dependency!!!  🥳

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
```diff
@@ -155,134 +74,5 @@
       @node.hardware.memory_mb = nil
       expect { @node.perf_collect_metrics('interval_name') }.to raise_error(described_class::TargetValidationError)
     end
-
-    # TODO: include also sort_and_normalize in the tests
-    METRICS_EXERCISES = [
```
agrare, line 160: NOTE we already have VCR specs for prometheus that cover these methods

Fryguy:
Do we need a data migration to drop any authentication records?

agrare:
@Fryguy yes I think we do

PR #474, opened 2022-10-27
Add allow_targeted_refresh to prevent fulls

agrare:
The EmsRefreshMixin checks for allow_targeted_refresh before queueing a target and it isn't supported it will queue a full.

This will cause the streaming refresher to shutdown all watch threads and do a full refresh again.

Ref: https://github.com/ManageIQ/manageiq-providers-vmware/pull/833


agrare:
cc @jrafanie

jrafanie (approved):
LGTM

PR #478, opened 2022-11-30
Improve the verification of default credentials

agrare:
Currently to verify the default credentials we open a kubeclient object and do discovery which is sufficient to check that the authentication is valid but doesn't check that we have access to anything.

This can lead to a scenario where the creds verify successfully but refresh/events/metrics collection immediately fails

https://github.com/ManageIQ/manageiq-providers-openshift/issues/235

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager.rb
@@ -704,7 +704,7 @@ Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
 
     case endpoint_name
     when 'default'
-      !!raw_connect(hostname, port, options)
+      verify_default_credentials(hostname, port, options)
     when 'kubevirt'
       verify_kubevirt_credentials(hostname, port, options)
     when 'hawkular'
@@ -796,6 +796,11 @@ Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
     Prometheus::ApiClient.client(:url => uri, :credentials => credentials, :options => prometheus_options)
   end
 
+  def self.verify_default_credentials(hostname, port, options)
+    kube = raw_connect(hostname, port, options)
+    kube && kube.api_valid?
+  end
+
   def self.verify_prometheus_credentials(hostname, port, options)
     !!prometheus_connect(hostname, port, options)&.query(:query => "ALL")&.kind_of?(Hash)
   end
@@ -849,6 +854,10 @@ Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
     end
 
     super(params, endpoints, authentications)
+  end
+
+  def verify_default_credentials(options)
+    with_provider_connection(options, &:api_valid?)
   end
 
   def verify_prometheus_credentials
@@ -942,7 +951,7 @@ Expecting to find com.redhat.rhsa-RHEL7.ds.xml.bz2 file there.'),
     when "kubevirt"
       verify_kubevirt_credentials
     else
-      with_provider_connection(options, &:api_valid?)
+      verify_default_credentials(options)
     end
   rescue SocketError,
          Errno::ECONNREFUSED,


app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -851,6 +856,10 @@ def edit_with_params(params, endpoints, authentications)
     super(params, endpoints, authentications)
   end
 
+  def verify_default_credentials(options)
+    with_provider_connection(options, &:api_valid?)
```
Fryguy, line 860: Should this also be checking more than just api_valid?

agrare: Yes haven't gotten to this one yet

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -851,6 +856,13 @@ def edit_with_params(params, endpoints, authentications)
     super(params, endpoints, authentications)
   end
 
+  def verify_default_credentials(options)
+    options[:service] ||= "kubernetes"
+    with_provider_connection(options) do |kube|
+      kube.api_valid? && !kube.get_namespaces(:limit => 1).nil?
```
Fryguy, line 862: Curiously why this doesn't quite match the class method.  Would it be possible to have this method call that method and DRY it up?

agrare: Ha was already working on that, PTAL

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -942,14 +962,14 @@ def verify_credentials(auth_type = nil, options = {})
...
-      with_provider_connection(options, &:api_valid?)
+      verify_default_credentials(options)
     end
   rescue SocketError,
          Errno::ECONNREFUSED,
          RestClient::ResourceNotFound,
          RestClient::InternalServerError => err
     raise MiqException::MiqUnreachableError, err.message, err.backtrace
-  rescue RestClient::Unauthorized   => err
+  rescue RestClient::Unauthorized, Kubeclient::HttpError => err
```
agrare, line 972: Kubeclient raises a `Kubeclient::HttpError` exception for this 403 case

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -934,23 +960,19 @@ def supported_auth_attributes
...
+        verify_default_credentials(options)
+      end
     end
-  rescue SocketError,
-         Errno::ECONNREFUSED,
-         RestClient::ResourceNotFound,
-         RestClient::InternalServerError => err
-    raise MiqException::MiqUnreachableError, err.message, err.backtrace
-  rescue RestClient::Unauthorized   => err
-    raise MiqException::MiqInvalidCredentialsError, err.message, err.backtrace
```
agrare, line 953: This set of rescues weren't present in the class-level validation so I moved this to a `connection_rescue_block` similar to how other providers handle this.

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -796,6 +809,14 @@ def self.prometheus_connect(hostname, port, options)
     Prometheus::ApiClient.client(:url => uri, :credentials => credentials, :options => prometheus_options)
   end
 
+  def self.verify_k8s_connectiton(kube)
```
Fryguy, line 812:
```suggestion
  def self.verify_k8s_connection(kube)
```

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -702,22 +702,35 @@ def self.verify_credentials(args)
...
 
+  def self.connection_rescue_block
+    yield
+  rescue SocketError,
+         Errno::ECONNREFUSED,
+         RestClient::ResourceNotFound,
+         RestClient::InternalServerError => err
+    raise MiqException::MiqUnreachableError, err.message, err.backtrace
+  rescue RestClient::Unauthorized, Kubeclient::HttpError => err
+    raise MiqException::MiqInvalidCredentialsError, err.message, err.backtrace
```
Fryguy, line 731: The err.backtrace shouldn't be necessary here, because the `cause` will internally be set anyway and stacktraces will automatically include it:

```
[1] pry(main)> begin
[1] pry(main)*   1 / 0
[1] pry(main)* rescue => err
[1] pry(main)*   raise StandardError, "oops"
[1] pry(main)* end
Traceback (most recent call last):
	 1: from (pry):2:in `__pry__'
(pry):2:in `/': divided by 0 (ZeroDivisionError)
	 1: from (pry):1:in `__pry__'
(pry):4:in `rescue in __pry__': oops (StandardError)
```

Fryguy: Well, to be clear, stacktraces printed by Ruby will automatically walk it - not sure if we manually walk it anywhere and those _might_ be affected.

agrare: It was like this before, I can remove if you want

agrare: Removed, but not sure why it was added in the first place and we do this in a number of other places so not sure if it is needed

agrare:
With this change if a Refresh Worker starts up with invalid credentials it immediately shuts down and marks the credentials invalid and does not start back up:
```
[----] I, [2022-12-01T13:59:42.230754 #15613:a1e0]  INFO -- evm: MIQ(ManageIQ::Providers::Openshift::ContainerManager#with_provider_connection) Connecting through ManageIQ::Providers::Openshift::ContainerManager: [api.openshift.local]
[----] W, [2022-12-01T13:59:42.528234 #15613:a1e0]  WARN -- evm: MIQ(ManageIQ::Providers::Openshift::ContainerManager#authentication_check_no_validation) type: ["bearer"] for [15] [api.openshift.local] Validation failed: invalid, namespaces is forbidden: User "system:serviceaccount:cp4waiops:ibm-infra-management-application-orchestrator" cannot list resource "namespaces" in API group "" at the cluster scope
[----] W, [2022-12-01T13:59:42.528467 #15613:a1e0]  WARN -- evm: MIQ(AuthToken#validation_failed) [ExtManagementSystem] [15], previously valid on: 2022-12-01 18:58:39 UTC, previous status: [Valid]
[----] I, [2022-12-01T13:59:43.272810 #15613:a1e0]  INFO -- evm: MIQ(MiqQueue.put) Message id: [26383], Zone: [default], Role: [], Server: [], MiqTask id: [], Handler id: [], Ident: [generic], Target id: [], Instance id: [], Task id: [], Command: [MiqEvent.raise_evm_event], Timeout: [600], Priority: [100], State: [ready], Deliver On: [], Data: [], Args: [["ManageIQ::Providers::Openshift::ContainerManager", 15], "ems_auth_invalid", {}]
[----] E, [2022-12-01T13:59:43.273658 #15613:a1e0] ERROR -- evm: MIQ(ManageIQ::Providers::Openshift::ContainerManager::RefreshWorker::Runner) ID [68] PID [15613] GUID [7322b929-059e-4a4d-b43c-b6a033d4a6fc] Error in before_exit: undefined method `emstype' for nil:NilClass
[----] I, [2022-12-01T13:59:43.323629 #15613:a1e0]  INFO -- evm: MIQ(ManageIQ::Providers::Openshift::ContainerManager::RefreshWorker#log_status) [Refresh Worker for Provider: api.openshift.local] Worker ID [68], PID [15613], GUID [7322b929-059e-4a4d-b43c-b6a033d4a6fc], Last Heartbeat [2022-12-01 18:59:33 UTC], Process Info: Memory Usage [303603712], Memory Size [743505920], Proportional Set Size: [282480000], Unique Set Size: [281532000], Memory % [0.04], CPU Time [476.0], CPU % [0.0], Priority [20]
[----] E, [2022-12-01T13:59:43.323892 #15613:a1e0] ERROR -- evm: MIQ(ManageIQ::Providers::Openshift::ContainerManager::RefreshWorker::Runner) ID [68] PID [15613] GUID [7322b929-059e-4a4d-b43c-b6a033d4a6fc] EMS id [15] failed authentication check. Worker exiting.
[----] I, [2022-12-01T13:59:43.326084 #15613:a1e0]  INFO -- evm: Deleting worker record for ManageIQ::Providers::Openshift::ContainerManager::RefreshWorker, id 68
```

app/models/manageiq/providers/kubernetes/container_manager.rb
Fryguy: Could be the version of Ruby - there was a time when cause _wasn't_ set (I think it was introduced in Ruby 2.1)

Fryguy:
Backported to `oparin` in commit ef6f0dee2ffa957840bca24b6f809bdc09fd56ce.

```text
commit ef6f0dee2ffa957840bca24b6f809bdc09fd56ce
Date:   Thu Dec 1 21:10:10 2022 -0500

    Merge pull request #478 from agrare/improve_verify_credentials
    
    Improve the verification of default credentials
    
    (cherry picked from commit c3767b6569f1c4cd02c43425ddac8c3da4412928)
```

PR #484, opened 2023-04-28
Fix Node network metrics collection on s390x

agrare:
Container Nodes on the s390x architecture have interfaces called `enc.*` which was causing no net_usage_rate_average values to be returned.

NOTE originally I reported that `br0` worked, and while it does x86 nodes also have a `br0` and we specifically request the underlying interfaces rather than the bridge.  For a fix for backport I wanted to maintain this behavior while investigating if there is a more reliable way of getting the metrics we're looking for.


Fryguy (approved):

Fryguy:
Backported to `petrosian` in commit 79ddf3a05b52041d864e561dc4bd781c3210254e.

```text
commit 79ddf3a05b52041d864e561dc4bd781c3210254e
Date:   Fri Apr 28 10:53:48 2023 -0400

    Merge pull request #484 from agrare/fix_node_network_metrics_collection_s390x
    
    Fix Node network metrics collection on s390x
    
    (cherry picked from commit 838aa96be5e4efac47fdf05dee276c3875e3d012)
```

Issue #485, opened 2023-04-28
Don't ignore all metrics if a single counter type is not returned


agrare:
There are occasions where due to a prometheus query error no metrics values will be returned for one or more counters.

We currently will reject every datapoint that does not have all metric counters values present which can end up returning 0 values if there is a counter not being returned for every timestamp.

[[ref]](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb#L34-L39)
```
    def ts_values
      # Filtering out entries that are not containing all the metrics.
      # This generally happens because metrics are collected at slightly
      # different times and could produce entries that are incomplete.
      @ts_values.select { |_, v| @metrics.all? { |k| v.key?(k) } }
    end
```

Fryguy:
Closed via #486

PR #486, opened 2023-04-28
Don't exclude all metrics if one counter is missing

agrare:
Only filter out timestamps which have partial metric values for counters which exist in other timestamps.  If no timestamp has a counter then we should consider it requested but not returned and continue on.

https://github.com/ManageIQ/manageiq-providers-kubernetes/issues/485


Fryguy:
Backported to `petrosian` in commit 200c159fc42be2559580f914485e468a14042a18.

```text
commit 200c159fc42be2559580f914485e468a14042a18
Date:   Mon May 1 17:25:33 2023 -0400

    Merge pull request #486 from agrare/dont_return_no_metrics_for_a_missing_counter
    
    Don't exclude all metrics if one counter is missing
    
    (cherry picked from commit 278692a957da50d4fd79dd967b62dceaf43988c1)
```

PR #489, opened 2023-05-19
Restart collector threads on 410 Gone

agrare:
Instead of trying to restart a watch when a 410 Gone error is received just let the thread die and be restarted.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
@@ -4,18 +4,18 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThr
   HTTP_UNAUTHORIZED = 401
   HTTP_GONE         = 410
 
-  def self.start!(ems, queue, entity_type, resource_version)
-    new(ems.connect_options, ems.class, queue, entity_type, resource_version).tap(&:start!)
+  def self.start!(ems, queue, entity_type, resource_versions)
+    new(ems.connect_options, ems.class, queue, entity_type, resource_versions).tap(&:start!)
   end
 
-  def initialize(connect_options, ems_klass, queue, entity_type, resource_version)
-    @connect_options = connect_options
-    @ems_klass       = ems_klass
-    @entity_type     = entity_type
-    @finish          = Concurrent::AtomicBoolean.new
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
```diff
@@ -61,14 +61,14 @@ def collector_thread
...
-            _log.warn("Received an error watching #{entity_type}: [#{code} #{reason}], [#{message}]")
-            break
+            _log.error("Received an error watching #{entity_type}: [#{code} #{reason}], [#{message}]")
+            raise
           end
 
-          current_resource_version = notice.object&.metadata&.resourceVersion
-          self.resource_version    = current_resource_version if current_resource_version.present?
+          current_resource_version       = notice.object&.metadata&.resourceVersion
+          resource_versions[entity_type] = current_resource_version if current_resource_version.present?
```
Fryguy, line 71: Since this is being accessed cross-thread, do we need to make sure we use a thread safe hash?  I understand that each watch thread is technically a separate key, but I'm not sure if concurrent updates even on different keys could cause problems.

spec/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread_spec.rb
```diff
@@ -1,12 +1,13 @@
-describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThread do
+  describe ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThread do
```
Fryguy, line 1: Typo?

app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
agrare: on CRuby `Hash` is thread safe (Concurrent::Hash == Hash) we could use `Concurrent::Hash` if we  want to support other rubies

Fryguy: Nah, this is fine.

agrare:
With just this in place we're still losing around `256KiB` per thread crash.  Previously we would spike memory usage growing to GiBs.  Still investigating the other leaks.

PR #490, opened 2023-05-19
Ensure WatchStream is always finished

agrare:
Don't restart a watch without finishing the previous one


app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
```diff
@@ -83,6 +83,8 @@ def collector_thread
 
         retry_connection = false
         retry
+      ensure
+        watch.finish rescue nil
```
Fryguy, line 87: Nice - this also handles the `retry` case.

Fryguy:
Skipping backport to `petrosian`, because it is already in the branch.

PR #493, opened 2023-06-12
Check metrics authentication validity before perf_capture_queue

agrare:
Check that the EMS's metrics authentication is valid before queueing perf_capture calls, potentially flooding the queue with calls that are going to fail later.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -81,11 +81,15 @@ module ManageIQ::Providers
       raise TargetValidationWarning, "metrics authentication isn't valid for #{target_name}" unless metrics_connection_valid?(ems)
     end
 
-    def capture_context(_ems, target, start_time, end_time)
+    def build_capture_context!(ems, target, start_time, end_time)
+      verify_metrics_connection!(ems)
       # make start_time align to minutes
       start_time = start_time.beginning_of_minute
 
-      prometheus_capture_context(target, start_time, end_time)
+      context = prometheus_capture_context(target, start_time, end_time)
+      raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if context.nil?
+
+      context
     end
 
     def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
@@ -97,11 +101,7 @@ module ManageIQ::Providers
                 "[#{start_time}] [#{end_time}]")
 
       begin
-        verify_metrics_connection!(ems)
-
-        context = capture_context(ems, target, start_time, end_time)
-
-        raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if context.nil?
+        context = build_capture_context!(ems, target, start_time, end_time)
       rescue TargetValidationError, TargetValidationWarning => e
         _log.send(e.log_severity, "[#{target_name}] #{e.message}")
         ems.try(:update,
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -73,8 +73,6 @@ module ManageIQ::Providers
     end
 
     def verify_metrics_connection!(ems)
-      t = target || ems
-      target_name = "#{t.class.name.demodulize}(#{t.id})"
       raise TargetValidationError, "no provider for #{target_name}" if ems.nil?
 
       raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if metrics_connection(ems).nil?
@@ -96,9 +94,7 @@ module ManageIQ::Providers
       start_time ||= 15.minutes.ago.beginning_of_minute.utc
       ems = target.ext_management_system
 
-      target_name = "#{target.class.name.demodulize}(#{target.id})"
-      _log.info("Collecting metrics for #{target_name} [#{interval_name}] " \
-                "[#{start_time}] [#{end_time}]")
+      _log.info("Collecting metrics for #{target_name} [#{interval_name}] [#{start_time}] [#{end_time}]")
 
       begin
         context = build_capture_context!(ems, target, start_time, end_time)
@@ -129,6 +125,15 @@ module ManageIQ::Providers
 
       [{target.ems_ref => VIM_STYLE_COUNTERS},
        {target.ems_ref => context.ts_values}]
+    end
+
+    private
+
+    def target_name
+      @target_name ||= begin
+        t = target || ems
+        target_name = "#{t.class.name.demodulize}(#{t.id})"
+      end
     end
   end
 end
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -49,12 +49,36 @@ module ManageIQ::Providers
       }
     }
 
+    def capture_ems_targets(_options = {})
+      begin
+        verify_metrics_connection!(ems)
+      rescue TargetValidationError, TargetValidationWarning => e
+        _log.send(e.log_severity, e.message)
+        return []
+      end
+
+      super
+    end
+
     def prometheus_capture_context(target, start_time, end_time)
       PrometheusCaptureContext.new(target, start_time, end_time, INTERVAL)
     end
 
     def metrics_connection(ems)
       ems.connection_configurations.prometheus
+    end
+
+    def metrics_connection_valid?(ems)
+      metrics_connection(ems)&.authentication&.status == "Valid"
+    end
+
+    def verify_metrics_connection!(ems)
+      t = target || ems
+      target_name = "#{t.class.name.demodulize}(#{t.id})"
+      raise TargetValidationError, "no provider for #{target_name}" if ems.nil?
+
+      raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if metrics_connection(ems).nil?
+      raise TargetValidationWarning, "metrics authentication isn't valid for #{target_name}" unless metrics_connection_valid?(ems)
     end
 
     def capture_context(_ems, target, start_time, end_time)
@@ -73,11 +97,7 @@ module ManageIQ::Providers
                 "[#{start_time}] [#{end_time}]")
 
       begin
-        raise TargetValidationError, "no provider for #{target_name}" if ems.nil?
-
-        connection = metrics_connection(ems)
-        raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if connection.nil?
-        raise TargetValidationWarning, "metrics authentication isn't valid for #{target_name}" unless connection.authentication&.status == "Valid"
+        verify_metrics_connection!(ems)
 
         context = capture_context(ems, target, start_time, end_time)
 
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -49,6 +49,12 @@ def log_severity
       }
     }
 
+    def capture_ems_targets(_options = {})
```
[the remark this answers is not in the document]
agrare: I like that idea, I'm trying to work out a way that works well with the other checks that we have in `def perf_collect_metrics` so we aren't duplicating the work

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -57,6 +68,19 @@ def metrics_connection(ems)
       ems.connection_configurations.prometheus
     end
 
+    def metrics_connection_valid?(ems)
+      metrics_connection(ems)&.authentication&.status == "Valid"
+    end
+
+    def verify_metrics_connection!(ems)
+      t = target || ems
```
agrare, line 76: @kbrock do we have a general way of handling this?  When called from `perf_capture_timer` as `ems.perf_capture_object.perf_capture_all_queue` `target` is `nil` and only `ems` is set

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
```diff
@@ -59,6 +59,33 @@
     end
   end
 
+  context "#perf_capture_all_queue" do
+    it "returns the objects" do
+      expect(@ems_kubernetes.perf_capture_object.perf_capture_all_queue).to include("Container" => [@container], "ContainerNode" => [@node])
```
agrare, line 64: TODO why only `@container` and `@node` and not `@group`

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
```diff
@@ -59,6 +59,33 @@
...
+    end
+
+    context "with a missing metrics endpoint" do
+      before do
+        @ems_kubernetes.endpoints.find_by(:role => "prometheus").destroy
```
agrare, line 69: NOTE I don't love how this spec sets up these ivars since it makes it difficult to override in sub-contexts.  I worked around it but I think I should refactor to use `let`

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
```diff
@@ -59,6 +59,33 @@
...
+
+      it "returns no objects" do
+        expect(@ems_kubernetes.perf_capture_object.perf_capture_all_queue).to be_empty
+      end
+    end
+
+    context "with invalid authentication on the metrics endpoint" do
+      before do
+        @ems_kubernetes.authentications.find_by(:authtype => "prometheus").update!(:status => "Error")
+        @ems_kubernetes.reload
```
agrare, line 80: NOTE a good example of why the `before {}` setting up ivars isn't great

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
kbrock: I really don't like how `@target` is set behind the scenes.
perf_capture_gap, perf_capture_all_queue, perf_capture_realtime_queue ( /via ui)

if you are running capture for a whole ems, `target` is null.
If you are running capture for a single object, `target` is single object.
If you are running capture for a batch/vmware, `target`/`targets` is an array

I think I'd just use `ems`

kbrock: Does it make sense to ensure that `prometheus` exists in here? (aka `capture_context`)

That way verify does all the raises here.

agrare: > I think I'd just use ems

Well when we use this below it is on a single object so this has to work with `target` or `ems`

agrare: > Does it make sense to ensure that prometheus exists in here? (aka capture_context)

Yeah I was thinking that but we use the return of `capture_context` and returning something from a `verify_*` method seems weird.

Also building a prometheus_capture_context object seems like overkill when it isn't used in the `perf_capture_timer` case here

agrare: I probably could call this from `capture_context` and raise if that is nil in that method

agrare: Okay I pushed a second commit which moves some things around, LMK if you like it better or worse and it is easy to revert out

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -73,15 +101,7 @@ def perf_collect_metrics(interval_name, start_time = nil, end_time = nil)
...
-        raise TargetValidationError, "no provider for #{target_name}" if ems.nil?
-
-        connection = metrics_connection(ems)
-        raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if connection.nil?
-        raise TargetValidationWarning, "metrics authentication isn't valid for #{target_name}" unless connection.authentication&.status == "Valid"
-
-        context = capture_context(ems, target, start_time, end_time)
-
-        raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if context.nil?
+        context = build_capture_context!(ems, target, start_time, end_time)
```
kbrock, line 104: yea. this is nice

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
kbrock: You know, it doesn't look like it needs to know about target at all beside building a couple of printf strings. Looks good for now.

TANGENT:

Yea, I guess the code is currently coded to reference a single target.

I have on my short list to convert this over to `targets`.
For some reason, I can't find the initialize method on `PrometheusCaptureContext`.
But from what I can tell, that code assumes a single target when building the labels.

Not sure how hard it would be to handle a list of targets over there.

kbrock (approved):
LGTM

agrare:
@kbrock can you re-review?  I refactored the specs to use `let()` and pulled out `target_name` into a method to keep the logic in one place.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -57,31 +68,36 @@ def metrics_connection(ems)
...
-    def capture_context(_ems, target, start_time, end_time)
+    def metrics_connection_valid?(ems)
+      metrics_connection(ems)&.authentication&.status == "Valid"
+    end
+
+    def verify_metrics_connection!(ems)
+      raise TargetValidationError, "no provider for #{target_name}" if ems.nil?
+
+      raise TargetValidationWarning, "no metrics endpoint found for #{target_name}" if metrics_connection(ems).nil?
+      raise TargetValidationWarning, "metrics authentication isn't valid for #{target_name}" unless metrics_connection_valid?(ems)
```
jrafanie, line 79: Do we need to check if the ems is paused?

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -49,6 +49,17 @@ def log_severity
       }
     }
 
+    def capture_ems_targets(_options = {})
+      begin
+        verify_metrics_connection!(ems)
+      rescue TargetValidationError, TargetValidationWarning => e
+        _log.send(e.log_severity, e.message)
+        return []
+      end
```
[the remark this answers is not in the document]
jrafanie: This is a "future" PR question... I'm fine with fixing kubernetes in this way now.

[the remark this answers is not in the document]
agrare: It is a little tricky because not all providers have different endpoints and/or authentications so what you need to check varies.

What I've considered doing is consolidating this in the `supports :metrics {}` block in ext_management_system so that the logic could be limited to the provider but that is going to be more of a cross-provider change.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
agrare: I wonder if we should do that up in core perf_capture_timer since that is going to be the same for all providers, good call though.  I'm a little wary of that just because it feels like every single queue method would have to implement this check but I don't know that there's a more central place to handle it.

agrare: https://github.com/ManageIQ/manageiq/pull/22564

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
```diff
@@ -1,78 +1,75 @@
...
-      :ems_ref               => 'target'
-    )
+  let(:ems)               { FactoryBot.create(:ems_kubernetes_with_zone, :with_metrics_endpoint) }
+  let(:container_project) { FactoryBot.create(:container_project,          :ext_management_system => ems) }
+  let!(:group)            { FactoryBot.create(:kubernetes_container_group, :ext_management_system => ems, :container_node => node) }
+  let!(:container)        { FactoryBot.create(:kubernetes_container,       :ext_management_system => ems, :container_group => group, :container_project => container_project) }
+  let(:node) do
+    FactoryBot.create(:kubernetes_node, :name => 'node', :ext_management_system => ems, :ems_ref => 'target').tap do |node|
+      node.computer_system.hardware = FactoryBot.create(:hardware, :cpu_total_cores => 2, :memory_mb => 2_048)
+    end
```
kbrock, line 9: lets keep. But in the future if we merge https://github.com/ManageIQ/manageiq/pull/22565 this could be:

```suggestion
    FactoryBot.create(:kubernetes_node, :name => 'node', :ext_management_system => ems, :ems_ref => 'target', :hardware => FactoryBot.create(:hardware, :cpu_total_cores => 2, :memory_mb => 2_048))
```

agrare: :+1: yeah I wanted to keep this dependent-less for simpler backport but that's a much better way to do it

kbrock (approved):
these updates look good.

spec/models/manageiq/providers/kubernetes/container_manager/metrics_capture_spec.rb
kbrock: agreed. keep. reduce dependencies

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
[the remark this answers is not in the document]
kbrock: updating `supports` to handle all the actual logic totally got me in a few PRs. they are still outstanding :(  ✨

Fryguy:
Backported to `petrosian` in commit 968fcea19c17ceab63e9272a3df91a85461d58aa.

```text
commit 968fcea19c17ceab63e9272a3df91a85461d58aa
Author: Keenan Brock <keenan@thebrocks.net>
Date:   Tue Jun 13 18:21:24 2023 -0400

    Merge pull request #493 from agrare/check_metrics_connection_validity_before_queueing_perf_captures
    
    Check metrics authentication validity before perf_capture_queue
    
    (cherry picked from commit ba04739fc4ab3eee4625630ec4622dd71d576d67)
```

PR #494, opened 2023-06-13
Improve performance capture ems targets

agrare:
For Kubernetes type providers if the ems has metrics credentials then all instances of container_nodes, container_groups, etc... supports capture.
    
We don't have to check `.supporting(:capture)` if we are in a kubernetes provider, but this isn't generally true so only implement this behavior in k8s.

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/493

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
@@ -57,7 +57,12 @@ module ManageIQ::Providers
         return []
       end
 
-      super
+      MiqPreloader.preload([ems], :container_images => :tags, :container_nodes => :tags, :container_groups => [:tags, :containers => :tags])
+
+      with_archived(ems.all_container_nodes) +
+        with_archived(ems.all_container_groups) +
+        with_archived(ems.all_containers) +
+        with_archived(ems.container_images)
     end
 
     def prometheus_capture_context(target, start_time, end_time)


app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -49,6 +49,22 @@ def log_severity
...
+        _log.send(e.log_severity, e.message)
+        return []
+      end
+
+      MiqPreloader.preload([ems], :container_images => :tags, :container_nodes => :tags, :container_groups => [:tags, :containers => :tags])
+
+      with_archived(ems.all_container_nodes) +
+        with_archived(ems.all_container_groups) +
+        with_archived(ems.all_containers) +
+        with_archived(ems.container_images)
```
agrare, line 65: NOTE there are no container_images that support capture, but they are rolled up.
Do we need to return them from capture_ems_targets for rollups to work?  `ems.container_images.supporting(:capture)` will return `[]` always.

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
```diff
@@ -57,7 +57,12 @@ def capture_ems_targets(_options = {})
         return []
       end
 
-      super
+      MiqPreloader.preload([ems], :container_images => :tags, :container_nodes => :tags, :container_groups => [:tags, :containers => :tags])
```
kbrock, line 60: `:tags` are included so we can ask each of the objects if they have capture enabled. (answered through tags)

If we are punting on that behavior, I think we can drop `=> :tags`

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
kbrock: `capture_ems_target` are only called for capturing.

the images are captured and then that kicks off perf_capture_rollup which then rollsup images into containers via `ContainerImage::PERF_ROLLUP_CHILDREN`

So I don't see how we can roll those up without a container image going through metrics collection.

the `VimPerformanceState` does collect the images association, but I think that is unrelated

kbrock:
@agrare what do you need from me before I merge it?
I think this looks good to go (but am curious about your/my question - not enough to hold this up though)

app/models/manageiq/providers/kubernetes/container_manager/metrics_capture.rb
kbrock: @agrare ok. This is a simple answer: If no data will come back, then there is no reason to query them.

Was that the only thing that was holding this up?

kbrock:
Can we just add these changes with a TODO in there to double check that we still need the container_images?

Seem to remember that this bought us some performance and while we could get more if we dropped container images, seems getting a little bit would be best?

PR #496, opened 2023-06-23
Fix started_at/finished_at always being nil

agrare:
The `state_info` had symbol keys like `:startedAt` and `:finishedAt` and we were looking for string keys like `"startedAt"` and `"finishedAt"` causing these values to always be `nil` even if data existed in the API response.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
+++ b/app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
@@ -851,13 +851,16 @@ class ManageIQ::Providers::Kubernetes::Inventory::Parser::ContainerManager < Man
 
   def parse_container_state(state_hash)
     return {} if state_hash.to_h.empty?
-    res = {}
+
     # state_hash key is the state and value are attributes e.g 'running': {...}
-    (state, state_info), = state_hash.to_h.to_a
-    res[:state] = state
-    %w(reason started_at finished_at exit_code signal message).each do |attr|
-      res[attr.to_sym] = state_info[attr.camelize(:lower)]
+    (state, state_info), = state_hash.to_h.deep_symbolize_keys.to_a
+
+    res = {:state => state}
+
+    %i[reason reason started_at startedAt finished_at finishedAt exit_code exitCode signal signal message message].each_slice(2) do |attr, state_info_attr|
+      res[attr] = state_info[state_info_attr]
     end
+
     res
   end
 
--- a/spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
@@ -108,14 +108,10 @@ describe ManageIQ::Providers::Kubernetes::ContainerManager::Refresher do
           # :backing_ref => "docker://87cd51044d7175c246fa1fa7699253fc2aecb769021837a966fa71e9dcb54d71"
         )
 
-        [
-          @container.started_at,
-          @container.finished_at,
-          @container.last_started_at,
-          @container.last_finished_at,
-        ].each do |date_|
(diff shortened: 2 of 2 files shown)


kbrock (approved):

kbrock:
@Fryguy Is it possible to raise the priority of this bugfix? Or are we thinking the deleted_on logic is not critical?

spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
```diff
@@ -114,7 +114,7 @@ def assert_specific_container
           @container.last_started_at,
           @container.last_finished_at,
         ].each do |date_|
-          expect(date_.kind_of?(ActiveSupport::TimeWithZone) || date_.kind_of?(NilClass)).to be_truthy
+          expect(date_.kind_of?(ActiveSupport::TimeWithZone))
```
Fryguy, line 117: Was this change intentional?  This expectation is no longer expecting anything?

Fryguy, line 117: Was this change intentional?  This expectation is no longer expecting anything?

```suggestion
          expect(date_.kind_of?(ActiveSupport::TimeWithZone)).to be
```

or

```suggestion
          expect(date_.kind_of?(ActiveSupport::TimeWithZone)).to be_truthy
```

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
```diff
@@ -853,10 +853,10 @@ def parse_container_state(state_hash)
     return {} if state_hash.to_h.empty?
     res = {}
     # state_hash key is the state and value are attributes e.g 'running': {...}
-    (state, state_info), = state_hash.to_h.to_a
+    (state, state_info), = state_hash.to_h.deep_symbolize_keys.to_a
     res[:state] = state
     %w(reason started_at finished_at exit_code signal message).each do |attr|
```
Fryguy, line 858: Maybe not for this PR, but we can pre-symbolize everything and not need to the .to_sym calls - Something like:

```suggestion
    %i(reason reason started_at startedAt finished_at finishedAt exit_code exitCode signal signal message message).each_slice(2) do |attr, state_info_attr|
```

Fryguy:
@kbrock we'll backport to petrosian as part of the set, yes.  Even so, we'll need to deal with existing records anyway.

spec/models/manageiq/providers/kubernetes/container_manager/refresher_spec.rb
agrare: I'm actually just going to drop the whole loop since `expect time or nil` is what hid this bug in the first place

app/models/manageiq/providers/kubernetes/inventory/parser/container_manager.rb
kbrock: yea, instead of each_slice, just using a hash is easier for me to read. But both work

agrare: Done

kbrock:
Rebase please. I think kubernetes HEAD fixes this issue

Fryguy:
Backported to `petrosian` in commit 7793ce09194180ee171d172d73f9de22c95afdbb.

```text
commit 7793ce09194180ee171d172d73f9de22c95afdbb
Author: Keenan Brock <keenan@thebrocks.net>
Date:   Fri Jul 14 16:09:47 2023 -0400

    Merge pull request #496 from agrare/fix_started_at_finished_at_timestamps
    
    Fix started_at/finished_at always being nil
    
    (cherry picked from commit 1c753fe979d1ecec638b211d88f9ff7af8d8ed92)
```

PR #497, opened 2023-06-26
Convert the duration to a float

jrafanie:
Thread#join used to accept duration objects but not since ruby 3/rails 6.1

See also https://github.com/ManageIQ/manageiq-providers-vmware/pull/832

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
@@ -31,7 +31,7 @@ class ManageIQ::Providers::Kubernetes::ContainerManager::RefreshWorker::WatchThr
 
     finish.make_true
     watch&.finish rescue nil
-    thread&.join(join_limit)
+    thread&.join(join_limit.to_f)
   end
 
   protected


Fryguy (approved):

agrare (approved):

app/models/manageiq/providers/kubernetes/container_manager/refresh_worker/watch_thread.rb
```diff
@@ -31,7 +31,7 @@ def stop!(join_limit = 10.seconds)
 
     finish.make_true
     watch&.finish rescue nil
-    thread&.join(join_limit)
+    thread&.join(join_limit.to_f)
```
Fryguy, line 34: Can you also change the incoming default parameter to just `10.0`?  This way we're not needlessly creating a duration only to convert it.  We still need the .to_f, in case the incoming parameter is not a float.

```ruby
def stop!(join_limit = 10.0) # 10 seconds
```

Fryguy:
Backported to `petrosian` in commit 51d050413e54f22e4f9f292659b0b7ade211885a.

```text
commit 51d050413e54f22e4f9f292659b0b7ade211885a
Author: Adam Grare <adam@grare.com>
Date:   Mon Jun 26 11:38:58 2023 -0400

    Merge pull request #497 from jrafanie/fix_join_with_duration
    
    Convert the duration to a float
    
    (cherry picked from commit 1371d53a794a3c0aaea3a2614d37245a424d7e63)
```

PR #500, opened 2023-07-25
Fix No Method get_nodes for NilClass invalid token

agrare:
If a token is invalid when a refresh starts the connect method was eating the exception and returning `nil` rather than allowing the refresh to fail.


Fryguy:
Backported to `petrosian` in commit cf6a947769dc880deb119d9c7506810cb850f3ee.

```text
commit cf6a947769dc880deb119d9c7506810cb850f3ee
Date:   Tue Jul 25 16:00:54 2023 -0400

    Merge pull request #500 from agrare/fix_no_method_get_nodes_nil_class
    
    Fix No Method get_nodes for NilClass invalid token
    
    (cherry picked from commit 2bec00fe4ee67e4b81164b428616795ffd7469ee)
```

PR #501, opened 2023-08-11
Reject any events which do not have a timestamp

agrare:
The event timestamp is critical for uniquely identifying events and ensuring there are no duplicates.

We cannot properly handle events without a timestamp so we need to not queue these for the event handler.

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
@@ -48,7 +48,10 @@ module ManageIQ::Providers::Kubernetes::ContainerManager::EventCatcherMixin
 
   def queue_event(event)
     event_data = extract_event_data(event)
-    return if event_data.nil?
+    if !event_valid?(event_data)
+      _log.info "#{log_prefix} Skipping invalid event [#{event_data[:event_type]}]"
+      return
+    end
 
     _log.info "#{log_prefix} Queuing event [#{event_data}]"
     event_hash = ManageIQ::Providers::Kubernetes::ContainerManager::EventParser.event_to_hash(event_data, @cfg[:ems_id])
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
```diff
@@ -48,13 +48,16 @@ def monitor_events
 
   def queue_event(event)
     event_data = extract_event_data(event)
+    return if event_data.nil?
+
     _log.info "#{log_prefix} Queuing event [#{event_data}]"
```
jrafanie, line 53: isn't it Queueing 🤣 ... I guess both are ok... silly me for googling it https://www.qminder.com/blog/queue-management/queue-vs-que-queuing-language/#:~:text=The%20long%20and%20short%20of,form%20of%20the%20same%20word

agrare:
I'm going to rework this a bit since there won't be anything useful to log if we just return `nil`

app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
```diff
@@ -106,4 +110,11 @@ def extract_event_data(event)
 
     event_data
   end
+
+  def event_valid?(event_data)
+    # If there is no timestamp we cannot properly handle the event
+    return false if event_data[:timestamp].nil?
+
+    true
```
agrare, line 118: NOTE I know I could just do `event_data[:timestamp].present` but wanted to set this method up for multiple checks on the data not just one in the future

agrare:
@Fryguy ready for review

app/models/manageiq/providers/kubernetes/container_manager/event_catcher_mixin.rb
Fryguy: Not for this PR, but we might want the reason to come out of this method when it is invalid, then we could say something like "Skipping invalid event [foo] because it is missing a timestamp"

Fryguy:
Backported to `petrosian` in commit 272dade1b16cfb20c7c72bc5a7cfb43678fd5c6a.

```text
commit 272dade1b16cfb20c7c72bc5a7cfb43678fd5c6a
Date:   Fri Aug 11 13:48:42 2023 -0400

    Merge pull request #501 from agrare/reject_events_without_a_timestamp
    
    Reject any events which do not have a timestamp
    
    (cherry picked from commit c2df77bc3d66e712872a86e9ffac690342aca737)
```

PR #506, opened 2023-09-22
The prometheus_alerts endpoint is no longer supported

agrare:
OCPv3 had prometheus-alert-buffer
https://github.com/openshift/prometheus-alert-buffer which was used by
the prometheus_alerts endpoint and MonitoringManager.

This is not present on OCPv4 and it isn't possible to add a
prometheus_alerts endpoint.

Support for OCPv3 has been dropped so there is no supported version of
openshift where you are able to add an events endpoint.

PR #507, opened 2023-09-22
Delete Prometheus Alerts MonitoringManager

agrare:
This was only supported on OCPv3 and is not expected to be reimplemented in the future.

Based on: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/506

NOTE I split this out from https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/506 because we might want to backport that (it causes user confusion) where as this is strictly a backend cleanup that also will require some data migrations to delete any old records.


kbrock:
kicking

kbrock:
ok, looks like the test failure is trivial but legit. The list has changed:

```
       the missing elements were: [
         "manageiq-providers-openshift_monitoring_manager_event_catcher.target",
         "manageiq-providers-openshift_monitoring_manager_event_catcher@.service"
       ]
     # ./spec/manageiq/spec/models/miq_worker/systemd_common_spec.rb:19:in `block (3 levels) in <top (required)>'
```

Confused why this is showing up here though.
Maybe a rebase would fix?

agrare:
The test is correct I forgot to remove the systemd unit files for the deleted worker classes

agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/518

PR #518, opened 2024-02-29
Delete Prometheus Alerts MonitoringManager

agrare:
Prometheus Alert support was removed in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/506 and this gem is no longer used.

Depends:
- [x] https://github.com/ManageIQ/manageiq-providers-kubevirt/pull/231
Co-Depends:
* https://github.com/ManageIQ/manageiq/pull/22926
* https://github.com/ManageIQ/manageiq-providers-openshift/pull/257


kbrock:
think this was a valid failure

```
Failure/Error: parser.get_additional_attributes_graph(inventory)
```

agrare:
Interesting, I can't imagine how...

agrare:
@kbrock those are pending specs, the real failure is that we still have the PrometheusAlerts `MonitoringManager` here.  This will take some more work to untangle as we'll need a data migration to remove the records.

I did find an opportunity to remove some tech debt here though as we're including `HasInfraManagerMixin` and `HasMonitoringManagerMixin` in the base `ContainerManager` class which isn't right.  Opened https://github.com/ManageIQ/manageiq/pull/22926 to undo this.

config/settings.yml
```diff
@@ -1,10 +1,6 @@
 ---
 :ems:
   :ems_kubernetes:
-    :ems_monitoring:
-      :alerts_collection:
-        :open_timeout: 10.seconds
-        :timeout: 20.seconds
```
Fryguy, line 7: Do we need a database migration to delete any changes in SettingsChanges?

Fryguy: Or, for that matter, any existing workers or queue items?

Fryguy: and/or the EMS itself?

kbrock:
ok. I reproduced the error locally. (and it does not show up in `master` for me)

```bash
rspec spec/manageiq/spec/models/miq_worker/systemd_common_spec.rb:5 # MiqWorker::SystemdCommon.service_base_name every worker has a matching systemd target and service file
```

config/settings.yml
agrare: @Fryguy https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/518#issuecomment-1973540739

>  This will take some more work to untangle as we'll need a data migration to remove the records.

app/models/manageiq/providers/kubernetes/container_manager.rb
```diff
@@ -22,12 +23,6 @@ def allow_targeted_refresh?
     true
   end
 
-  def monitoring_manager_needed?
-    connection_configurations.roles.include?(
-      ManageIQ::Providers::Kubernetes::MonitoringManagerMixin::ENDPOINT_ROLE.to_s
-    )
-  end
```
kbrock, line 29: wonder if this is causing the issue

agrare: No I just need to delete the systemd units here

Fryguy:
> Haha slow down guys I'm working on it,

We're all so excited for the 🔥 🗑️ 💣 ✂️ 

(I suggest WIP it 😛)

agrare:
> (I suggest WIP it 😛)

No need now, was just making a flurry of changes earlier and some things were broken but it is good now

Fryguy (approved):

PR #531, opened 2024-06-28
Provide backward compatible way to globally set permitted classes

jrafanie:
This supports rails 7 and 6.1.

Similar to the change in: https://github.com/ManageIQ/manageiq/pull/22887

### DIFF
--- a/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job/dispatcher_spec.rb
+++ b/spec/models/manageiq/providers/kubernetes/container_manager/scanning/job/dispatcher_spec.rb
@@ -40,7 +40,12 @@ RSpec.describe ManageIQ::Providers::Kubernetes::ContainerManager::Scanning::Job:
     context "with container and vms jobs" do
       let(:container_image_classes) { ContainerImage.descendants.collect(&:name).append('ContainerImage') }
       before do
-        ActiveRecord::Base.yaml_column_permitted_classes |=  [ManageIQ::Providers::Openshift::ContainerManager::ContainerImage]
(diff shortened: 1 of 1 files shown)


Fryguy:
Minor, but can we
1. move this into a spec/support file in core
2. use that shared spec/support method here
3. Put a note in the spec/support entry that somehow let's us find it later (maybe a comment about the Rails version or something that we can search on).

My main concerns are duplication of code and then being able to find all those later when we need to undo it.

Fryguy:
Actually i said spec/support but the original code is in yaml_permitted_classes, so maybe it's just easy to refactor into another method on that module and use it here.  Just trying to avoid the duplication.

jrafanie:
> Minor, but can we
> 
> 1. move this into a spec/support file in core
> 2. use that shared spec/support method here
> 3. Put a note in the spec/support entry that somehow let's us find it later (maybe a comment about the Rails version or something that we can search on).
> 
> My main concerns are duplication of code and then being able to find all those later when we need to undo it.

Sure, I'll add a follow up item to the rails 7 todo list to find a generic way for all of the various repos to do the same function.  Or maybe just delete the rails 6.1 way when it's dropped.  Currently debugging tough api and ui-classic test failures so cleanup is less priority.

jrafanie:
ok, I added a followup item to https://github.com/ManageIQ/manageiq/issues/22052

jrafanie:
@Fryguy this is ready to go.  I'll come back and clean these up in the follow task: https://github.com/ManageIQ/manageiq/issues/22052.  When we drop rails 6.1, it should be easy to mass remove them.

Fryguy (approved):

PR #532, opened 2024-07-17
Clarify that KubeVirt is OSV

Fryguy:
@agrare Please review.


agrare:
@Fryguy I thought we were going to add a subclass of the KubeVirt provider for Openshift Virtualization?  We can do this too but I think when we have a proper provider type we would revert this.

Fryguy:
@agrare Yeah this is a temporary fix just to make it clear in the UI - Agreed that once we have the dedicated provider, then we revert this.

agrare (approved):

Fryguy:
Backported to radjabov via merge of master into radjabov branch

PR #539, opened 2024-09-30
Remove newlines from help message

Fryguy:
Newlines can cause trouble with translations as well as presentation in the UI. This particular string doesn't actually need the newlines, so this commit changes it to just not have them.

Alternative to ManageIQ/manageiq#23213

@agrare Please review. cc @jrafanie


Fryguy:
@agrare Total aside, but the entirety of the `Options` class uses `N_`, which I don't think actually does the translation.  `N_` is meant for constants where you want to delay the translation until later, but _something_ needs to call `_` on it.  I think this particular set is meant to be queried over the API and thus translated, so I think these are all wrong, or am I missing some detail here?

agrare (approved):

agrare:
@Fryguy any of the params_for_create/params_for_update methods are all `OPTIONS` calls over the API so I'm not sure if the locale is set in that context.  I'm having trouble finding where those `Options` are used maybe it is old code that is no longer needed?  Will need more investigation.

PR #549, opened 2025-03-17
Remove cben from CODEOWNERS

agrare:
As much as we love cben I bet he's tired of getting these notifications


cben:
Huh, I was just recently thinking to ask you how to drop me from owners :-)  I can easily filter the mails, but the public presence under "Reviewers" was misleading.  
LGTM

As one of the team said long ago "See you in the upstream". 👋  
(not that I'm much in Ruby nor k8s circles at present, but who knows :man_shrugging:)

PR #567, opened 2026-08-03
Add Pod Logs feature

ishaaq-48:
## Purpose

Implement provider-specific support for retrieving pod logs from Kubernetes/OpenShift clusters.

## Changes

- Add pod log retrieval methods
- Used Kubeclient API to fetch logs

## Testing

Verified logs are retrieved from a running pod in a Kubernetes/OpenShift cluster.

Depends on:
- https://github.com/ManageIQ/manageiq-ui-classic/pull/10201
- https://github.com/ManageIQ/manageiq/pull/23942 
- https://github.com/ManageIQ/manageiq-api/pull/1339

### DIFF
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_group/logs.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_group/logs.rb
@@ -1,0 +1,14 @@
+module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerGroup::Logs
+  def logs(container_name, options = {})
+    params = {:container => container_name}.merge(options)
+
+    ext_management_system
+      .connect(:service => "kubernetes")
+      .get_pod_log(
+        name,
+        container_project.name,
+        **params
+      )
+      .body
+  end
+end
--- a/app/models/manageiq/providers/kubernetes/container_manager/container_group.rb
+++ b/app/models/manageiq/providers/kubernetes/container_manager/container_group.rb
@@ -1,4 +1,6 @@
 class ManageIQ::Providers::Kubernetes::ContainerManager::ContainerGroup < ::ContainerGroup
+  include Logs
+
   alias_attribute :pod_uid, :ems_ref
 
   supports :capture


app/models/manageiq/providers/kubernetes/container_manager/container_group/logs.rb
```diff
@@ -0,0 +1,14 @@
+module ManageIQ::Providers::Kubernetes::ContainerManager::ContainerGroup::Logs
+  def logs(container_name, options = {})
```
Fryguy, line 2: ok I think I have idea how this works and I have some concerns.

Basically, if exposed directly over the API or the UI, then the API/UI pod would be making a direct connection to the management system, which we don't allow. Instead this logs request should go over the queue with the correct role for the EMS (I'm guessing ems_operations - @agrare ?).

The way we've done this on other providers, such as ansible tower, is to create a task and queue item associated with it. Then when the queue item completes, it stuff the response into the task. Then the caller can watch the task for completion and pull the response (i.e. the logs) from the task and then delete the task.

Over the API, this typically looks like call that returns a task_id.  Then the caller of the API can hit the tasks API and wait for completion.

ishaaq-48: Thanks for the detailed explanation @Fryguy ! That makes sense. I'm still learning the ManageIQ architecture and the common patterns used across the project, so this was very helpful. I'll rework the implementation to follow the queue/task pattern and update the PR.

ishaaq-48:
HI @Fryguy,

I tried like this is it correct 
```ruby
  def logs_queue(userid, options = {})
    raise MiqException::MiqUnsupportedOperation, unsupported_reason(:logs) unless supports?(:logs)

    task = MiqTask.create!(
      :name    => "Fetch logs for Pod #{name}",
      :userid  => userid,
      :state   => MiqTask::STATE_QUEUED,
      :status  => MiqTask::STATUS_OK,
      :message => "Queued log retrieval for pod #{name}"
    )

    MiqQueue.put(
      :class_name  => self.class.name,
      :instance_id => id,
      :method_name => "logs",
      :args        => [task.id, sanitize_log_options(options)],
      :role        => "ems_operations",
      :zone        => ext_management_system.my_zone,
      :queue_name  => ext_management_system.queue_name_for_ems_operations,
      :task_id     => task.id,
      :msg_timeout => 10.minutes
    )

    task.id
  end

  
  def logs(task_id, options = {})
    task = MiqTask.find_by(:id => task_id)
    task&.update_status(MiqTask::STATE_ACTIVE, MiqTask::STATUS_OK, "Fetching logs")

    body = raw_logs(options)

    task&.update!(:task_results => body)
    task&.update_status(MiqTask::STATE_FINISHED, MiqTask::STATUS_OK, "Logs retrieved")
    body
  rescue => err
    _log.error("Pod log fetch failed for #{name}: #{err.message}")
    task&.update_status(MiqTask::STATE_FINISHED, MiqTask::STATUS_ERROR, err.message)
    raise
  end

  def raw_logs(options = {})
    ext_management_system
      .connect(:service => "kubernetes")
      .get_pod_log(name, container_project.name, **options.symbolize_keys)
      .body
  end
```

API I kept post action same

```ruby
    def logs_resource(type, id, data)
      container_group = resource_search(id, type, ContainerGroup)
      api_action(type, id) do
        task_id = container_group.logs_queue(User.current_user&.userid, data || {})
        action_result(true, "Fetching logs for #{container_group.name}", :task_id => task_id)
      end
    end
``` 

image

Is this right approach or if you want any specific approach can u share any example or reference to better understand?

